PR #28, opened 2020-07-31
Initial work on matching displayfunc.c, aobj.c, fobj.c, wobj.c

PsiLupan:
* Match several functions across the files, with aobj.c bringing in all asm and inlining it
* Showcases how to continue to use inlined asm without affecting subsequent functions by declaring peephole on via pragma, since the asm keyword is disabling it
* Adds some newlines that were missing on the ends of files


src/sysdolphin/aobj.c
```diff
@@ -0,0 +1,1385 @@
...
+
+    if (aobj->flags & AOBJ_FIRST_PLAY) {
+        aobj->flags &= 0xF7FFFFFF;
+        rate = lbl_804DE438;
+    } else {
+        rate = aobj->framerate;
+        aobj->curr_frame += aobj->framerate;
+    }
+
+if ((aobj->flags & AOBJ_LOOP) && aobj->end_frame <= aobj->curr_frame) {
```
ProjectRevoTPP, line 130: formatting

PR #231, opened 2022-01-04
remove vestigial makefile (completely unused)

EpochFlame:


PsiLupan:
I don't see any reason to remove a makefile for building elf2dol on its own.

PR #241, opened 2022-01-12
label most math.s functions from Pikmin map

EpochFlame:
also split files a bit further. not sure about two of the global data labels, but the order is correct.


EpochFlame:
specifically, idk if lbl_80400770 and lbl_80400774 should be in trigf.s

ProjectRevoTPP (approved):

PR #275, opened 2022-02-17
[WIP] ftwalljump.c

Ka-lua:
- Replaces ftwalljump.s with ftwalljump.c
- fighter.h structs only contains used variables
- the function in ftwalljump.c has tentative variables names, needs more time to work out the game logic
- should I add the extern functions that I use to functions.h? The HSD_GObj is a problem then because of circular include dependencies. I kept the extern function in ftwalljump.c, and put them in comments in functions.h


PsiLupan:
Yes, declare them in functions.h - You can do so by declaring the GObj using `struct _HSD_GObj*` and the likes rather than attempting to use the typedef'd struct.

You also need to rebase your branch, because your changes to fighter are currently conflicting with Master.

PR #280, opened 2022-02-19
Disassemble ftzako{boy,girl}.s

r-burns:


PsiLupan:
Please add the function prototypes to the respective headers, then this PR should be good..

PR #287, opened 2022-02-20
Match ftgigakoopa

r-burns:
Also add some extra matching functions to other files in comments


src/melee/ft/chara/ftgigakoopa.c
```diff
@@ -0,0 +1,141 @@
+#include "ftgigakoopa.h"
+
+asm void ftGKoopa_OnDeath()
```
ProjectRevoTPP, line 3: Why not do this one?

r-burns: This is the only nontrivial function, and I think it's going to need some rethinking regarding the Fighter internals since this function is doing f32 operations on values that are currently scalar only. Might require refactoring into a union or FighterHeader idiom as you suggested. So I'd rather leave that decision to those more experienced :)

r-burns:
(rebased to fix conflicts)

PR #293, opened 2022-02-20
[WIP] fighter.c

Ka-lua:
EDIT: In a state to be merged after review & potential cleanup, if keeping the inline asm is okay.

I want to push this to keep fighter.h in sync with the repo, and to enable poeple who want to help to push individual functions to fighter.c. fighter.c should not be linked until all functions are done, but this PR currently builds OK.

All data is still in fighter.s, but all functions are in fighter.c.

fighter.h is relatively clean
- contains no unused variables except for a list of callbacks
- many Fighter struct variables still have the unverified akaneia names, I hope that's okay

fighter.c is more messy as the functions are still being decompiled.
- all functions are there as inline asm
- C functions can be ifdef'ed in and out for compilations.
- people already started working out of order (including my relatively large procUpdate function), so the inline asm must probably stay

TODO:
[x] remove the function prototypes that don't belong to fighter.h and move them somewhere else
    [x] fighter.h only contains function prototypes that are defined in fighter.s now
    [ ] moved most to fighter.c, still have to figure out their types and then move them to functions.h


src/sysdolphin/baselib/jobj.h
```diff
@@ -84,14 +84,14 @@ typedef struct _HSD_JObj {
     struct _HSD_JObj* next;
     struct _HSD_JObj* parent;
     struct _HSD_JObj* child;
-    u32 flags;
+    u32 x14_flags;
```
PsiLupan, line 87: You don't have to do this with sysdolphin related things. I do need to go back and put comments though about the position.

src/melee/ft/fighter.h
```diff
@@ -234,4 +575,45 @@ typedef struct _Fighter {
...
+void func_80067BB4(HSD_GObj* pPlayerEntity);
+void func_80067C98(Fighter*);
+void func_80068354();
+void func_800686E4();
+void func_8006876C();
+void func_80068854();
+void func_80068914();
+void func_80068E64();
+void func_80068E98();
+void func_800693AC(void*, s32, s32, s32, f32, f32, f32);
```
r-burns, line 591:
```suggestion
void func_800693AC(HSD_GObj*, s32, s32, s32, f32, f32, f32);
```

Ka-lua:
I think my fighter.c is ready if having everything as inline asm for building is okay. You can CTRL-F for ======= to see which code sections there are in fighter.c, these are:
- a **fpclassify function** that should go to math.h eventually (It's the same as in pikmin2)
- **external functions** I don't know their proper prototypes yet, so I didn't move them to functions.h yet. Should I move them to functions.h anyway? Would save some work I think.
- **external variables** same question: keep them in fighter.c for now or move to variables.h?
- **fighter.c variables** Those are from the fighter.s data sections. Some of these should go to fighter.h and variables.h too I guess, but their definition will stay in fighter.c.
- **fighter.c functions** mostly inline asm, C function alternatives can be switched on and they compile but don't OK yet (data loading and some remaining matching issues)

PR #295, opened 2022-02-21
Match ftdrmario

r-burns:
func_80149954 needed to be in a separate TU to have the correct relocs
for the function pointer it uses, so I created the file ftdrmario_2.c


PsiLupan:
Rather than name this _2, if you can determine which Special it's the start of (I'm assuming SpecialN) then we can properly name it as all the ft files need to be split into respective Specials.

PR #296, opened 2022-02-21
Match ftcliffcommon.s

JohnnyUrosevic:
gotta do these faster so I don't have to worry about merging


src/melee/ft/fighter.h
```diff
@@ -6,18 +6,42 @@
 
 #include "sysdolphin/baselib/gobj.h"
 
-typedef struct { float x, y;    } Vec2;
-typedef Vec Vec3;
+// Points to data in PiCo.dat
+typedef struct _ftCommonData {
+  u8 data_filler_1[0x480];
```
r-burns, line 11: if you're reindenting this file, would you mind using 4 spaces like the others?

JohnnyUrosevic: i would not mind, I only did that because it was what others were doing in that file

PR #314, opened 2022-02-26
Match `HSD_JObj`

PsiLupan:
Has some work done on JObj. Can't push till the full file is done due to the usage of inlines in the code.


ribbanya (approved):
Up-to-date with `master` and `OK`. Should merge imo.

PR #315, opened 2022-02-26
WIP: HSD TObj match

PsiLupan:
Has some work done on matching functions. Since I've decided to not extern floats, it's non-matching presently.

Since I plan to finish the full file before committing, putting up this PR for posterity.


ribbanya (approved):
Up-to-date and `OK`.

PR #316, opened 2022-02-26
Match `HSD_LObj`, ongoing

PsiLupan:
Matched several functions in the file. Work still on-going, so putting up PR for posterity.


ribbanya (approved):
Up-to-date and `OK`. I had to wrap `HSD_LObjSetup` and  `HSD_LObjGetLightVector` in `NON_MATCHING` for now but it should be quite easy to bridge the gap.

ribbanya (changes requested):
Found some extra files created by the merge. Fixing.

ribbanya (approved):

PR #333, opened 2022-03-25
Split out & match ftlib

r-burns:
Also applies some treewide relabeling to use HAL names exposed in ftlib's assert statements


src/melee/ft/chara/ftMars/ftMarsSpecialHi.c
```diff
@@ -140,9 +140,9 @@ void lbl_80138638(HSD_GObj* gobj) {
     s32 num;
 
     // this is required, dunno why
-    ft->xE0_airState;
+    ft->ground_or_air;
 
-    if (ft->xE0_airState == 1) {
+    if (ft->ground_or_air == GA_Air) {
```
PsiLupan, line 145: I'd still like the naming to retain the offset.

src/melee/it/id.h
```diff
@@ -0,0 +1,278 @@
...
+    0x31, // MAYBE Dr. Mario's pill?
+    0x32, // Kirby's Cutter beam
+    0x33, // Kirby's Hammer
+    0x34, //
+    0x35, //
+    0x36, // Fox's Laser
+    0x37, // Falco's Laser
+    0x38, // Fox's shadow
+    */
+    It_Kind_FalcoShadow = 0x39, // Falco's shadow
```
PsiLupan, line 75: Call it AfterImage. That's the internal name in structs and makes more sense than Shadow which would be their floor shadow, if you didn't know better.

PsiLupan (changes requested):
itGetKind will probably also change later to retain the naming per contribution guidelines. I just don't know the filename it's in offhand to keep to convention.

I realize it's one of the things we have in an assert string but I'm trying to avoid a situation of having a file with 10 different naming conventions that can't be followed.

PR #344, opened 2022-04-04
Ftmario

nicolasrmerz:


src/melee/ft/chara/ftmario.c
```diff
@@ -160,3 +181,134 @@ s32 func_800E0CE0(s32 arg0) {
 
     return lbl_803C5A20[offset - 0xe];
 }
+
+//https://decomp.me/scratch/cdQ5o
+void func_800E0D1C(HSD_GObj* gobj) {
```
r-burns, line 186: This needs to return something, as used by drmario:

https://github.com/doldecomp/melee/blob/ae0c84368812db6f4f8dc403e0881ce3ab1fdf47/src/melee/ft/chara/ftdrmario_2.c#L13-L14

I think if you just `return r3` it will still match

nicolasrmerz: Got it, thanks for the catch :)

PsiLupan:
Curious why you have a function named "copy_nonmatching" despite it being a matching function?

nicolasrmerz:
> Curious why you have a function named "copy_nonmatching" despite it being a matching function?

@r-burns wrote it in his fork of func_800E0EE0 on decomp.me (the one in the comment) and I just copied it verbatim, but if I understand the naming correctly it means nonmatching not in the sense of decomp but to copy over integers into the array that don't match the ft attributes in the conditional. I'll rename it to something else for less confusion though.

PR #346, opened 2022-04-06
Split & match gmmain

r-burns:


src/melee/gm/gmmain.c
```diff
@@ -0,0 +1,201 @@
...
+
+extern GXRenderModeObj lbl_80401168;
+extern PadLibData HSD_PadLibData;
+extern char lbl_803EA6C8[]; // build timestamp string
+extern s32* seed_ptr;
+extern void lbl_8015FD24();
+extern void lbl_803762C4();
+extern void func_801692E8(int, struct datetime* datetime);
+
+enum {
```
PsiLupan, line 18: These DB levels aren't really unknown, since the menus state them.

See: https://github.com/eigenform/melee-re/blob/master/docs/DEBUG.md

PR #401, opened 2022-07-04
Change github actions from wine to wibo

BR-:
Not sure if this will work, but worth trying to get rid of the spurious wine failures.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,9 +32,11 @@ jobs:
         mv tools/GC tools/mwcc_compiler
     - name: Build WiBo
       working-directory: tools/WiBo
-      run: make -j$(nproc)
+      run: |
+        cmake -B build
+        cmake --build build
     - name: Build Melee
-      run: make -j$(nproc) ${{ matrix.makeflags }} WINE=./tools/WiBo/wibo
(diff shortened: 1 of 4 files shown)


BR-:
Don't merge this yet.

BR-:
Ready to merge! As a bonus it'll output the current match % and upload the map file on successful builds.

BR-:
WiBo is about to switch from Makefile to CMake. As that'll break this PR, I'll fix and reopen this once it's ready.

r-burns:
I think you have to explicitly link the filesystem library on gcc 8. Try adding -lstdc++fs to LDFLAGS

BR-:
Yeah, I added that for the makefile but now it was lost in the transition to cmake. Going to PR it on the wibo side.

r-burns:
I don't think that should be PR'd upstream since it's a peculiarity of certain older gcc versions, modern compilers don't need this. At the very least it should be conditional on the result of a try_compile.

r-burns:
Does the linker error only show up when compiled with optimizations? You could try to cmake build in Debug mode to check.

BR-:
Hopefully good to merge. If the patch is bad we'll find out sooner or later with testing. Not like we can do much worse than the current wine setup.

EDIT: oop, left WIBO_DEBUG on by default. removed

PR #402, opened 2022-07-06
Decompile mtx.s

CelestialAmber:
Marking this as a draft for now because I want to try getting rid of the fakematches, especially in the determinant function.
I also updated the Makefile so that it would work on Mac.
(Thanks to Ventri and everyone else for their help!)


CelestialAmber:
Ignore the latest commit, there's still one nonmatching function.

src/sysdolphin/baselib/mtx.c
```diff
@@ -0,0 +1,516 @@
...
+        PSMTXScale(temp, 1.0d/arg4->x, 1.0d/arg4->y, 1.0d/arg4->z);
+        PSMTXConcat(temp, arg0, arg0);
+    }
+    
+    PSMTXTrans(temp, arg3->x, arg3->y, arg3->z);
+    PSMTXConcat(temp, arg0, arg0);
+}
+
+#ifdef NON_MATCHING
+void func_8037A54C(Mtx arg0, Mtx arg1, Mtx arg2, f32 arg3)
```
ProjectRevoTPP, line 394: https://decomp.me/scratch/h8te3

src/sysdolphin/baselib/mtx.h
```diff
@@ -1,10 +1,49 @@
...
+void func_8037A6D8(void* arg0);
 HSD_ObjAllocData* HSD_VecGetAllocData(void);
+void HSD_VecInitAllocData(void);
 HSD_ObjAllocData* HSD_MtxGetAllocData(void);
+void HSD_MtxInitAllocData(void);
+
+extern float func_80022DBC(float);
+extern float func_someCalcAngle_80022C30(float, float);
+
+inline float sqrtf(float x)
```
ProjectRevoTPP, line 30: please use the one from math.h and dont duplicate macros.

Makefile
```diff
@@ -56,23 +56,25 @@ MWCC_LD_VERSION := 1.1
...
+  CPP     := $(DEVKITPPC)/bin/powerpc-eabi-cpp.exe -P
+  SHA1SUM := sha1sum
 else
   WINE ?= wine
   # Disable wine debug output for cleanliness
   export WINEDEBUG ?= -all
   # Default devkitPPC path
   DEVKITPPC ?= /opt/devkitpro/devkitPPC
+  CPP     := $(DEVKITPPC)/bin/powerpc-eabi-cpp -P
+  SHA1SUM := shasum
```
r-burns, line 68: Why was this changed? `sha1sum` is part of gnu coreutils so this should be the same on windows and linux

CelestialAmber: I had to change this for Mac. Maybe we can just handle that in another if statement.

r-burns: Ah yeah I see, macos just has shasum. On linux shasum is part of perl, so I'd rather we keep using sha1sum on windows/linux and make this part mac-specific.

PR #414, opened 2022-07-13
Use shasum command on macOS, sha1sum otherwise

r-burns:
@CelestialAmber lmk if this still works for you


CelestialAmber:
It works for me

PR #458, opened 2022-08-13
Create `convert_data.py`

ribbanya:
* Converts `4byte` data in `.s` sections to ascii strings, floats, and/or integers.
* Options include toggling each data type and setting acceptable value ranges for each, along with several options to handle `NULL`.
* Demonstration: https://github.com/ribbanya/melee/commit/8634196dfe1bcc4410320a3bfa162d911dcfc7c8


ribbanya:
Superseded by #454.

PR #460, opened 2022-08-13
restore item2.c

VetriTheRetri:
This is a file that was matched a while back, which contains two functions that break Frank, causing it to create branches that lead out of bounds.

They have been "patched" with peephole off on NON_MATCHING builds and hybrid ASM on matching ones.


ribbanya (approved):

ribbanya (approved):

asm/melee/ft/chara/ftfox.s
```diff
@@ -2,7 +2,6 @@
 
 .section .text  # 0x80005940 - 0x803B7240
 
-
 .section .data
     .balign 8
 .global as_table_fox
```
ribbanya, line 7: This can be converted to C too, right?

ribbanya (changes requested):

PR #461, opened 2022-08-13
Always use `.asciz` for strings in `convert_data.py`

ribbanya:
Removes false positives and hanging `NULL` bytes.


ribbanya:
Superseded by #462.

PR #466, opened 2022-08-20
Create global context generation script

ribbanya:
Creates `generate_context.sh`, which generates context for the entire codebase. Header files are modified to not throw errors when run through m2c.

Usage:
```sh
tools/generate_context.sh
python tools/m2c/m2c.py -t ppc-mwcc-c --context ctx.c -f func_DEADBABE asm/path/to/file.s
```
Run it from MSYS2 terminal. Make sure your `python` command has the necessary packages to run m2c.


ribbanya:
Superseded by #462.

PR #475, opened 2022-08-23
Match HSD class

r-burns:


src/sysdolphin/baselib/class.c
```diff
@@ -142,3 +143,338 @@ HSD_MemoryEntry* GetMemoryEntry(s32 idx)
...
+        cur = next;
+    }
+    class_info->amnesia(class_info);
+    class_info->head.child = NULL;
+    class_info->head.parent = NULL;
+    func_80420FBC(class_info, 0, 1);
+}
+
+// (name based on KAR address)
+void func_804211B4(const char* name, HSD_ClassInfo* class_info)
```
r-burns, line 411: Left these comments on functions I don't know the names of, which exist in KAR but are stripped in melee because their only uses are auto-inlined. Maybe Psi knows what to call them.

ribbanya: Maybe rename them to `kar_804211B4` for now so people don't think it's a Melee address?

ribbanya (approved):

src/sysdolphin/baselib/class.c
ribbanya: I'll just make a new issue so this can be merged.

PR #493, opened 2022-08-26
Replace `f` with `F` in `f32` literals

ribbanya:
Replaces `\b[+-]?(\d+([.]\d*)?([eE][+-]?\d+)?)f\b`
with `$1F`
in `**/*.c`, `**/*.h`

As per #487.

Note that implicit `f32` literals, and all `f64` literals, still need to be addressed.


ribbanya:
This will be introduced more gradually as code is updated.

PR #500, opened 2022-08-27
Fix naming for GXInitTlutObj/GXInitTlutRegion

r-burns:


ribbanya (approved):
Thanks, how did you know they were wrong?

r-burns:
They're both exact matches with the ones from Zoids VS which have debug symbols

PR #503, opened 2022-08-27
More dolphin + sysdolphin labeling

r-burns:


ribbanya (approved):
In the future, could you provide a table of the renames like
From|To
-|-
func_80337CD4|DVDClose
et|cetera

... or a `sed` command or something? That way when reviewing it's easy to just run the renames locally on `master` and compare to the PR.

r-burns:
For sure, that's a good idea. Would make resolving conflicts trivial too.

PR #523, opened 2022-09-01
Match more of HSD LObj

r-burns:
Not finished, but at a point where this is still shiftable.


src/sysdolphin/baselib/lobj.c
```diff
@@ -220,13 +220,14 @@ void HSD_LObjReqAnimAll(HSD_LObj *lobj, f32 startframe)
     }
 }
 
-Vec const position = {0.0F, 0.0F, 0.0F};
-Vec const interest = {0.0F, 0.0F, 0.0F};
-
-#ifdef NON_MATCHING
+Vec const lbl_803B94A0 = {0.0F, 0.0F, 0.0F};
+Vec const lbl_803B94AC = {0.0F, 0.0F, 0.0F};
```
ribbanya, line 224:
```suggestion
Vec const lbl_803B94A0 = { 0.0F, 0.0F, 0.0F };
Vec const lbl_803B94AC = { 0.0F, 0.0F, 0.0F };
```

src/sysdolphin/baselib/lobj.c
```diff
@@ -340,106 +285,1065 @@ void HSD_LObjSetup(HSD_LObj *lobj, GXColor color, f32 shininess, u32 unused)
...
+        GXInitLightSpot(&lobj->lightobj, cutoff, spot_func);
+    }
+}
+
+extern char lbl_804D5D24[4];
 #pragma push
-asm void HSD_LObjSetup(HSD_LObj *lobj, GXColor color, f32 shininess, u32 unused)
-{ // clang-format off
+asm void HSD_LObjSetupInit()
+{
```
ribbanya, line 416: Was this a mistake?

src/sysdolphin/baselib/lobj.h
```diff
@@ -46,34 +46,34 @@
 
 typedef struct _HSD_LightPoint {
     f32 cutoff;
-    u8 point_func;
+    u32 point_func;
     f32 ref_br;
     f32 ref_dist;
-    u8 dist_func;
+    u32 dist_func;
```
ribbanya, line 52: Are these enums, or pointers? Or unsigned constants?

ribbanya (changes requested):

src/sysdolphin/baselib/lobj.h
r-burns: Not sure, I don't think their point of use has been decomped yet.

ribbanya: I kind of would like there to be a `typedef s32 UnkEnum;` in `types.h` at this point, but what do you think?

src/sysdolphin/baselib/lobj.c
r-burns: Not sure what you mean?

ribbanya:
```diff
-asm void HSD_LObjSetup(HSD_LObj *lobj, GXColor color, f32 shininess, u32 unused)
-{ // clang-format off
+asm void HSD_LObjSetupInit()
+{
```

src/sysdolphin/baselib/lobj.h
r-burns: I don't think we're certain it's an enum, and hiding the type behind an alias seems to add another layer of indirection for little gain.

ribbanya: Hm, alright.

src/sysdolphin/baselib/lobj.c
ribbanya: Oh, the diff just shifted weirdly. Nevermind.

r-burns: HSD_LObjSetup is now decomped. HSD_LObjSetupInit is a different function.

ribbanya (approved):

PR #528, opened 2022-09-02
Continue matching HSD jobj

r-burns:
Also name some locals from zoids debug map


src/sysdolphin/baselib/jobj.c
```diff
@@ -23,3 +27,249 @@ void HSD_JObjCheckDepend(HSD_JObj* jobj)
...
+    if (!(jobj->flags & 0x1000)) {
+        HSD_JObj* child = jobj->child;
+        while (child != NULL) {
+            func_8036EFAC(child, cb, args);
+            child = child->next;
+        }
+    }
+}
+
+void func_8036F0F0(HSD_JObj* jobj, void (*cb)(HSD_JObj*, void*, u32), void* args)
```
ribbanya, line 87: Should this cb be a typedef?

src/sysdolphin/baselib/jobj.c
```diff
@@ -23,3 +27,249 @@ void HSD_JObjCheckDepend(HSD_JObj* jobj)
...
+{
+    BOOL result = FALSE;
+    if (jobj != NULL && jobj->scl != NULL)
+        result = TRUE;
+    return result;
+}
+
+void HSD_JObjMakeMatrix(HSD_JObj* jobj)
+{
+    Vec* scl;
```
ribbanya, line 114: Does this mean "scale"? Was it an assert symbol?

src/sysdolphin/baselib/jobj.c
```diff
@@ -23,3 +27,249 @@ void HSD_JObjCheckDepend(HSD_JObj* jobj)
...
+
+void HSD_JObjRemoveAnimAllByFlags(HSD_JObj* jobj, u32 arg1)
+{
+    if (jobj != NULL) {
+        HSD_JObjRemoveAnimByFlags(jobj, arg1);
+        if (!(jobj->flags & 0x1000)) {
+            HSD_JObj* child_jobj = jobj->child;
+            while (child_jobj != NULL) {
+                HSD_JObjRemoveAnimAllByFlags(child_jobj, arg1);
+                child_jobj = child_jobj->next;
```
ribbanya, line 198:
```suggestion
            HSD_JObj* child = jobj->child;
            while (child != NULL) {
                HSD_JObjRemoveAnimAllByFlags(child, arg1);
                child = child->next;
```

src/sysdolphin/baselib/jobj.c
```diff
@@ -23,3 +27,249 @@ void HSD_JObjCheckDepend(HSD_JObj* jobj)
...
+    }
+    fobj_p = &aobj->fobj;
+    while ((fobj = *fobj_p) != NULL) {
+        if (fobj->obj_type == 0xC) {
+            *fobj_p = fobj->next;
+            fobj->next = aobj->fobj;
+            aobj->fobj = fobj;
+            break;
+        }
+        fobj_p = &fobj->next;
```
ribbanya, line 273: This reads better to me:
```suggestion
    HSD_FObj* fobj;
    HSD_FObj** cur;

    if (aobj == NULL || aobj->fobj == NULL) {
        return;
    }
    cur = &aobj->fobj;
    while ((fobj = *cur) != NULL) {
        if (fobj->obj_type == 0xC) {
            *cur = fobj->next;
            fobj->next = aobj->fobj;
            aobj->fobj = fobj;
            break;
        }
        cur = &fobj->next;
```

ribbanya (changes requested):

src/sysdolphin/baselib/jobj.c
r-burns: I'd rather keep `fobj_p` or similar to indicate that this is the same as `fobj`. This is a weird pattern where the pointer and pointer-pointer are both needed to match.

src/sysdolphin/baselib/jobj.c
r-burns: Name based on jobj->scl

src/sysdolphin/baselib/jobj.c
r-burns: Will do so once it is used elsewhere in the future.

src/sysdolphin/baselib/jobj.c
ribbanya: Yeah, I was just wondering what it means and why it's so abbreviated.

src/sysdolphin/baselib/jobj.c
ribbanya: Are you ok with `fobj_ptr`?

ribbanya:
Force-pushing makes the history weird (my suggestions can't be one-click committed anymore, for example). It's all squashed anyway so please only do linear pushes if possible.

src/sysdolphin/baselib/jobj.c
r-burns: Not sure, can change it to scale if that's clearer

ribbanya: It's fine for now. If it matches HAL then cool; if it doesn't then we can look at it later.

src/sysdolphin/baselib/jobj.c
```diff
@@ -23,3 +27,249 @@ void HSD_JObjCheckDepend(HSD_JObj* jobj)
...
+            has_dobj = TRUE;
+        }
+        if (has_dobj) {
+            HSD_DObjRemoveAnimAllByFlags(jobj->u.dobj, flags);
+        }
+        HSD_RObjRemoveAnimAllByFlags(jobj->robj, flags);
+    }
+}
+
+void HSD_JObjRemoveAnimAllByFlags(HSD_JObj* jobj, u32 arg1)
```
PsiLupan, line 190: Fix the arg name to `flags`

src/sysdolphin/baselib/jobj.c
```diff
@@ -23,3 +27,249 @@ void HSD_JObjCheckDepend(HSD_JObj* jobj)
...
+    }
+}
+
+void HSD_JObjResetRST(HSD_JObj* jobj, HSD_Joint* joint)
+{
+    if (jobj == NULL || joint == NULL) {
+        return;
+    }
+    JObjResetRST(jobj, joint);
+    if (!(jobj->flags & 0x1000)) {
```
PsiLupan, line 52: 0x1000 is `INSTANCE`

src/sysdolphin/baselib/jobj.c
```diff
@@ -23,3 +27,250 @@ void HSD_JObjCheckDepend(HSD_JObj* jobj)
...
+
+void HSD_JObjRemoveAnimByFlags(HSD_JObj* jobj, u32 flags)
+{
+    BOOL has_dobj;
+    if (jobj != NULL) {
+        if (flags & 1) {
+            HSD_AObjRemove(jobj->aobj);
+            jobj->aobj = NULL;
+        }
+        if (jobj->flags & 0x4020) {
```
PsiLupan, line 179: This looks like `union_type_dobj` or some type of inline, given it shows up multiple times.

ribbanya (changes requested):
I want to look at the strings a little more before merging this.

ribbanya (approved):

PR #531, opened 2022-09-02
Use new `calcprogress`

ribbanya:
`Makefile` and `build.yml` are modified to accomodate the new path.

Draft until it's tested properly.


ribbanya (commented):
Looks okay to me now. Can someone else review this?

PR #534, opened 2022-09-03
Fix SetScaleHack by making HSD_JObjSetMtxDirty a macro.

ProjectRevoTPP:
This affects at least two functions, also affects HSD_JObjSetFlagsAll since the manual inlining was required to fix the function. The solution may be that this function is a defined macro instead of an inline, which is rather odd, but it's the only thing that fixes both hacks/functions.

### DIFF
--- a/include/sysdolphin/baselib/jobj.h
+++ b/include/sysdolphin/baselib/jobj.h
@@ -145,6 +145,7 @@ void HSD_JObjAnimAll(HSD_JObj *); // asm/sysdolphin/baselib/jobj.s
 
 #pragma push
 #pragma always_inline on
+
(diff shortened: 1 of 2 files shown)


ribbanya (commented):
Does this allow
https://github.com/doldecomp/melee/blob/7f6bb365b68e1d5b4a532c229a629709bf83ea43/include/sysdolphin/baselib/jobj.h#L147
to be removed?

RevoSucks:
> Does this allow
> 
> https://github.com/doldecomp/melee/blob/7f6bb365b68e1d5b4a532c229a629709bf83ea43/include/sysdolphin/baselib/jobj.h#L147
> 
> to be removed?

Yes.

ribbanya (approved):

PR #535, opened 2022-09-03
Continue matching HSD jobj

r-burns:
All but 3 of jobj's functions are matched.

The remaining 3 have some issues such as regswaps or stack usage,
but their NON_MATCHING code should be functionally equivalent.

First pass of decomp, some more cleanup/renaming is needed.

Also fixes shiftability.


include/sysdolphin/baselib/jobj.h
```diff
@@ -61,7 +61,8 @@
 #define SPLINE (1 << 14)
 #define FLIP_IK (1 << 15)
 #define SPECULAR (1 << 16)
-#define USE_QUATERNION (1 << 17)
+#define USE_QUATERNION (1 << 17) // TODO remove
```
ribbanya, line 64: Can make an issue for this.

src/sysdolphin/baselib/jobj.c
```diff
@@ -170,18 +175,12 @@ void HSD_JObjMakeMatrix(HSD_JObj* jobj)
...
             HSD_AObjRemove(jobj->aobj);
             jobj->aobj = NULL;
         }
-        if (jobj->flags & 0x4020) {
-            has_dobj = FALSE;
-        } else {
-            has_dobj = TRUE;
-        }
-        if (has_dobj) {
+        if (jobj->flags & 0x4020 ? FALSE : TRUE) {
```
ribbanya, line 183: I see this a lot everywhere so it should be a macro. Not sure if the `x & flags ? FALSE : TRUE` can be the same macro as `x & flags ? TRUE : FALSE` though.

src/sysdolphin/baselib/jobj.c
```diff
@@ -274,3 +273,1803 @@ void JObjSortAnim(HSD_AObj* aobj)
...
+    jobj->robj = HSD_RObjLoadDesc(joint->robjdesc);
+    jobj->rotate.x = joint->rotation.x;
+    jobj->rotate.y = joint->rotation.y;
+    jobj->rotate.z = joint->rotation.z;
+    jobj->scale = joint->scale;
+    jobj->translate = joint->position;
+    PSMTXIdentity(jobj->mtx);
+    jobj->scl = NULL;
+    if (joint->mtx != NULL) {
+        jobj->envelopemtx = HSD_MtxAlloc();
```
ribbanya, line 960: Is this an assert symbol? Otherwise should be `envelope_mtx`.

src/sysdolphin/baselib/jobj.c
r-burns: It's probably the actual definition of `union_type_dobj`. Lots of other macros need to be changed based on assert strings, was going to make another pass to clean these up.

PR #539, opened 2022-09-04
Replace hardcoded pi values with M_PI

r-burns:


ribbanya (commented):
Did you search the binary for exact matches, or only the C source?

r-burns:
Just the source. There are probably more occurrences, these are just the ones I noticed.

ribbanya:
Do you want to do a full pass in this PR? We could use a `.define` in GAS and your `#define` in C.

src/melee/ft/chara/ftMario/ftMario_SpecialHi.c
```diff
@@ -51,7 +51,7 @@ void ftMario_SpecialAirHi_Anim(HSD_GObj* gobj)
 }
 
 #define DEGREES_TO_RADIANS (3.14159265359f / 180.0f)
```
ribbanya, line 53: Is this instance different bitwise from the define?
```suggestion
#define DEGREES_TO_RADIANS (3.14159265359f / 180.0f)
```

ribbanya (changes requested):

r-burns:
I'm just doing the no-brainers here, there are some nontrivial cases that will require some discussion - like the f32-typed `3.1415927410125732f` values, which are probably from a macro or an inline function.

src/melee/ft/chara/ftMario/ftMario_SpecialHi.c
r-burns: Yeah not sure what to do with these, should probably consolidate all the deg-to-rad conversion constants as well. This is float-typed so it needs to be part of a macro or an inline.

ribbanya (approved):

PR #556, opened 2022-09-09
Match HSD fog

r-burns:


include/sysdolphin/baselib/fog.h
```diff
@@ -3,6 +3,53 @@
 
 #include <dolphin/types.h>
 
-void HSD_FogSet(s32);
+#include <sysdolphin/baselib/object.h>
+
+typedef struct _HSD_FogAdj {
+    HSD_Obj parent;
+    s16 center; // 0x08
```
PsiLupan, line 10: Correct these comments to the `/* 0xoffset */ type var;` style, please.

src/sysdolphin/baselib/fog.c
```diff
@@ -0,0 +1,256 @@
...
+
+typedef struct _GXFogAdjTbl {
+	u16 r[10];
+} GXFogAdjTbl;
+
+const GXColor lbl_804DE6F0 = { 0 };
+
+void HSD_FogSet(HSD_Fog* fog)
+{
+    GXFogAdjTbl sp88;
```
PsiLupan, line 19: Let's get these renamed from autogenerated names for the sake of cleanup. I have the actual names labeled in FRAY.

src/sysdolphin/baselib/fog.c
```diff
@@ -0,0 +1,256 @@
...
+void HSD_FogInit(HSD_Fog* fog, HSD_FogDesc* desc)
+{
+    if (fog != NULL) {
+        if (desc != NULL) {
+            fog->type = desc->type;
+            fog->start = desc->start;
+            fog->end = desc->end;
+            fog->color = desc->color;
+        } else {
+            f32 sp10[6];
```
PsiLupan, line 108: Another autogenerated name. In this case it's `v` - you can extend that to viewport if you want to be explicit but I think the function name covers that.

src/sysdolphin/baselib/fog.c
```diff
@@ -0,0 +1,256 @@
...
+{
+    HSD_FogAdj* adj = HSD_FogAdjAlloc();
+    adj ? (void) 0 : __assert("fog.c", 0xE6, "adj");
+    HSD_FogAdjInit(adj, desc);
+    return adj;
+}
+
+void HSD_FogAdjInit(HSD_FogAdj* adj, HSD_FogAdjDesc* desc)
+{
+    f32 sp10[6];
```
PsiLupan, line 138: Another autogenerated name for `v`

src/sysdolphin/baselib/fog.c
```diff
@@ -0,0 +1,256 @@
...
+        case 20:
+            if (fog->fog_adj != NULL) {
+                fog->fog_adj->center = fv->fv;
+            }
+            break;
+        }
+    }
+}
+
+inline BOOL ref_DEC(void* o)
```
PsiLupan, line 220: `ref_DEC` is an object.c function, let's move this there.

include/sysdolphin/baselib/fog.h
ribbanya: This will need to be addressed in a new issue because both styles are being used throughout the codebase right now. Your style is probably more readable but this style is more friendly to intellisense (particularly VS Code's).

ribbanya (approved):

include/sysdolphin/baselib/fog.h
PsiLupan: The reason it's mixed usage is because of me copy-pasting from structs in FRAY, which I think was done here. It's been something I've tried to cleanup as I actually decompile files themselves.

I was using this style because of Intellisense at the time and changed to match how other projects handle it with `/* 0xA */, but we should settle on which needs to be done if that's the case.

ribbanya: Tracked in #557.

ribbanya (approved):

src/sysdolphin/baselib/fog.c
r-burns: Sure, there are a couple others in jobj.c that need to be moved as well. I can't move the ones containing asserts yet though, because the object.h \_\_FILE__ string is referenced from the nonmatching HSD_JObjRemoveAll's inline asm.

r-burns:
Okay, all comments should be addressed now :)

ribbanya (dismissed):

PsiLupan (approved):

PR #566, opened 2022-09-16
Lints for function declarations

r-burns:


src/melee/gr/stage.h
```diff
@@ -119,7 +113,7 @@ typedef struct _StageInfo {
             GXColor color;
         }* ptr;
     }* x12C;
-    Vec3 x130, x13C, x148, x154, x160, x16C;
+    Vec x130, x13C, x148, x154, x160, x16C;
```
ribbanya, line 116: Is `Vec3` gone everywhere now?

r-burns: No, only in this header - just wanted to get rid of that weird #ifndef M2CTX

ribbanya (approved):

PR #576, opened 2022-10-24
Migrate `.text` section of `ftcoll` to C and match a few functions

ribbanya:

PR #578, opened 2022-10-24
func_8007B7A4

stephenjayakar:


ribbanya (changes requested):
Please wait on #576. In the future, you may want to submit fewer PRs with more commits each. Most people do one PR per file, or sometimes one PR per module (`ft`, `gx`, etc).

PR #579, opened 2022-10-26
Migrate `os` to C

ribbanya:
This is a bit of an experiment. The idea is to roughly follow this workflow:

1. Migrate the `.text` section of a translation unit (TU) to C
2. Create header files for the public-facing methods and types in the TU
3. Discern data types present in the TU
4. Migrate the data sections to C (now the TU no longer has any `.s` parts)
5. Clean up the headers
6. Match the functions

The main advantage to this approach is that anyone, even brand-new contributors, can match functions in the middle of even very large TUs, without needing to match the whole thing. Another advantage is that once a TU is in C, it's much easier to tell how functions are meant to link together, much easier to identify common data types, etc., because of C's strong typing compared to assembly—even if no functions in the TU are matched (just converted to inline `asm`).

The one sticking point is copying the assembly of a function into a scratch in order to match it, but [dadosod](https://github.com/InusualZ/dadosod) already allows us to do this from C-inlined assembly, and future tooling will only make this easier. Furthermore, if a function is already inlined assembly within C, and someone wants to match it, `asm-differ` is able to diff the two versions, which it cannot easily do if the function originates in a `.s` file.

~~This week I plan to tweak `build.sh` so that it can optionally dump the assembly of completed builds to a folder, which can then be used to create scratches or to diff `build` vs. `expected` as plaintext.~~ Done in #580.


PsiLupan:
A vast amount of these functions, like most SDK, are handwritten asm. I'm not sure there's really anything of value here from the perspective new contributors, since it'll be removing the `unk_t` on most things and just making them return void. This is a large part of why I'd marked SDK low-prio.

I don't see any issue with this from the perspective of just doing it, but I don't think any new person should be pointed at it. It's strictly busy work on so much of it.

PsiLupan (approved):

ribbanya:
Yeah, the `unk_t` stuff is obviously just a placeholder. One of the reasons I find it useful, rather than being `void`, is that the compiler will complain if it shows up elsewhere (such as `functions.h`) and it forces you to come up with a real signature for it.

Perhaps OS was a strange choice of first candidate for this migration, because as you say a lot of it is handwritten asm anyway. But I'm working on `ft` now and it's already resulted in cleaning up a lot of old header files and `extern`s.

PR #592, opened 2022-11-19
Create `compile_flags.txt` and `.vimrc`

ribbanya:
`compile_flags.txt` tells `clangd` how to process the source files. This is a placeholder until I figure out how to get `clangd` to run `mwcc` directly, which I believe is possible.

`.vimrc` turns on syntax highlighting for C (instead of C++) and turns on doxygen highlighting. Also shows a visual indicator on column 100. Can be activated in Vim or NeoVim with `source .vimrc`, or by creating an autocmd to do so for you.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -1,0 +1,15 @@
+set tabstop=4
+set softtabstop=4
+set shiftwidth=4
+set noexpandtab
+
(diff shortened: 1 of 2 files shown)


ribbanya:
Closing for now.

PR #604, opened 2022-12-10
Match lbfile

r-burns:


include/dolphin/dvd/dvd.h
```diff
@@ -48,7 +48,6 @@ struct DVDFileInfo
 	/*0x00*/ DVDCommandBlock cb;
     /*0x30*/ u32 startAddr;
     /*0x34*/ u32 length;
-    /*0x38*/ DVDCallback callback;
```
r-burns, line 51: This is needed for some dolphin/dvd stack matches as well, so I think it's correct. I assume this struct def was copied from a different dolphin sdk version.

ribbanya (approved):

PR #606, opened 2022-12-10
Misc MetroTRK labeling

r-burns:


ribbanya (commented):
This is really horrible naming, where is it from?

r-burns:
Haha yeah it really is. Names are taken from the Zoids VS debug map, most functions are an exact match. I doubt we'll even try to match this code, since it has matching issues and isn't actually needed, afaik the TRK is just a leftover debugging tool.

ribbanya (approved):

PR #610, opened 2022-12-16
Misc dolphin matches

r-burns:


ribbanya (approved):
Would be nice to move these over to a shared repo with at least Prime.

r-burns:
> Would be nice to move these over to a shared repo with at least Prime.

Will definitely look into this. Merging this so I can port over more matches from Prime.

PR #614, opened 2022-12-17
Fix Fighter user_data accesses using `GET_FIGHTER` macro

### LINKED ISSUE #545: Replace `fighter_gobj->user_data` with  `GET_FIGHTER` where possible

ProjectRevoTPP:
Just made for tracking.

Resolves #545.


src/melee/ft/types.h
```diff
@@ -1307,6 +1307,27 @@ struct SpecialAttrs_Sandbag {
...
+        u8 x2072_b0_3 : 4;
+        u8 x2072_b4 : 1;
+        u8 x2072_b5 : 1;
+        u8 x2072_b6 : 1;
+        u8 x2072_b7 : 1;
+
+        u8 x2073;
+    };
+    s32 x2070_int;
+};
```
ribbanya, line 1329: I still don't see the point of extracting this into its own type. The `union` replacing a `volatile` variable makes sense, but you should be able to inline this into `Fighter`.

ribbanya (changes requested):
@ProjectRevoTPP Was there anything else you wanted to do with this?

src/melee/ft/types.h
ribbanya: Either that, or this is used a lot by HAL and we should convert more flag members of `Fighter` to use it.

ribbanya: I'm down to just slap a todo on this and merge. What do you think, @r-burns?
```suggestion
/// @todo Rename this and its members; investigate using it elsewhere.
union Struct2070 {
    struct {
        s8 x2070;

        u8 x2071_b0_3 : 4;
        u8 x2071_b4 : 1;
        u8 x2071_b5 : 1;
        u8 x2071_b6 : 1;
        u8 x2071_b7 : 1;

        u8 x2072_b0_3 : 4;
        u8 x2072_b4 : 1;
        u8 x2072_b5 : 1;
        u8 x2072_b6 : 1;
        u8 x2072_b7 : 1;

        u8 x2073;
    };
    s32 x2070_int;
};
```

r-burns: Yeah sounds good to me

r-burns (approved):

ribbanya (approved):

PR #619, opened 2022-12-21
Match `grtkoopa.c`

seizefire:


ribbanya (approved):
Nice work, thanks for the contribution!

PR #623, opened 2022-12-23
Match OSResetSW

r-burns:
From prime decomp


asm/dolphin/os/OSResetSW_data.s
```diff
@@ -1,19 +0,0 @@
-.include "macros.inc"
```
ribbanya, line 1: GitHub doesn't say this was deleted, is it an empty file now?

ribbanya (changes requested):

ribbanya (approved):
Nevermind, it does now.

asm/dolphin/os/OSResetSW_data.s
r-burns: Huh, weird

PR #625, opened 2022-12-25
Fix fighter.c .bss section

r-burns:
Move all fighter HSD_ObjAllocData structs to fighter.c to remove fakematchy pointer arithmetic in Fighter_800679B0

### DIFF
--- a/src/melee/ft/fighter.h
+++ b/src/melee/ft/fighter.h
@@ -2016,6 +2016,7 @@ typedef struct _Fighter {
     /* 0x23B0 */ s32 x23B0;
     /* 0x23B4 */ s32 x23B4;
     /* 0x23B8 */ s32 x23B8;
+    u8 filler_x23B8[0x23E8 - 0x23B8];
 } Fighter;
 
 inline Fighter* getFighter(HSD_GObj* fighter_gobj) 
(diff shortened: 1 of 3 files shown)


ribbanya (changes requested):
I think we can get some other improvements in here, like `sizeof` instead of alloc constants for known structs, as well as https://decomp.me/scratch/hCZ29.

src/melee/ft/fighter.h
```diff
@@ -2120,7 +2121,7 @@ void Fighter_UnkProcessShieldHit_8006D1EC(HSD_GObj* fighter_gobj);
 void Fighter_8006D9AC(HSD_GObj* fighter_gobj);
 void Fighter_UnkCallCameraCallback_8006D9EC(HSD_GObj* fighter_gobj);
 void Fighter_8006DA4C(HSD_GObj* fighter_gobj);
-void Fighter_Unload_8006DABC(Fighter* fp);
+void Fighter_Unload_8006DABC(void* fp);
```
ribbanya, line 2124:
```suggestion
void Fighter_Unload_8006DABC(void* user_data);
```
but whatever

ribbanya (approved):

src/melee/ft/fighter.h
r-burns: Fixed

PR #626, opened 2022-12-26
ft/chara cleanup

r-burns:
* Remove unnecessary casts
* Prefer native bool operators to FALSE comparisons
* Simplify float literals
* Remove unnecessary parentheses

### DIFF
--- a/src/melee/ft/chara/ftMars/ftMarsSpecialN.c
+++ b/src/melee/ft/chara/ftMars/ftMarsSpecialN.c
@@ -38,7 +38,7 @@ void ftMars_SpecialAirN_StartAction(HSD_GObj* gobj) {
 // 80136844 00133424
 // https://decomp.me/scratch/Go6FC
 void lbl_80136844(HSD_GObj* gobj) {
-    if (ftAnim_IsFramesRemaining(gobj) == 0) {
+    if (!ftAnim_IsFramesRemaining(gobj)) {
         func_80136E74(gobj);
         if (((Fighter*)gobj->user_data)->x4_fighterKind == FTKIND_MARS) {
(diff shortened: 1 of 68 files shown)


src/melee/ft/chara/ftMars/ftMarsSpecialN.c
```diff
@@ -318,7 +318,7 @@ void lbl_80137010(HSD_GObj *gobj) {
         //         Fighter_BonePersonalToCommon
         func_8000B1CC(fp->x5E8_fighterBones[func_8007500C(fp, 4)].x0_jobj, 0, &position);
         // AestheticWindEffect(r3=sourcelocation,r4=duration,f1=radiusSize,f2=effectdegradation,f3=unk)
-        func_800119DC(&position, 120, 0.9f, 0.02f, 1.0471975803375244f);
+        func_800119DC(&position, 120, 0.9f, 0.02f, 60 * M_PI / 180);
     }
     // FrameTimerCheck
     if (ftAnim_IsFramesRemaining(gobj) == 0) {
```
ribbanya, line 324: This is a `BOOL` not a frame count, so it's safe to replace it with 
```suggestion
    if (!ftAnim_IsFramesRemaining(gobj)) {
```

Appears quite a lot.

ribbanya (approved):
These issues are everywhere, thanks for starting the cleanup process.

src/melee/ft/chara/ftMars/ftMarsSpecialN.c
r-burns: Oof there are a ton of these. Think I took care of all of them.

PR #627, opened 2022-12-26
Use AT_ADDRESS macro for fixed-address declarations

r-burns:
This macro expands to an appropriate expression depending on the compiler, so these fixed-address declarations can still be used with decomp-permuter and non-MWCC compilers.


ribbanya (approved):
Good idea.

PR #628, opened 2022-12-27
clang-format all existing code

AltimorTASDK:
- Adds a `format` target to the makefile
- Modifies `.clang-format` to not change binary output (no include sorting, no assert formatting)
- Adds `// clang-format off` to all inline asm functions
- Applies new formatting to all source and include files


ribbanya (approved):
I skimmed this, I'm not going to proofread all of it.

PR #629, opened 2022-12-27
Create auto-publishing documentation using Doxygen

ribbanya:
Looks like [this](https://ribbanya.github.io/melee/). A new version of the site is deployed with each commit to `master`. Uses [Doxygen Awesome](https://github.com/jothepro/doxygen-awesome-css) (as a git subtree) for styling.

Includes several custom doc functions, which can be seen here:
https://github.com/doldecomp/melee/blob/6433fc39fe371016b8e0bebfd8190ffad8175758/Doxyfile#L272-L290


r-burns (commented):
Can we have the github action clone doxygen-awesome-css instead of embedding it in our tree? Some of its .png artifacts are rather large. Makes sense as an optional dependency, if it's possible to build locally with the default styling.

Are the only changes to the toplevel Doxyfile the ALIASES helpers? If so, LGTM. If there's more changes, it might be easier to review if the changes are a separate commit from one adding the default Doxyfile as-is.

ribbanya:
Okay, I can see about doing both of those things. Separate commit is easy.

Doxyfile
```diff
@@ -894,56 +913,15 @@ INPUT_ENCODING         = UTF-8
...
-                         *.f18 \
-                         *.f \
-                         *.for \
-                         *.vhd \
-                         *.vhdl \
-                         *.ucf \
-                         *.qsf \
-                         *.ice
+						 *.dox \
+						 *.md
```
r-burns, line 918: nit: mismatched indentation (ALIASES also needs space indentation to align in the github diff view)

r-burns (approved):
Looks great! Lots of useful stuff in the published docs

PR #635, opened 2022-12-30
Enforce 80-character column limit

ribbanya:
Also adds `compile_flags.txt`.


r-burns (approved):
Changes LGTM. Skimmed through some of the reformatted files and the formatter seems to be doing the right thing.

PR #636, opened 2023-01-01
Some ftparts.c decomp and various cleanup + naming

AltimorTASDK:
The diff got messed up because master didn't have make format run after the .clang-format changes.

PR #638, opened 2023-01-02
Fighter cleanup pass 1 by Altimor

ribbanya:
Includes equivalent commits from #636 up to 3cdac6c16058487654f4e9359b926bc1a9b1bbd5.


ribbanya (commented):
[Hiding whitespace](https://github.com/doldecomp/melee/pull/638/files?diff=split&w=1) really helps when reviewing this.

Makefile
```diff
@@ -126,6 +126,7 @@ endif
 # clang-format all source files
 format:
 	$(QUIET) find src include -type f \( -name '*.c' -o -name '*.h' \) -exec $(FORMAT) {} +
+	$(QUIET) tools/newlines.sh
```
ribbanya, line 129: Does this work as expected? I don't actually know how make works at all.

r-burns: You need to sh tools/newlines.sh, or add a shebang to it

ribbanya: Okay, I think it needs to be `#!/bin/bash` so I'll add that.

r-burns: Sounds good, but please use `#!/usr/bin/env bash` for better distro compatibility

r-burns (approved):
Glad I took a closer look at this with whitespace changes hidden, there is some great stuff here. Lots of type refinements in sysdolphin and better names for lots of melee functions. Nice work

PR #640, opened 2023-01-03
Improve dolphin forward declarations

r-burns:
Remove duplicate declarations, and remove declarations from functions.h

Future work: get dolphin code compiling with `-requireprotos`


ribbanya (approved):
This is good as a first pass. In the future I would want to see static functions cleaned up (removed from header files and declared as `static`), and aggregate headers (like `os.h`) split up into their translation units. If we want we can have an `os.h` that `#include`s its individual TU headers, but I would like the function declarations to map 1:1 between `.c`/`.s` and `.h`.

Regarding `-requireprotos`, I don't see the need to turn it on module-by-module, we can just turn it on globally when possible.

PR #642, opened 2023-01-03
Divide `functions.h` into Doxygen modules

ribbanya:
Temporary measure to make removing `functions.h` and `variables.h` easier. Can be viewed at the [Modules](https://ribbanya.github.io/melee/modules.html) page.


r-burns (approved):
Needs a conflict resolution but LGTM

PR #643, opened 2023-01-04
Add editorconfig and github action check

r-burns:
I copied the github action from nixpkgs because
* I am very familiar with it from using nixpkgs
* Nixpkgs is a very large repo so performance of this action will never be an issue
* This action has already gone through many rounds of revisions so it will be more robust than something we roll ourselves

~~There are tons of whitespace issues even with the -disable-indent-size option, so it might be worth doing another reformatting pass separate from this PR. Alternatively we can whitelist existing files, make the check more lenient, etc.~~ All whitespace issues have been resolved in https://github.com/doldecomp/melee/pull/647

Resolves https://github.com/doldecomp/melee/issues/612


.editorconfig
```diff
@@ -0,0 +1,13 @@
+# EditorConfig configuration for doldecomp/melee
+# https://EditorConfig.org
+
+# Top-most EditorConfig file
+root = true
+
+[*]
```
ribbanya, line 7: Ia this going to be a problem for stuff like yml and gas?

r-burns: Not with the -disable-indent-size option. If we remove that, we'll need to reindent the asm files. Let me add a special case for the 2-space indentation of our .yml files.

.editorconfig
```diff
@@ -11,3 +11,6 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 charset = utf-8
+
+[*.yml]
```
ribbanya, line 15:
```suggestion
[*.yml,.clang-format]
```

.editorconfig
r-burns: TBH, there's so much mixed spaces/tabs in the asm that I might just specify in the .editorconfig that either is fine in .s files. LMK what you think

ribbanya: I think we should just replace all tabs with spaces. Either in a separate PR (which merges before this one) or in this PR, I don't really care.

r-burns: Sounds good. I'll issue a separate PR with the whitespace changes.

ribbanya (changes requested):
Do you have an example of this workflow running? I don't see one on [your actions](https://github.com/r-burns/doldecomp-melee/actions).

r-burns:
It only runs on pull requests, so I think it would have to be merged first.

ribbanya:
Would adding a `push: master` trigger hurt anything? Then you can just sync your fork to this PR.

r-burns:
It uses the pull request API to get the changed files so running on a branch wouldn't work, but I tested it on PR to my fork here https://github.com/r-burns/doldecomp-melee/pull/1

ribbanya (approved):

PR #644, opened 2023-01-04
Remove `HSD` library from `functions.h`

ribbanya:


src/melee/gr/ground.c
```diff
@@ -804,6 +804,8 @@ HSD_JObj* func_801C13D0(s32 arg0, s32 depth)
     return result;
 }
 
+#include <sysdolphin/baselib/gobjplink.h>
```
r-burns, line 807: some of these includes are after function declarations, should move these up

ribbanya: I can move them up in HSD. But how do you feel about includes in like, [this thing](https://github.com/doldecomp/melee/blob/dab8f4460545d43bcc03df506177697da88c40ff/src/melee/ft/ft_unknown_006.c) for example? Would you still want them all at the top?

r-burns: hmm didnt realize that file already mixed includes with definitions. I'd be fine leaving files like that as is then, but for anything that isnt like that I'd prefer keeping them at the top. Seems really unusual to not do that always

ribbanya: I agree about confirmed splits, which HSD mostly or entirely consists of. Not so sure about Melee because I mean, just look at Master Hand.

r-burns: oh I see, so the motivation there is that that file is unsplit, so the includes are roughly where they would need to be once it's split? that makes sense then, I'm okay with this for files that will definitely need to be split up later

r-burns (approved):

PR #646, opened 2023-01-05
Match exception runtime functions

r-burns:


src/Runtime/Gecko_ExceptionPPC.c
```diff
@@ -0,0 +1,44 @@
...
+} ProcessInfo;
+
+static ProcessInfo fragment_info[1];
+
+#ifndef NON_MATCHING
+// unused asm function to trigger peephole off bug
+static asm void unused()
+{
+}
+#endif
```
ribbanya, line 16: This works differently than just `#pragma peephole off`?

src/Runtime/__init_cpp_exceptions.c
```diff
@@ -1,46 +1,28 @@
...
-} // clang-format on
+void __init_cpp_exceptions(void)
+{
+    if (fragmentID == -2) {
+        /* HACK: TODO: _eti_init_info_ should be _eti_init_info, we can't use
+         * the appropriate name yet due to the linker not being able to generate
+         * it*/
+        fragmentID = __register_fragment(_eti_init_info_, GetR2());
+    }
+}
```
ribbanya, line 28: I would place this above `_eti_init_info_`'s declaration but Doxygen gets confused by `__declspec` right now.

```suggestion
/**
 * @todo HACK: #_eti_init_info_ should be @c _eti_init_info, we can't use
 * the appropriate name yet due to the linker not being able to generate it
 */
void __init_cpp_exceptions(void)
{
    if (fragmentID == -2)
        fragmentID = __register_fragment(_eti_init_info_, GetR2());
}
```
(nice diff, GitHub)

ribbanya (changes requested):

src/Runtime/Gecko_ExceptionPPC.c
r-burns: No it works the same, just felt it lines up with the original source better than a bare pragma.

ribbanya (approved):

PR #649, opened 2023-01-05
Remove remaining usages of `#pragma once`

ribbanya:


r-burns (approved):
Haven't tested it, but the clang-tidy `llvm-header-guard` check seems like it would enforce this

PR #650, opened 2023-01-05
Fix `deprecated-non-prototype` warnings, pass 1

ribbanya:
Working toward `-requireprotos`.


obj_files.mk
```diff
@@ -581,7 +581,7 @@ TEXT_O_FILES +=\
     $(BUILD_DIR)/asm/dolphin/mtx/vec.s.o\
 
 TEXT_O_FILES +=\
-    $(BUILD_DIR)/src/dolphin/os/OS.c.o\
+    $(BUILD_DIR)/src/dolphin/os/OSInit.c.o\
```
r-burns, line 584: OS.c is the correct file name as provided by debug maps

r-burns (changes requested):

src/sysdolphin/baselib/aobj.c
```diff
@@ -259,45 +259,47 @@ void HSD_AObjFree(HSD_AObj* aobj)
 }
 
 void callbackForeachFunc(struct _HSD_AObj* aobj, void* obj, HSD_Type type,
-                         void (*func)(), AObj_Arg_Type arg_type,
-                         callbackArg* arg)
+                         void* func, AObj_Arg_Type arg_type, callbackArg* arg)
```
r-burns, line 262: This seems like a case where not specifying the argument types is cleaner. Requiring the explicit casts here adds no useful information

obj_files.mk
ribbanya: Well that's fair but we use wine to compile so something has to be done about `os.h` if we want `OS.h`.

src/sysdolphin/baselib/aobj.c
ribbanya: A small price to pay imo.

ribbanya: Just to clarify, `-Wstrict-prototypes` on clang is a warning but `-requireprotos` does not have any flexibility to it. MWCC will abort if it comes across non-compliant code with that option enabled. So I agree the casts suck but I don't think code like that outweighs the benefits.

obj_files.mk
r-burns: Hmm that's true. I suppose we could do dolphin/os/OS.h for this TU and dolphin/os.h as the module catchall, but that would require restructuring.

ribbanya: Long-term all of `dolphin` is being outsourced, so personally I don't have a problem with naming this wrong temporarily. But I leave the decision up to you, I was just trying to get it to compile.

src/sysdolphin/baselib/aobj.c
r-burns: `void (*func)()` is compliant with -requireprotos though?

ribbanya: I can double-check but I'm pretty sure it's not.

r-burns: Oh I get the issue, `void (*func)()` isn't C99 compliant so you're having trouble using it with clang-tools. That's probably gonna be an issue for a lot of modern tooling. I can deal with the ugliness in that case

r-burns (approved):

src/sysdolphin/baselib/aobj.c
ribbanya: Actually no, I elevated the warning to an error in `compile_flags.txt` _because_ `-requireprotos` doesn't allow empty parentheses.

This was like a month ago though so I'm not 100% sure anymore.

r-burns: Well I just tested it and does compile with -requireprotos, but a stricter -Werror=strict-prototypes flag is probably a net improvement.

ribbanya: I see, good to know, thanks.

PR #651, opened 2023-01-05
Create `.vimrc`

ribbanya:
Doesn't enforce anything, just makes Vim look better.


.vimrc
```diff
@@ -0,0 +1,15 @@
+set tabstop=4
+set softtabstop=4
+set shiftwidth=4
+set noexpandtab
```
r-burns, line 4:
```suggestion
```

I recommend editorconfig-vim, which will do this automatically now that we have a .editorconfig.
It will also use the correct 2-space indent for .yml and .clang-format files.

r-burns (changes requested):
Since you said you're sourcing this manually, let's put this in a subdirectory such as tools/.

.vimrc
ribbanya: Thanks, I'll look into it.

r-burns (approved):

PR #653, opened 2023-01-06
Fix `implicit-function-declaration` warnings, passes 1..3

ribbanya:


src/placeholder.h
```diff
@@ -0,0 +1,9 @@
+#ifndef PLACEHOLDER_H
+#define PLACEHOLDER_H
+
+#include <dolphin/os/os.h>
+
+#define NOT_IMPLEMENTED(name)                                                  \
+    OSPanic(__FILE__, __LINE__, "#name is not implemented!");
```
ribbanya, line 7: @r-burns Do you think `__func__` would work instead of `name`? I don't know how standard it is or if there are any gotchas.

r-burns: Yeah I think that would work great. You just have to be careful as it's not really a macro https://stackoverflow.com/questions/31430765/why-func-function-and-pretty-function-arent-preprocessor-macros and detecting support is a little different https://stackoverflow.com/questions/15305310/predefined-macros-for-function-name-func

ribbanya: So like this?
```suggestion
    OSPanic(__FILE__, __LINE__, "%s is not implemented!", __func__);
```

r-burns: Yeah exactly

src/dolphin/card/CARDCreate.c
```diff
@@ -112,9 +122,12 @@ s32 CARDCreateAsync(s32 chan, const char* fileName, u32 size,
     fileInfo->chan = chan;
     fileInfo->fileNo = freeNo;
 
-    result = __CARDAllocBlock(chan, size / card->sectorSize, CreateCallbackFat);
-    if (result < 0) {
+    /// @todo Eliminate cast to #CARDCallback.
+    result = __CARDAllocBlock(chan, size / card->sectorSize,
+                              (CARDCallback) CreateCallbackFat);
```
r-burns, line 127: Seems like you can remove these CARDCallback casts if you keep propagating EXIChannel to these function definitions

src/dolphin/os/OSAlarm.c
```diff
@@ -24,9 +24,10 @@ void OSInitAlarm(void)
     }
 }
 
-void OSCreateAlarm(s32* arg0)
+void OSCreateAlarm(OSAlarm* alarm)
 {
-    *arg0 = 0;
+    /// @todo Hack.
+    *((s32*) alarm) = 0;
```
r-burns, line 30:
```suggestion
    alarm->handler = NULL;
```

src/dolphin/os/OSExi.c
```diff
@@ -111,17 +114,21 @@ BOOL EXIImmEx(EXIChannel chan, void* buf, s32 len, u32 mode)
...
 
-        if (!EXISync(chan)) {
+        if (!EXISync(chan))
             return FALSE;
-        }
 
+/// @todo what
+#if defined(__MWERKS__) && defined(GEKKO)
         (u8*) buf += xLen;
+#endif
```
r-burns, line 127:
```suggestion
        buf = (u8*) buf + xLen;
```

src/melee/ft/chara/ftCLink/ftclink.h
```diff
@@ -4,6 +4,7 @@
 #include <melee/ft/fighter.h>
 #include <sysdolphin/baselib/gobj.h>
 
+/// @todo this is #ftLinkAttributes.
```
r-burns, line 7: do you mean ftCLinkAttributes? Or are they the same for both

include/string.h
```diff
@@ -3,7 +3,6 @@
 
 #include <stddef.h>
 
-void memcpy(void* dst, void* src, size_t num);
```
r-burns, line 6: Why'd this get deleted?

r-burns (commented):
+1 to your plans to remove all the redundant typedef aliases and delete include/.

include/dolphin/types.h
```diff
@@ -64,11 +64,17 @@ typedef jmp_t jtbl_t[];
...
+
 #elif defined(__MWERKS__)
+
 #define AT_ADDRESS(x) : (x)
+
 #else
+
 #define AT_ADDRESS(x)
+
 #endif
```
ribbanya, line 78: @r-burns This, right?
```suggestion
#define AT_ADDRESS(x) = FIXEDADDR(x)
#elif defined(__MWERKS__)
#define AT_ADDRESS(x) : (x)
#else
#define AT_ADDRESS(x)
#endif
```

include/string.h
ribbanya: Was a duplicate of something, either `Runtime/__mem.h` or `cstring.h`.

src/melee/ft/chara/ftCLink/ftclink.h
ribbanya: Yeah, they are the same struct. Same applies to all clone characters it seems like (and Nana).

src/dolphin/os/OSExi.c
ribbanya: Did you compile this? I think I tried it.

src/dolphin/card/CARDCreate.c
ribbanya: Yeah I know, just seemed out of scope for this PR.

include/dolphin/types.h
r-burns: yeah lol

include/string.h
r-burns: well I think it belongs here, it's C standard and <cstring> is a C++ thing

src/melee/ft/chara/ftCLink/ftclink.h
r-burns: Interesting, thanks

src/dolphin/card/CARDCreate.c
r-burns: Ok that's fair

src/dolphin/os/OSExi.c
r-burns: Yeah it works

include/string.h
ribbanya: Then our splits are wrong and it needs to be addressed at the TU level, not the header level.

src/dolphin/os/OSExi.c
ribbanya: It also just does nothing as far as I can tell...

include/string.h
r-burns: Nah where it's implemented is arbitrary, it just needs to be made available from this header. So just have this header include `<Runtime/__mem.h>` and I think we're good.

ribbanya: > So just have this header include `<Runtime/__mem.h>` and I think we're good.

Okay.

src/dolphin/os/OSExi.c
r-burns: It offsets buf by xLen bytes for the next iteration.

ribbanya: Oh right, duh.

r-burns (approved):

r-burns (approved):

PR #654, opened 2023-01-08
Fix `implicit-function-declaration` warnings, pass 4

ribbanya:


Makefile
```diff
@@ -92,11 +92,11 @@ LDFLAGS := -fp hard -nodefaults
 ifeq ($(GENERATE_MAP),1)
 	LDFLAGS += -map $(MAP)
 endif
-
 CFLAGS  = -msgstyle gcc \
 		-cwd source \
 		-Cpp_exceptions off \
 		-DGEKKO \
+		-DMWERKS_GEKKO \
```
ribbanya, line 99: Just a shorthand for `__MWERKS__ && GEKKO`, regardless of whether the target is meant to match.

Makefile
```diff
@@ -105,7 +105,9 @@ CFLAGS  = -msgstyle gcc \
 		-maxerrors $(MAX_ERRORS)
 
 ifeq ($(NON_MATCHING),1)
-CFLAGS += -DNON_MATCHING
+	CFLAGS += -DNON_MATCHING
+else
+	CFLAGS += -DMUST_MATCH
```
ribbanya, line 110: Both imply `__MWERKS__ && GEKKO`, but differ in whether the target must match. Honestly I'd like to phase out `NON_MATCHING` because by definition it implies a target architecture and compiler, but semantically any build on other platforms would be "non-matching." Now that we care about target platform, `!NON_MATCHING` no longer equals `MUST_MATCH`.

compile_flags.txt
```diff
@@ -6,7 +6,6 @@
 -Iinclude
 -Wall
 -Wextra
--Wno-unknown-pragmas
```
ribbanya, line 9: Would prefer to wrap all platform-specific `#if`s than just ignore the warning.

include/dolphin/os/OSReset.h
```diff
@@ -17,12 +17,9 @@ void OSResetSystem(BOOL reset, u32 resetCode, BOOL forceMenu);
 void OSRegisterResetFunction(OSResetFunctionInfo* func);
 void __OSDoHotReset(s32 arg0);
 
-/// @defgroup OS_RESET_CODE OSReset Code
```
ribbanya, line 20: This is a Doxygen module thing, I need to figure out how they work still.

src/melee/ft/chara/ftFox/ftFox_SpecialN.c
```diff
@@ -153,40 +153,48 @@ void ftFox_RemoveBlaster(HSD_GObj* fighter_gobj)
...
     Fighter* fp;
     f64 launchAngle;
     s32 ftKind;
 
+#ifdef MUST_MATCH
+    /// @todo Seems fake, probably one or more missing @c inline functions.
     fp = fp = getFighter(fighter_gobj);
+#else
+    fp = getFighter(fighter_gobj);
+#endif
```
ribbanya, line 178: This only exists because the original line generates a warning in `clang`. I think it's pretty reasonable to aim for no warnings/errors.

src/melee/ft/chara/ftFox/ftfox.h
```diff
@@ -181,8 +181,13 @@ void ftFox_AppealS_OnTakeDamage(HSD_GObj* fighter_gobj);
 
 // Neutral Special - Blaster (SpecialN) //
 
+/// Get Fox/Falco's Item Hold Bone Position for Blaster GFX
```
ribbanya, line 184: In general, briefs (the short description of a member) were moved to headers away from the implementation. Doxygen will take both at once, but it's generally considered better to put the usage information in the header and the implementation details in the C file.

src/melee/ft/chara/ftGameWatch/ftGameWatch_SpecialN.c
```diff
@@ -1,15 +1,11 @@
...
+#include <melee/ft/code_80081B38.h>
+#include <melee/it/code_8027CF30.h>
 #include <melee/ft/ftparts.h>
 #include <melee/lb/lbunknown_001.h>
 
-/* https://decomp.me/scratch/P7Z9j - first function of SpecialN.c; terrifying
-   code https://decomp.me/scratch/kFkx6 - closer match, but just what is going
-   on in this one? temp_fp suggests there is an inline somewhere in the
-   function, but an inline depth of 2 causes the compiler to start using branch
-   link; no combination of inlines / lack thereof I have tried fully match */
```
ribbanya, line 11: Moved down in the file.

src/melee/ft/chara/ftGameWatch/ftGameWatch_SpecialN.c
```diff
@@ -25,13 +21,117 @@ f32 ftGameWatch_SpecialN_SetFloatOrder3(void)
     return 0.0f;
 }
 
-extern HSD_GObj* func_802C837C(HSD_GObj*, Vec3*, s32, u32, f32);
+#ifdef MUST_MATCH
```
ribbanya, line 24: `#if`/`#else` order swapped. In general I think it reads better if the matching implementation appears first.

src/melee/ft/chara/ftGameWatch/ftgamewatch.h
```diff
@@ -204,8 +204,6 @@ void ftGameWatch_AttackS4_Coll(HSD_GObj* fighter_gobj);
 void ftGameWatch_ItemManholeSetup(HSD_GObj* fighter_gobj);
 void ftGameWatch_ItemManholeRemove(HSD_GObj* fighter_gobj);
 void ftGameWatch_ItemManholeOnDamage(HSD_GObj* fighter_gobj);
-void ftGameWatch_ItemManholeEnterHitlag(HSD_GObj* fighter_gobj);
-void ftGameWatch_ItemManholeExitHitlag(HSD_GObj* fighter_gobj);
```
ribbanya, line 208: `static`s removed.

src/melee/ft/chara/ftGigaKoopa/ftgigakoopa.c
```diff
@@ -78,9 +77,9 @@ void func_8014F98C(s32 arg0, s32* arg1, s32* arg2)
     }
 }
 
-extern const s32 lbl_803D3984[];
+/* static */ extern const s32 lbl_803D3984[];
 
-// UB warning: this function may use offset uninitialized
+/// @remarks UB warning: this function may use offset uninitialized
```
ribbanya, line 82: No idea what "UB" stands for.

src/melee/ft/chara/ftIceClimber/fticeclimber2_nana.c
```diff
@@ -1,9 +1,7 @@
 #include <melee/ft/chara/ftIceClimber/fticeclimber.h>
-
 #include <melee/ft/fighter.h>
+#include <melee/ft/ftparts.h>
 
-/// TODO remove float externs when file is finished and float data is removed
-/// from asm
```
ribbanya, line 6: Kind of just self-evident imo.

src/melee/ft/chara/ftKirby/ftkirby.c
```diff
@@ -116,62 +127,72 @@ void ftKirby_800EE7B8(HSD_GObj* fighter_gobj)
...
     func_800F9090(fighter_gobj);
     func_800F1A8C(fighter_gobj);
-    fp->cb.x21E0_callback_OnDeath = 0;
+    fp->cb.x21E0_callback_OnDeath = NULL;
+}
+
+/// @file
+/// @todo Matching, but needs more data moved over for DOL match
+
+#if FALSE
```
ribbanya, line 136: Commented (matching) code → `#if FALSE`.

src/melee/ft/chara/ftLuigi/ftLuigi_SpecialS.c
```diff
@@ -1,18 +1,21 @@
+#include "sysdolphin/baselib/gobjproc.h"
```
ribbanya, line 1: Auto-added by vim and I'm too lazy to fix it right now.

src/melee/ft/chara/ftMario/ftMario_SpecialLw.c
```diff
@@ -274,8 +275,9 @@ void _ftMario_800E25C4_800E2778_helper(HSD_GObj* gobj)
     }
 }
 
-static ftCollisionBox ftMario_SpecialLw_CollisionBox = { 12.0F, 0.0F, -6.0F,
-                                                         6.0F,  6.0F, 6.0F };
+static ftCollisionBox ftMario_SpecialLw_CollisionBox = {
+    12.0F, 0.0F, { -6.0F, 6.0F }, { 6.0F, 6.0F }
+};
```
ribbanya, line 280: `clang` was emitting a warning for not grouping the sub-structures. Can be fixed automatically via `clang-tidy`/LSP so I think it's fine to comply.

src/melee/ft/chara/ftMasterHand/dir.dox
```diff
@@ -0,0 +1,3 @@
+/** @dir
+ * @defgroup ftmasterhand ftMasterHand
+ */
```
ribbanya, line 3: Again, don't really get how Doxygen modules work at the moment. I'll figure it out in a future PR.

src/melee/ft/chara/ftMasterHand/ftMasterHand_5.c
```diff
@@ -1,17 +1,19 @@
...
-// 80151428 14E008
-// https://decomp.me/scratch/mNdcr
-// probably isCrazyHandLoaded func
 u32 lbl_80151428(Vec* vec)
 {
     HSD_GObj* fetched_gobj;
     // Fighter* ft_userdata;
 
-    fetched_gobj =
-        func_8015C3E8(0x1B); // probably attempt to find crazy hand func.
```
ribbanya, line 14: Moved to brief of `func_8015C3E8`.

src/melee/ft/chara/ftMewtwo/ftMewtwo_SpecialHi.c
```diff
@@ -393,13 +402,12 @@ inline void ftMewtwo_SpecialHi_SetVars(HSD_GObj* fighter_gobj)
...
 }
 
-extern BOOL func_8009A134(HSD_GObj*);
-
 // 0x80145990
 // https://decomp.me/scratch/kFSF2 // Mewtwo's grounded Teleport Zoom Action
 // State handler
 void ftMewtwo_SpecialHi_Action(HSD_GObj* fighter_gobj)
 {
+    /// @todo Shared @c inline with #ftMewtwo_SpecialAirHi_Action.
```
ribbanya, line 410: There are SO MANY of these. I gave up on marking them at some point. And they're really hard to detect automatically, I think.

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
```diff
@@ -353,22 +342,18 @@ void ftNess_YoyoApplySmash(
...
-    }
+
     func_8007EBAC(fp, 0x24U, 0U);
     func_80088510(fp, 0x334A7, 0x7F, 0x40);
 }
 
-extern void func_802C0010(HSD_GObj*, void*); /* extern */
-
-const Vec3 YoyoChargePos = { 0.0f, 0.0f, 0.0f };
+static Vec3 const YoyoChargePos = { 0 };
```
ribbanya, line 353: I think this looks better for zeroing structs but let me know what you think.

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
```diff
@@ -383,21 +368,26 @@ void ftNess_YoyoSetChargeDamage(
...
-    fp->nessVars[0].AttackHi4.yoyoCurrentFrame =
-        14; // Current animation frame, integer //
+
+    // Current animation frame, integer
+    fp->nessVars[0].AttackHi4.yoyoCurrentFrame = 14;
+
     yoyo_GObj = fp->sa.ness.x222C_yoyoGObj;
-    if (yoyo_GObj != NULL) {
+
+    if (yoyo_GObj) {
```
ribbanya, line 377: I'm so used to looking at C# code that I normally lean toward `!= NULL`, but based on asserts I think HAL always does it this way. What do you think we should use?

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
```diff
@@ -737,30 +693,28 @@ void ftNess_AttackHi4_Anim(
...
 
-extern void func_8008A4D4(HSD_GObj*);
-
-// 0x80115E74 //
-// https://decomp.me/scratch/0Hcrg //
+// 0x80115E74
+// https://decomp.me/scratch/0Hcrg
 void ftNess_AttackHi4_IASA(
-    HSD_GObj* fighter_gobj) // Ness's Up Smash IASA callback //
+    HSD_GObj* fighter_gobj) // Ness's Up Smash IASA callback
```
ribbanya, line 699: Gave up on fixing these comments. Will do it in bulk later.

src/melee/ft/chara/ftPikachu/ftpikachu4.c
```diff
@@ -280,7 +280,10 @@ void ftPikachu_80126614(HSD_GObj* fighter_gobj)
 
     ftPikachuAttributes* pika_attr = fp->x2D4_specialAttributes;
 
-    if (!func_80082888(fighter_gobj, &pika_attr->height_attributes)) {
+    /// @todo Eliminate cast (by changing type of field)
+    if (!func_80082888(fighter_gobj,
+                       (ftCollisionBox*) &pika_attr->height_attributes))
```
ribbanya, line 285: These are either `f32[6]` or `ftCollisionBox*` depending on where you are in the codebase.

src/melee/ft/fighter.h
```diff
@@ -893,15 +890,6 @@ typedef struct _FighterBone {
...
 
-typedef struct _Hitbox {
-    s32 x0;
-    s32 x4;
-    s32 x8;
-    f32 xC;
-    u8 filler[0x128]; // slightly more detail, f32 needed for 'manual' mods to
-                      // hitbox damage //
-} Hitbox;
-
```
ribbanya, line 904: Moved to `lbcollision` (or something).

Makefile
r-burns: This can be done with
```c
#if defined(__MWERKS__) && defined(GEKKO)
#define MWERKS_GEKKO
#endif
```
in a header somewhere

ribbanya: I did that at first but `clang` (and I imagine other compilers) complain that "expanding `defined() in a preprocessor directive is undefined behavior."

ribbanya: Oh wait, nevermind. That would work. Is it better?

Makefile
r-burns: But `!NON_MATCHING` is literally `MUST_MATCH`, that's just how it works. If you're targeting any other platform then your build is not matching.

ribbanya: The point I was trying to make is only the MWCC script is defining `NON_MATCHING`. When there's a GCC build, would you expect it to defined `NON_MATCHING`? I'd rather it not have to.

src/melee/ft/chara/ftGigaKoopa/ftgigakoopa.c
r-burns: Undefined Behavior, in the context of the C/C++ standard

Makefile
r-burns: A GCC build will not be matching, so yeah. I could see replacing all NON_MATCHING with !MUST_MATCH, but having both is redundant.

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
r-burns: I also like the explicit NULL comparison and so does Psi based on the style guide. It can't be used in asserts but otherwise I don't think it ever changes codegen. So I'd say keep it.

Makefile
ribbanya: There's just a lot of code (in this PR but also going forward) where you just want `#if defined(MWERKS_GEKKO) && !defined(NON_MATCHING)`. So I don't really like `NON_MATCHING` anymore.

Replacing all of them seems a bit out of scope, but I can do it now if you prefer.

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
ribbanya: Okay, will revert.

Makefile
r-burns: To me !defined(NON_MATCHING) requires MWERKS_GEKKO. You could enforce this in a header with
```c
#if !defined(NON_MATCHING) && !defined(MWERKS_GEKKO)
#error "A matching build requires MWerks Gekko support"
#endif
```

ribbanya: idk seems kind of hacky compared to just getting rid of it imo.

r-burns: Getting rid of what?

ribbanya: Getting rid of `NON_MATCHING` and then there's just `MWERKS_GEKKO` (to access MWCC asm/features regardless of match) and `MUST_MATCH` for stuff that's only in the code because it needs to match.

I think `!MUST_MATCH` covers all the existing use cases of `NON_MATCHING`, which is basically for partial matches.

ribbanya: The key difference being, that GCC (and Permuter, and clang, and M2C, etc etc) don't need to define `NON_MATCHING` or worry about an error. The symbols are all opt-in instead of declaring a negative.

r-burns: I think my main objection to seeing this was seeing two flags that are defined as opposites of each other, i.e. what happens when you compile with -DNON_MATCHING -DMUST_MATCH. There should be a single one that you provide (or don't provide) to the preprocessor, and the other is determined automatically via ifdefs.

After that, I suppose the choice of whether that flag is NON_MATCHING or MUST_MATCH is arbitrary. But every other decomp project uses NON_MATCHING, so I think we should stick to that.

ribbanya: I see, well in general I don't really know (or care tbh) what other decomp projects are doing.

But anyway, in this case, if we agree that `NON_MATCHING` means "uses the original hardware and compiler but happens to not match" then we can continue to use that.

Makefile
r-burns: Yeah I think in general it's a good idea to push this kind of logic out of the makefile and into the sources

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
```diff
@@ -129,27 +122,28 @@ s32 ftNess_YoyoCheckEnvColl(
...
         retval |= 0x2000;
-    }
+
     return retval;
 }
 
+#ifdef MUST_MATCH
+#pragma push
 #pragma dont_inline on
+#endif
```
r-burns, line 145: Would be nice if there were a way to make a MATCHING_DONT_INLINE macro for this. Modern compilers have a `_Pragma` expression that can expand from macros but I don't think it works with MWCC.

r-burns (changes requested):
Looks good aside from the redundant NON_MATCHING/MUST_MATCH C defines. Based on our discussion I think we settled on using NON_MATCHING in the makefile and switching to MUST_MATCH only in C code.

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
r-burns: (not a request for change, just a side remark)

src/melee/ft/chara/ftNess/ftNess_AttackHi4.c
r-burns: Yeah I believe ` = { 0 }` is a common idiom in C to zero a struct. Looks good

r-burns (approved):

PR #656, opened 2023-01-10
Fix `implicit-function-declaration` warnings, pass 6

ribbanya:
Finishes the rest of `ft` and `gm`.

### DIFF
--- a/src/melee/ft/ftparts.c
+++ b/src/melee/ft/ftparts.c
@@ -14,7 +14,9 @@
 #include <sysdolphin/baselib/pobj.h>
 #include <sysdolphin/baselib/util.h>
 
-#define MAX_FT_PARTS 140
+#include <dolphin/os/os.h>
+
+#define MAX_FT_PARTS (140)
 
 #define JOBJ_NEXT(jobj) ((jobj) == NULL ? NULL : (jobj)->next)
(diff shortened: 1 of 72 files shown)


src/melee/ft/ftparts.c
```diff
@@ -14,7 +14,9 @@
 #include <sysdolphin/baselib/pobj.h>
 #include <sysdolphin/baselib/util.h>
 
-#define MAX_FT_PARTS 140
+#include <dolphin/os/os.h>
+
+#define MAX_FT_PARTS (140)
```
r-burns, line 19:
```suggestion
#define MAX_FT_PARTS 140
```
nit, parens not necessary for literals

r-burns (approved):

src/melee/ft/ftparts.c
ribbanya: Really? There are no adverse edge cases?

r-burns: Nope, parens are only necessary when you are expanding an arg in a function-like macro or when you are expanding to multiple tokens

ribbanya: Ok well I'm not motivated to fix it in bulk but I'll start phasing out the parens for literals.

PR #657, opened 2023-01-10
Clean up `dolphin/types.h` and change empty brace formatting

ribbanya:
* Moved most stuff in `dolphin/types.h` to `platform.h`
  * `dolphin/types.h` can actually be removed, it just has deprecated macros now
* Added doc comments for almost everything in `platform.h`
* Created `enum_t`
  * Equivalent to `int`
  * Used for when we know something is an enum but haven't made the enum yet
  * Is a lot more searchable than `int` or `s32`
  * Platform-independent
* Created `any_t`
  * Equivalent to `void*`
  * Used to differentiate from `unk_t`
  * Eventually, when there are no more `unk_t`, can be easily find/replaced with `void*`
  * Example use cases are HSD `user_data` and memory allocations

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,7 +6,7 @@
   TabWidth: 4,
   BreakBeforeBraces: Custom,
   BraceWrapping: { AfterFunction: true, AfterControlStatement: MultiLine },
-  AllowShortFunctionsOnASingleLine: false,
+  AllowShortFunctionsOnASingleLine: Empty,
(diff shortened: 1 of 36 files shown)


ribbanya:
Forgot to mention that the codegen for `int` (`enum_t`) is different than for `s32`, which is being used almost everywhere as the `enum`-standin right now. `lblanguage` is one example of it mattering, and I believe stuff in `pl` is the same case.

PR #658, opened 2023-01-11
Fix `implicit-function-declaration` warnings, pass 7

ribbanya:
`ground.c` took quite a while...


include/common_structs.h
```diff
@@ -38,7 +38,7 @@ struct DemoMotionFiles {
     struct DemoMotionSymbols* x0_demoMotionPtr[14];
 };
 
-typedef Vec Vec3;
+typedef Vec3 Vec3;
```
ribbanya, line 41: This slipped through my renames, but I'll get to it during my header cleanup.

src/melee/ft/chara/ftMario/ftMario_SpecialHi.c
```diff
@@ -22,7 +22,7 @@ void ftMario_SpecialAirHi_StartAction(HSD_GObj* gobj)
 {
     Fighter* fp;
     ftMarioAttributes* sa;
-    u8 unused[4];
+    char unused[4];
```
ribbanya, line 25: The rationale for `u8` to `char` here is a little hard to explain but ask me on Discord if needed.

src/sysdolphin/baselib/lobj.h
```diff
@@ -1,27 +1,26 @@
...
+#define HSD_A_L_A0 (13)
+#define HSD_A_L_A1 (14)
+#define HSD_A_L_A2 (15)
+#define HSD_A_L_K0 (16)
+#define HSD_A_L_K1 (17)
+#define HSD_A_L_K2 (18)
+#define HSD_A_L_CUTOFF (19)
+#define HSD_A_L_REFDIST (20)
+#define HSD_A_L_REFBRIGHT (21)
+#define HSD_A_L_LITC_A (22)
```
ribbanya, line 23: Yeah sorry, will fix these at some point.

include/sysdolphin/baselib/memory.h
```diff
@@ -1,9 +1,10 @@
 #ifndef _memory_h_
 #define _memory_h_
 
+#include "Runtime/platform.h"
 #include <dolphin/types.h>
 
-void HSD_Free(void* ptr);
-void* HSD_MemAlloc(s32 size);
+void HSD_Free(any_t ptr);
```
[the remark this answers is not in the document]
r-burns: Discussed this in more detail, we'll keep these as markers during decomp and convert them back to void* when there are no problematic uses of void*. Marking as resolved

src/melee/ft/chara/ftMario/ftMario_SpecialHi.c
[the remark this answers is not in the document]
r-burns: Per discussion this was motivated by m2c having special handling for this syntax. I'm going to try to have m2c recognize this `u8` pattern as well.

src/melee/gr/ground.c
```diff
@@ -2646,6 +2848,19 @@ static asm void func_801C34AC(s32, HSD_JObj*, HSD_Joint*)
...
+static void func_801C34AC(s32 arg0, HSD_JObj* arg1, HSD_Joint* arg2)
+{
+    NOT_IMPLEMENTED;
+}
+
+#endif
+
+#ifdef MWERKS_GEKKO
+
+#ifdef MWERKS_GEKKO
```
ribbanya, line 2862: Woops, will fix this.

src/sysdolphin/baselib/initialize.c
```diff
@@ -264,7 +264,7 @@ static void HSD_ObjInit(void)
     HSD_ZListInitAllocData();
 }
 
-static char unused[] = "pix_fmt != GX_PF_RGB565_Z16";
+static u8 unused[] = "pix_fmt != GX_PF_RGB565_Z16";
```
r-burns, line 267:
```suggestion
static char unused[] = "pix_fmt != GX_PF_RGB565_Z16";
```
this one really is a char array ;)

src/sysdolphin/baselib/memory.c
```diff
@@ -1,16 +1,18 @@
...
+#include <sysdolphin/baselib/initialize.h>
 
-void HSD_Free(void* ptr)
+void HSD_Free(any_t ptr)
 {
     OSFreeToHeap(HSD_GetHeap(), ptr);
 }
 
-void* HSD_MemAlloc(s32 size)
+any_t HSD_MemAlloc(ptrdiff_t size)
```
r-burns, line 13:
```suggestion
any_t HSD_MemAlloc(ssize_t size)
```

Should definitely be `size_t` but doesn't match because it needs to be signed. Let's add a `ssize_t` typedef, since that's probably the closest to correct.

src/sysdolphin/baselib/initialize.c
ribbanya: :see_no_evil:

src/sysdolphin/baselib/memory.c
ribbanya: That's literally what [`ptrdiff_t`](https://en.cppreference.com/w/cpp/types/ptrdiff_t) is as far as I understand it.

r-burns: They're pretty much always the same underlying type, but a ptrdiff is a difference of pointers and a size_t counts a number of bytes, so it makes more sense to pass a ssize_t to an allocator IMO

r-burns (approved):

PR #665, opened 2023-01-11
Fix `implicit-function-declaration` warnings, pass 9

ribbanya:


include/Runtime/platform.h
```diff
@@ -112,6 +112,16 @@ typedef void (*Event)(void);
 #endif
 #endif
 
+#ifndef ATTRIBUTE_NORETURN
+#if defined(__MWERKS__) || defined(__GNUC__)
+#define ATTRIBUTE_NORETURN
```
r-burns, line 117: This should be supported on GCC

src/melee/gr/grdynamicattr.c
```diff
@@ -32,11 +33,16 @@ void func_801CA0B4(void)
     }
 }
 
-UnkStruct* func_801CA0F8(s32 arg0, Vec3* v, s32 floor_id, f32 f, s32 arg3)
+UnkStruct* func_801CA0F8(s32 arg0, Vec3* v, enum_t floor_id, f32 f, s32 arg3)
 {
+#ifdef MUST_MATCH
+    u8 unused[8];
```
r-burns, line 39: Why did the padding grow?

src/dolphin/os/OSAlloc.c
```diff
@@ -233,7 +233,8 @@ size_t OSCheckHeap(OSHeapHandle heap)
         CHECK(922, cell->next == NULL || cell->next->prev == cell)
         CHECK(923, MINOBJSIZE <= cell->size)
         CHECK(924, OFFSET(cell->size, ALIGNMENT) == 0)
-        CHECK(925, cell->next == NULL || (char*) cell + cell->size < (char*) cell->next)
+        CHECK(925, cell->next == NULL ||
+                       (char*) cell + cell->size < (char*) cell->next)
```
r-burns, line 237: How does the newline not mess up the assert string? And shouldn't the setting you added to .clang-format prevent this from getting formatted?

include/Runtime/platform.h
```diff
@@ -112,6 +112,16 @@ typedef void (*Event)(void);
...
+#elif defined(__clang__)
+#define ATTRIBUTE_NORETURN __attribute__((noreturn))
+#else
+#error unknown compiler
```
r-burns, line 121: Let's not error here, the fallback should be to define it to nothing and continue

src/dolphin/os/OSAlloc.c
ribbanya: I noticed the same thing. No idea, but it matches so I'm not complaining.

src/melee/gr/grdynamicattr.c
ribbanya: It didn't, I've been converting them all to `u8` because it's quicker to read.

r-burns: Ah of course

r-burns (approved):

PR #669, opened 2023-01-12
Fix `implicit-function-declaration` warnings, pass 10

ribbanya:


src/melee/ft/fighter.h
```diff
@@ -44,135 +45,6 @@ extern void* lbl_804D6548;
 extern void* lbl_804D654C;
 extern void* lbl_804D6550;
 
-typedef enum FighterKind {
```
ribbanya, line 47: Moved to `ft/types.h` to avoid circular dependencies.

src/melee/pl/player.h
```diff
@@ -62,16 +71,18 @@ typedef struct _StaticPlayer {
 
     /*0x68 - 0x6C*/ s32 falls[2]; /// other index for nana falls
 
-    /*0x70-0x84*/ s32 kos_by_player[6];
+    /*0x70-0x84*/ uint kos_by_player[6];
```
ribbanya, line 74: `u32` will not match but this is used in unsigned comparisons. In my opinion, further proof that HAL's default data type is `int` and not `long`.

src/stddef.h
```diff
@@ -1,7 +0,0 @@
-#ifndef __STDDEF_H__
```
ribbanya, line 1: Moved to `MSL`.

src/melee/lb/lbcommand.c
```diff
@@ -77,7 +77,7 @@ void Command_07(CommandInfo* info)
 void Command_08(CommandInfo* info)
 {
     info->u.data_position += 1;
-    info->timer = FLT_MAX;
+    info->timer = F32_MAX;
```
r-burns, line 80:
```suggestion
    info->timer = FLT_MAX;
```
FLT_MAX is in the C standard

ribbanya: I hear you, but we're in a weird spot because we need ports to have the same behavior and `FLT_MAX` does not guarantee a particular value (it happens to be the 32-bit IEEE max on most platforms). I will revert it if you insist but I just want to guarantee that this value is what HAL used, without relying on literals.

src/string.h
```diff
@@ -1,8 +1,8 @@
 #ifndef __STRING_H__
 #define __STRING_H__
 
+#include <MSL/stddef.h>
```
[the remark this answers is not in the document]
ribbanya: So adding the path in `Makefile` should be enough?

src/melee/lb/lbcommand.c
r-burns: If you want to specify f32, than the header that does `typedef float f32` can also include <float.h> and `#define F32_MAX FLT_MAX`.

ribbanya: But wouldn't that make `F32_MAX` dependent on the platform? I'm not sure I'm understanding.

r-burns: Yes, and platform-dependent typedefs are in Runtime/platform.h. That's where you've typedef'd f32 as float. So Runtime/platform.h "knows" that a float is 32 bits, and it can also include <float.h> and define `F32_MAX` to be equivalent to `FLT_MAX`. And then if you want to use F32_MAX that definition is available in Runtime/platform.h.

ribbanya: `typedef float f32` is technically an implementation detail on `MWERKS_GEKKO` but as far as I understand, `float.h` doesn't guarantee that `FLT_MAX` refers to 32-bit floats, correct?

This is my understanding:
- `f32`: Guaranteed 32-bits on any platform
- `F32_MAX`: Guaranteed to be `3.4028235E38` (bit for bit)
- `FLT_MAX`: Guaranteed to be the maximum single-precision floating-point value for the current platform

... sorry if I'm being dumb.

r-burns: Ok yeah, we're on the same page. I was thinking this header should reuse the knowledge that f32 == float for the F32/FLT define. TBH, it's not that important, we can leave it hardcoded here for now. I'm going to resolve this thread.

src/string.h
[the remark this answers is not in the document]
r-burns: Yeah

src/Runtime/platform.h
```diff
@@ -90,10 +99,10 @@ typedef jmp_t jtbl_t[];
...
 /// #BOOL true
-#define TRUE (1)
+#define TRUE 1
 
 /// #BOOL false
-#define FALSE (0)
+#define FALSE 0
 
 /// A null pointer
 #define NULL ((any_t) 0)
```
r-burns, line 108: don't need to do this here but NULL also belongs in stddef.h

r-burns (approved):

PR #670, opened 2023-01-12
Fix `generate_context.sh`

ribbanya:


src/common_structs.h
```diff
@@ -126,4 +128,40 @@ typedef enum GroundOrAir {
...
+    uint io_mode : 3;
+    uint buffer_mode : 2;
+    uint file_kind : 3;
+    uint file_orientation : 2;
+    uint binary_io : 1;
+} __file_modes;
+
+typedef struct FILE FILE;
+
+struct FILE {
```
r-burns, line 142: This should be in stdio.h

src/melee/ft/types.h
```diff
@@ -492,7 +492,8 @@ typedef enum ftCommonAction {
     ASID_CAPTUREDAMAGECRAZYHAND,
     ASID_CAPTUREWAITCRAZYHAND,
     ASID_THROWNCRAZYHAND,
-    ASID_BARREL
+    ASID_BARREL,
+    ASID_MAX
```
r-burns, line 496: Maybe ASID_COUNT or similar is more accurate here? To me ASID_MAX implies the max valid ftCommonAction value, or ASID_MAX == ASID_BARREL.

ribbanya: I was just following `FTKIND_MAX` but I had planned on changing them all to pascal case and `max` to `count` anyway. Just wanted the context to compile in this case.

r-burns: Ah didn't realize there was precedent here, sounds good.

ribbanya (commented):
This seems fine to merge for now, but I want to get rid of `generate_context.sh` soon in favor of a pure Python solution.

r-burns (approved):

PR #671, opened 2023-01-13
Fix `implicit-function-declaration` warnings, pass 11

ribbanya:


src/melee/it/item.h
```diff
@@ -20,664 +21,10 @@
 #include <sysdolphin/baselib/jobj.h>
 #include <sysdolphin/baselib/random.h>
 
-// Item State Change Flags
```
ribbanya, line 23: Moved to `it/types.h`.

src/melee/it/types.h
```diff
@@ -0,0 +1,854 @@
...
+    /// @brief When grabbing a fighter, run this function on self.
+    HSD_GObjEvent on_grab;
+
+    /** @at{D38} @sz{4}
+     * @brief When grabbing a fighter, run this function on them.
+     *
+     * @p gobj0 - The victim of the grab. \n
+     * @p gobj1 - This item's entity.
+     */
+    HSD_GObjInteraction on_grabbed_for_victim;
```
ribbanya, line 545: Started reworking callback names. `if_` means a predicate (returns a `BOOL`) while `on_` is an event (returns `void`).

r-burns (approved):

src/melee/it/types.h
r-burns: Don't like the `if_` predicate because we'll end up seeing `if (if_jumped_on()` which is weird. How about `is_jumped_on`, etc

ribbanya: You don't think that's too suggestive of a boolean field?

r-burns: Or maybe just skip the prefix so we have `if (jumped_on`, `if (collided`, etc. It'll be pretty obvious from context when something is returning a bool.

ribbanya: And drop the `on_` too?

ribbanya: > Don't like the `if_` predicate because we'll end up seeing `if (if_jumped_on()` which is weird. How about `is_jumped_on`, etc

Keep in mind that can't happen exactly because it would have to at least be
```c
if (ip->if_jumped_on(fp))
```

r-burns: I like the `on_` prefix. `on_grab`/`on_anim`/`on_phys` are very clear what they're doing, but without the prefix `grab`/`anim`/`phys` could easily be misinterpreted as a number of other imperative things such as `do_grab`, `update_anim`, `recompute_phys` etc

ribbanya: What do you propose for the collision predicate?

r-burns: How about just `collided`? Alternatively `did_collide` or `detected_collision`

ribbanya: Ok so no prefixes to indicate a callback. That works for me, it's how Microsoft does it (`collided` for after a collision happens, `colliding` for before a collision happens).

src/melee/it/types.h
```diff
@@ -0,0 +1,854 @@
...
+
+    /// @at{2C} @sz{4}
+    f32 facing_dir;
+
+    f32 init_facing_dir;
+    f32 x34_spin_speed;
+    f32 x38_scale;
+    f32 x3C;
+    Vec3 x40_vel;
+    Vec3 pos;
```
r-burns, line 343: Oh yeah, and if we're dropping the hex offsets from the name, let's keep the offset in a comment until decomp is done.

src/melee/it/types.h
ribbanya: Before|After
-|-
`if_anim`|`animated`
`on_phys`|`physics_updated`
`if_coll`|`collided`
`on_accessory`|`on_accessory`
`if_touched`|`touched`
`on_enter_hitlag`|`entered_hitlag`
`on_exit_hitlag`|`exited_hitlag`
`if_jumped_on`|`jumped_on`
`on_grab`|`grab_dealt`
`on_grabbed_for_victim`|`grabbed_for_victim`

src/melee/it/types.h
ribbanya: Oh, missed that one.

src/melee/it/types.h
r-burns: I like it! Also will carry over to nice descriptive function names, like `void itCapsulePhysicsUpdated(void)`.

src/melee/it/types.h
r-burns: Also needed for `hold_kind`, `owner`, and `init_facing_dir`.

PR #673, opened 2023-01-13
Implement `stdbool.h`

ribbanya:


ribbanya (commented):
78f8d841cd69a0a2cbb80612e17a66dcc2c39d16 is just
```sh
replace_symbol BOOL bool
replace_symbol TRUE true
replace_symbol FALSE false
```

so please don't click through it.

src/MSL/stdbool.h
```diff
@@ -0,0 +1,17 @@
+/// @file
+/// @todo Assumes @c GEKKO.
+#ifndef STDBOOL_H
+#define STDBOOL_H
+
+/// A signed integer used to contain boolean values
+/// @todo Rename to @c bool.
```
r-burns, line 7:
```suggestion
```

src/MSL/stdbool.h
```diff
@@ -0,0 +1,17 @@
+/// @file
+/// @todo Assumes @c GEKKO.
```
r-burns, line 2: wdym assumes gekko?

ribbanya: The C standard is an unsigned integer.

r-burns (approved):

PR #674, opened 2023-01-13
Fix `item.h`, pass 1

ribbanya:


src/melee/ft/fighter.h
```diff
@@ -45,77 +46,16 @@ extern void* lbl_804D6548;
 extern void* lbl_804D654C;
 extern void* lbl_804D6550;
 
-// Ternary macro for fcmpo-based facing direction check
```
ribbanya, line 48: Moved to `ft/inlines.h` to simplify includes.

r-burns (approved):

PR #676, opened 2023-01-14
Fix `item.h`, pass 2

ribbanya:
There are some weird Doxygen comments in this one because I still haven't fully figured out how certain things work. It seems to have some pretty annoying tendencies, like you can't mark a function `@private` from the header, it must be marked at the definition.

So please just ignore Doxygen syntax weirdness for now.


src/melee/it/inlines.h
```diff
@@ -10,9 +14,33 @@ static inline Item* GetItemData(HSD_GObj* item_gobj)
...
+    case 6:
+    case 7: {
+        int temp_r3 = func_8017E068();
+        if (temp_r3 >= 0)
+            it->xC3C = lbl_804D6D28->x80_float[temp_r3];
+    }
+    }
+}
+
+static inline void HSD_JObjSetScaleItem(Item* it, HSD_JObj* jobj, Vec3* scl)
```
r-burns, line 39: these can all stay in item.c

ribbanya: I was hoping you wouldn't notice so I could move them later when I'm not being lazy but ok

r-burns (approved):

ribbanya:
Was above to get rid of `it/inlines.h`.

src/melee/it/item.c
```diff
@@ -132,6 +132,17 @@ void HSD_JObjSetScaleItem(Item* it, HSD_JObj* jobj, Vec3* scl)
...
+#ifdef MUST_MATCH
+inline
+#endif
+    void
+    HSD_JObjSetFacingDirItem(HSD_JObj* jobj, Item* it)
+{
+    if (it->xDC8_word.flags.x19 == true)
+        HSD_JObjSetRotationY(jobj, M_PI / 2 * it->facing_dir);
+}
+
```
ribbanya, line 145: I think this might be in `lb` or something because `Fighter` and `Item` both have their `facing_dir` at 0x2C.

PR #677, opened 2023-01-14
Fix `item.h`, pass 3

ribbanya:


src/melee/it/types.h
```diff
@@ -44,40 +58,75 @@
...
+    struct ItemStateContainer* states;
+
+    /// @at{4} @sz{4}
+    void (*spawned)(HSD_GObj* item);
+
+    /// @at{8} @sz{4}
+    void (*destroyed)(HSD_GObj* item);
+
+    /// @at{C} @sz{4}
+    void (*picked_up)(HSD_GObj* item);
```
r-burns, line 93:
```suggestion
    void (*destroyed)(HSD_GObj* item); /// @at{8} @sz{4}
    void (*picked_up)(HSD_GObj* item); /// @at{C} @sz{4}
```

Can we do something like this for the structs? They're taking up so much vertical space, makes it way harder to get a gist of them

ribbanya: using a different struct as an example because it's small. Options are...

1. The current way
```c
/// @sz{1C}
struct Item_DynamicBones {
    /// @at{0} @sz{4}
    /// @todo If this is 256, dynamics are not processed
    int flags;

    /// @at{4} @sz{4}
    HSD_JObj* skeleton;

    /// @at{8} @sz{4}
    /// @todo Stored at @c 8000FDD4, comes from a nonstandard heap at
    ///       @c -0x52fc(r13)
    unk_t unk_ptr;

    /// @at{C} @sz{4}
    /// @brief The number of bones in this bone set.
    int count;

    /// @at{10} @sz{C}
    Vec3 unk_vec;
};
```

2. What you suggested, annotations to the right. I personally hate it because it requires you to parse variable length lines of text just to see the offsets (when looking at the source), also looks pretty claustrophobic. Note that the `<` is required by Doxygen.
```c
/// @sz{1C}
struct Item_DynamicBones {

    /// @todo If this is 256, dynamics are not processed
    int flags; ///< @at{0} @sz{4}

    HSD_JObj* skeleton; ///< @at{4} @sz{4}

    /// @todo Stored at @c 8000FDD4, comes from a nonstandard heap at
    ///       @c -0x52fc(r13)
    unk_t unk_ptr; ///< @at{8} @sz{4}
    
    /// @brief The number of bones in this bone set.
    int count; ///< @at{C} @sz{4}

    /// @at{10} @sz{C}
    Vec3 unk_vec;
};
```

3. Annotations on the left. Looks pretty awful imo, and if the struct goes into multi-digit addresses then it's not clear where the space should go.
```c
/// @sz{1C}
struct Item_DynamicBones {
    /// @todo If this is 256, dynamics are not processed
    /** @at{8} @sz{4} */ int flags;

    /** @at{8} @sz{4} */ HSD_JObj* skeleton;

    /// @todo Stored at @c 8000FDD4, comes from a nonstandard heap at
    ///       @c -0x52fc(r13)
    /** @at{8} @sz{4} */ unk_t unk_ptr;

    /// @brief The number of bones in this bone set.
    /** @at{C} @sz{4} */ int count;

    /** @at{10} @sz{C} */ Vec3 unk_vec;
};
```

Those are the most vertically compact options, I think. What's the use case when you say "harder to get a gist of them"? The docs summary looks like this
image

ribbanya: Keep in mind that the goal is to eventually have at least one sentence per member, often full paragraphs, at which point scanning the structs will be very unpleasant no matter how it's formatted.

ribbanya: Option 4: annotate structs in separate markdown files.

ribbanya: Which would look something like this, copypasted from stackoverflow
```c
/** @struct foreignstruct
 *  @brief This structure blah blah blah...
 *  @var foreignstruct::a 
 *  Member 'a' contains...
 *  @var foreignstruct::b 
 *  Member 'b' contains...
 */
```

r-burns: That's fair, once these are fully populated with info it'll make more sense

r-burns (approved):

PR #678, opened 2023-01-14
Replace GEKKO define with MWCC __PPCGEKKO__

r-burns:


src/Runtime/platform.h
```diff
@@ -89,6 +89,10 @@ typedef jmp_t jtbl_t[];
 /// A @c void callback with no arguments.
 typedef void (*Event)(void);
 
+#if defined(__PPCGEKKO__)
+#define GEKKO
+#endif
+
```
ribbanya, line 95: Shouldn't we just find/replace instead?

r-burns: Well it'll need to be defined manually by gcc et al so I think it makes sense to abstract it behind GEKKO.

ribbanya: Yeah but then they'd need to manually define `GEKKO`, right? So they can just manually define `__PCCGEKKO__` instead.

r-burns: Ok fair enough, I'll just replace them all.

ribbanya (approved):

PR #679, opened 2023-01-14
Fix `item.h`, pass 4

ribbanya:


src/melee/it/types.h
```diff
@@ -262,18 +215,25 @@ struct ItemAttr {
...
 
 struct ItemDynamicsDesc {
-    s32 x0_boneID;   // bone index;
-    void* x4_params; // dynamics params;
-    s32 x8_num;      // number of children bones to make dynamic
+    int x0_boneID;   // bone index;
+    unk_t x4_params; // dynamics params;
+
+    /// @at{8} @sz{4}
+    /// @brief Number of children bones to make dynamic.
```
ribbanya, line 222: We decided on out-of-line docs midway through the creation of this PR so I'll move these over later.

tools/functions.sh
```diff
@@ -5,7 +5,8 @@ function replace_symbol {
     local find=${find//[.]/\\.}
     local replace=${2//\//\\\/}
     echo "$find -> $replace"
-    find asm src include -type f -regex 'Makefile\|.*\.\(c\|h\|s\|mk\)$' -exec sed -i "s/\b$find\b/$replace/g" {} +
+    find asm src -type f -regex 'Makefile\|.*\.\(c\|h\|s\|mk|dox|md\)$' \
```
ribbanya: Removed `include` dir , added `dox` and `md`.

r-burns (approved):

PR #681, opened 2023-01-15
Fix `implicit-function-declaration` warnings, pass 12

ribbanya:
`gr` and `it` done.


src/melee/lb/lbfile.c
```diff
@@ -48,12 +52,15 @@ bool lbFile_800161A0()
...
 {
     cancel = false;
-    HSD_DevComRequest(r3, r4, r5, r6, r7, r8, lbFile_8001615C, 0);
-    do {
-    } while (!lbFile_800161A0());
+    HSD_DevComRequest(arg0, arg1, arg2, arg3, arg4, arg5, lbFile_8001615C, 0);
+
+    do
+        continue;
+    while (!lbFile_800161A0());
```
r-burns, line 63: This looks super weird, please keep the brackets on this

ribbanya: I'll put both then. Anyway I didn't mean to commit this lol.

src/Runtime/platform.h
```diff
@@ -151,4 +151,8 @@ typedef void (*Event)(void);
 #define U32_MAX 0xFFFFFFFF
 #define F32_MAX 3.4028235E38
 
+#define FLAGS_NONE 0
+#define HAS_FLAG(value, flag) (((value) & (flag)) != FLAGS_NONE)
+#define NOT_FLAG(value, flag) (((value) & (flag)) == FLAGS_NONE)
```
r-burns, line 156:
```suggestion
#define HAS_FLAG(value, flag) ((value) & (flag))
#define NOT_FLAG(value, flag) (!((value) & (flag)))
```

ribbanya: Kind of the same thing as `!gobj` vs `!= NULL` and wouldn't work in a language like C# but if you think it's better then ok.

src/melee/gr/grbattle.c
```diff
@@ -315,9 +342,13 @@ static void func_8021A3BC(HSD_GObj* gobj)
     HSD_JObj* jobj = gobj->hsd_obj;
     HSD_GObj* bg_gobj;
     s32 temp_r0_2;
-    s16 sp28[BATTLE_BG_MAX] = { 1, 2, 4 };
+    s16 sp28[] = { 1, 2, 4 };
```
r-burns, line 345:
```suggestion
    s16 sp28[BATTLE_BG_MAX] = { 1, 2, 4 };
```

r-burns (approved):

src/melee/gr/grbattle.c
r-burns: Just thought we should keep this one, since it makes the iteration a little more clear.

ribbanya: `clang` complains and I don't think it's worth turning off the relevant warning (`-Wgnu-folding-constant`). Could branch the preprocessor but... not for this, right?

r-burns: Not familiar with that warning, reading the docs and not seeing why it would triggered here. BATTLE_BG_MAX is a constant so why is it complaining?

ribbanya: Because it's `const` and not a define. I'll change that and it'll work on both compilers.

PR #682, opened 2023-01-15
Fix `implicit-function-declaration` warnings, pass 13

ribbanya:
Just `sysdolphin` remaining.


src/melee/mp/mpcoll.c
```diff
@@ -1135,7 +1145,7 @@ asm void func_80043C6C(CollData* arg0, s32 arg1, s32 arg2)
 } // clang-format on
 #pragma pop
 
-#else
+#elif false
```
r-burns, line 1148: huh

ribbanya: Bad signature in the function. Feel free to match it instead.

r-burns: Ah okay we can leave it for now

r-burns (approved):

PR #690, opened 2023-01-16
Fix `clang` errors

ribbanya:
Fixes all errors (not warnings) generated by `clang`.


src/Runtime/__mem.c
```diff
@@ -7,12 +7,24 @@ void* memset(void* dst, int val, unsigned long /*size_t*/ n)
...
+#define INCREMENT_ASSIGN(ptr, type, value) (*++((type*) (ptr)) = (value))
+#else
+#define INCREMENT_ASSIGN(ptr, type, value)                                     \
+    do {                                                                       \
+        type* __INCREMENT_ASSIGN_tmp;                                          \
+        (ptr) = ((type*) (ptr)) + 1;                                           \
+        __INCREMENT_ASSIGN_tmp = ((type*) (ptr));                              \
+        *__INCREMENT_ASSIGN_tmp = (value);                                     \
+    } while (false);
+#endif
```
ribbanya, line 20: This implementation was tested by substituting it on line 36, and the dol matched. Using the same macro does not work in the loop below (line 51), so keeping the original code was necessary.

If you have a more elegant solution, I'd be happy to hear it.

src/melee/ft/ftparts.c
```diff
@@ -1464,9 +1611,20 @@ asm void func_80074A4C(struct _HSD_GObj*, u32, s32)
...
+
+#endif
+
+#ifdef MWERKS_GEKKO
+
 #pragma push
-asm s32 func_80074A74(HSD_GObj*, s32)
-{ // clang-format off
+asm s32 func_80074A74(HSD_GObj*,
+                      s32){ // clang-format off
```
ribbanya, line 1627: I really don't know, I think it's trying to express itself.

src/sysdolphin/baselib/forward.h
```diff
@@ -0,0 +1,6 @@
+#ifndef SYSDOLPHIN_BASELIB_FORWARD_H
+#define SYSDOLPHIN_BASELIB_FORWARD_H
+
+typedef struct _HSD_Obj HSD_Obj;
+
+#endif
```
ribbanya, line 6: Just a quick fix to eliminate `object.h`⮀`class.h` circular dependency. Will do a full forward declaration sweep later.

r-burns (approved):

PR #691, opened 2023-01-16
Match OSContext, OSInterrupt

r-burns:


src/dolphin/os/OSInterrupt.h
```diff
@@ -44,6 +44,17 @@ typedef enum {
...
+
+#define OS_INTRMASK_MEM_0 (0x80000000U >> OS_INTR_MEM_0)
+#define OS_INTRMASK_MEM_1 (0x80000000U >> OS_INTR_MEM_1)
+#define OS_INTRMASK_MEM_2 (0x80000000U >> OS_INTR_MEM_2)
+#define OS_INTRMASK_MEM_3 (0x80000000U >> OS_INTR_MEM_3)
+#define OS_INTRMASK_MEM_ADDRESS (0x80000000U >> OS_INTR_MEM_ADDRESS)
+#define OS_INTRMASK_DSP_AI (0x80000000U >> OS_INTR_DSP_AI)
+#define OS_INTRMASK_DSP_ARAM (0x80000000U >> OS_INTR_DSP_ARAM)
+#define OS_INTRMASK_DSP_DSP (0x80000000U >> OS_INTR_DSP_DSP)
+#define OS_INTRMASK_AI_AI (0x80000000U >> OS_INTR_AI_AI)
```
ribbanya, line 57: It's whatever for this PR but wouldn't it make sense to `#define BASE_VIRT_ADDR 0x80000000U` or something?

src/dolphin/os/OSInterrupt.h
```diff
@@ -52,17 +63,56 @@ typedef enum {
...
+#define OS_INTRMASK_PI_PE (OS_INTRMASK_PI_PE_TOKEN | OS_INTRMASK_PI_PE_FINISH)
 
 typedef void (*OSInterruptHandler)(__OSInterrupt, OSContext*);
 
-extern u32 __OSLastInterruptSrr0;
-extern s16 __OSLastInterrupt;
-extern s64 __OSLastInterruptTime;
+extern volatile u32 __OSLastInterruptSrr0;
+extern volatile s16 __OSLastInterrupt;
+extern volatile s64 __OSLastInterruptTime;
```
ribbanya, line 108: Also not immediately important but how do you want to handle `volatile` across different platforms?

ribbanya (approved):
Some things are a bit messy (missing `(void)`, enum declared in C when it's probably used elsewhere) but I'm not going to nitpick on this PR. Can all be cleaned up later.

src/dolphin/os/OSInterrupt.h
r-burns: This isn't an address, it's a mask value.

src/dolphin/os/OSInterrupt.h
r-burns: I don't think we'll need to, the semantics of volatile are pretty well defined.

PR #696, opened 2023-01-17
Match OSInit

r-burns:
Based on prime decomp. Still needs lots of cleanup.


src/dolphin/db/db.c
```diff
@@ -64,9 +64,9 @@ static void __DBExceptionDestination(void)
 
 #endif
 
-int __DBIsExceptionMarked(u8 a)
+bool __DBIsExceptionMarked(u8 exception)
```
ribbanya, line 67: Does `OSException` not match?

src/dolphin/os/OSInit.c
```diff
@@ -70,603 +105,312 @@ void ClearArena(void)
...
+        // time to grab a bunch of debug info from the DVD
+        // the address for where the BI2 debug info is, is stored at
+        // OS_BI2_DEBUG_ADDRESS
+        DebugInfo = (BI2Debug*) *((u32*) OS_BI2_DEBUG_ADDRESS);
+
+        // if the debug info address exists, grab some debug info
+        if (DebugInfo != NULL) {
+            BI2DebugFlag = &DebugInfo->debugFlag; // debug flag from DVD BI2
+            __PADSpec =
+                (u32) DebugInfo->padSpec; // some other info from DVD BI2
```
ribbanya, line 177: I won't ask you to change it right now but in the future please place comments above the line they describe, nor beside.

src/dolphin/db/db.c
r-burns: Nope, these really do need to be u8 :/

PR #699, opened 2023-01-17
Enable `-requireprotos` and `-warn iserror`

ribbanya:
Depends on #698. Prototypes are now required, and warnings are considered errors, by default. Either option can be disabled using `make` flags.


src/MSL/string.c
```diff
@@ -1,5 +1,10 @@
+#include <MSL/string.h>
 #define EOF -1
+
+#ifdef MWERKS_GEKKO
```
r-burns, line 4:
```suggestion
#ifdef __MWERKS__
```
Not gekko-specific?

src/MSL/string.h
```diff
@@ -6,9 +6,10 @@
 char* strncpy(char* __restrict dst, const char* __restrict src,
               unsigned long n);
 int __StringRead(void* isc, int ch, int Action);
-unsigned long(strlen);
 char* strchr(const char* str, int chr);
 int strncmp(const char* str1, const char* str2, unsigned long n);
 int strcmp(const char* str1, const char* str2);
+char*(strcpy) (char* __restrict dst, const char* __restrict src);
+unsigned long(strlen)(const char* s);
```
r-burns, line 13:
```suggestion
char* strcpy(char* __restrict dst, const char* __restrict src);
size_t strlen(const char* s);
```

src/MSL/string.c
ribbanya: [p. 267](https://www.nxp.com/docs/en/reference-manual/CWMCUCFCMPREF.pdf)

src/MSL/string.h
r-burns: Looks like got reverted somehow?

r-burns (approved):

Issue #595, opened 2022-11-27
Match stuff


ribbanya:
Author: @VetriTheRetri 

* `ftcoll`
  * #708
    - [x] [`func_80078A2C`](https://decomp.me/scratch/KQhCQ)
  * #917
    - [x] [`func_80078C70`](https://decomp.me/scratch/sN0Ib)
  * #916
    - [x] [`func_8007B1B8`](https://decomp.me/scratch/vdfkO)
    - [x] [`ftColl_CreateReflectHit`](https://decomp.me/scratch/svoDD)
    - [x] [`ftColl_CreateAbsorbHit`](https://decomp.me/scratch/0ndz9)
* `lbcollision`
  * #724
    - [x] [`func_8000ACFC`](https://decomp.me/scratch/di7dZ)
* `ftaction`
  * #868
    - [x] [`func_80073240`](https://decomp.me/scratch/5kHNA)

ribbanya:
Not everything matches how it does in the scratches, but everything is in the repo now.

PR #708, opened 2023-01-18
Match `func_80078A2C`

ribbanya:
Depends on #706. Resolves part of #595.

Cannot be fully matched due to float literals.

```yml
Co-authored-by: VetriTheRetri <74500716+VetriTheRetri@users.noreply.github.com>
```


src/melee/ft/ftcoll.c
```diff
@@ -179,3 +192,3028 @@ f32 func_800765F0(Fighter* fp, HSD_GObj* victim,
...
+                    }
+                }
+            }
+        }
+    next_gobj:
+        victim_gobj = victim_gobj->next;
+    }
+}
+
+#endif
```
ribbanya, line 3219: Function is here. Needs a lot of cleanup but so does a lot of code in the codebase. Matches except that it adds two entries to `sdata2`.

src/melee/lb/lbcollision.h
```diff
@@ -1,24 +1,39 @@
...
 
+#include <melee/ft/forward.h>
+#include <melee/lb/forward.h>
 #include <sysdolphin/baselib/jobj.h>
 
-/// @todo Does not belong here.
-typedef struct _Hitbox {
-    s32 x0;
+struct Hitbox {
+    Tangibility tangiblity;
```
ribbanya, line 9: This is probably wrong, or at least the elements of `Tangibility` are named wrong, but having it be any enum is better than the alternative. It may or may not share an enum with `Hurtbox::tangibility`.

r-burns (approved):

PR #712, opened 2023-01-20
Use `GET_FIGHTER` macro, pass 2

ribbanya:
Depends on #711 ([diff](https://github.com/ribbanya/melee/compare/pr-getfighter-1...pr-getfighter-2)).

Replaces all easily matching instances of `gobj->user_data` with `GET_FIGHTER(gobj)`. Not at parity with #614 yet.

### DIFF
--- a/src/melee/ft/chara/ftDonkey/ftdonkey_SpecialN.c
+++ b/src/melee/ft/chara/ftDonkey/ftdonkey_SpecialN.c
@@ -10,7 +10,7 @@
 
 void ftDonkey_SetCallbacks_SpecialN(HSD_GObj* fighter_gobj)
 {
-    Fighter* fp = fighter_gobj->user_data;
+    Fighter* fp = GET_FIGHTER(fighter_gobj);
     fp->cb.x21DC_callback_OnTakeDamage = &ftDonkey_8010D774;
     fp->cb.x21E4_callback_OnDeath2 = &ftDonkey_8010D774;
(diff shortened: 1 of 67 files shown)


src/melee/ft/chara/ftDonkey/ftdonkey_SpecialN.c
```diff
@@ -30,7 +30,13 @@ void ftDonkey_UpdateDKVelocityAfterPunch(HSD_GObj* fighter_gobj)
 void ftDonkey_SpecialN_StartAction(HSD_GObj* fighter_gobj)
 {
     Fighter* fp = GET_FIGHTER(fighter_gobj);
-    ftDonkeyAttributes* donkey_attr = getFtSpecialAttrs(fp);
+    ftDonkeyAttributes* donkey_attr = fp->x2D4_specialAttributes;
+
+    /// @todo Unused stack.
+#ifdef MUST_MATCH
+    u8 unused[8];
+#endif
```
r-burns, line 38: Why'd these get changed

ribbanya: There's something wrong with that inline that causes `GET_FIGHTER` to not match in cases where it should. Given that there are so few exceptions to its usage in `fighter.c`, which does not access special attributes, I think it should be deprecated until we figure out attributes better (including state variable union hell).

r-burns (approved):

PR #729, opened 2023-01-28
Port `build.sh` to Python

ribbanya:
This PR creates a new CLI app. To install it, run `pip install -r requirements.txt`. This will add the `melee` command to your (venv's) path.

From there, `melee -h` produces:
```
usage: melee [-h] {build,b} ...

positional arguments:
  {build,b}
    build (b)
              Build the project

options:
  -h, --help  show this help message and exit
```

`melee b -h` produces:
```
usage: melee build [-h] [-c] [-r REBUILD] [-e] [-f] [-m] [-n] [-s] [-x] [-l] [-d] [-j JOBS] [-p]

options:
  -h, --help            show this help message and exit
  -c, --clean           "make clean" before running make
  -r REBUILD, --rebuild REBUILD
                        delete *.o files matching the given pattern; -c takes priority
  -e, --expected        after a successful make run, sync build to expected/build
  -f, --frank           pass EPILOGUE_PROCESS=1 to make
  -m, --map             pass GENERATE_MAP=1 to make
  -n, --non-matching    pass NON_MATCHING=1 to make
  -s, --skip-check      pass SKIP_CHECK=1 to make
  -x, --clear           clear the console before executing this script
  -l, --log             tee output to build.log
  -d, --dump            dump the built dol to asm (requires dadosod)
  -j JOBS, --jobs JOBS  pass a number of threads to make
  -p, --pretty          run "make format" before running make
```

The plan is to move all loose scripts to `melee-cli` and use it as an entry point to anything you would reasonably want to do on the command line when working with melee. For example, `melee diff` would be an entry point to `tools/asm-differ/diff.py`, `melee dump` would be an entry point to `dadosod`, etc.


tools/parse_map.py
```diff
@@ -1,10 +1,9 @@
 import csv
```
ribbanya, line 1: No changes to this file, just ran `black`.

ribbanya:
Will revisit this after moving to `ninja`.

PR #735, opened 2023-01-30
Port `replace_symbol` to Python

ribbanya:
Depends on #729 ([diff](https://github.com/ribbanya/melee/compare/wip-cli..pr-replace)).

```
$ melee r -h                         
usage: melee replace [-h] [-b] [-i] [-r] PATTERN=REPLACEMENT [PATTERN=REPLACEMENT ...]

positional arguments:
  PATTERN=REPLACEMENT  replace PATTERN with REPLACEMENT in all files

options:
  -h, --help           show this help message and exit
  -b, --no-boundaries  do not respect word boundaries
  -i, --ignore-case    perform case-insensitive replacements
  -r, --reverse        perform the replacements in reverse
```

Can take multiple patterns, for example
```
melee r func_80005BB0=lbColl_80005BB0 func_80005C44=lbColl_80005C44 \
    func_80005EBC=lbColl_80005EBC func_80005FC0=lbColl_80005FC0
```

It compiles all of the patterns into one regex, so it is quite fast.


ribbanya:
Will implement this in a different way.

PR #736, opened 2023-02-09
Label more HSD functions/data

r-burns:
Correct label names obtained from Zoids VS, Pokemon XD, and Killer7.


ribbanya:
Did you keep a list of the renames? Would save time reviewing but if not it's okay.

r-burns:
Ah no, I didn't.

asm/sysdolphin/baselib/texp.s
```diff
@@ -3347,7 +3308,7 @@ lbl_804D5FE4:
 .global lbl_804DE7D8
 lbl_804DE7D8:
     .4byte 0x437F0000
-    .4byte 0x00000000
+    .balign 8
```
ribbanya, line 3311: `lbl_804DE7D8` is `255.0F` and `lbl_804DE7E0` is `255.0`. Should not need a manual `balign`. If you want to use `.balign 8` instead of `.4byte NULL` while it's GAS that's fine, but might as well change them to `.float` and `.double`.

ribbanya (changes requested):

ribbanya (approved):

PR #740, opened 2023-02-19
Use custom container for GitHub Actions

ribbanya:
* Upgrade devkitpro's old Debian release (`buster`) to Testing (`bookworm`)
* The main reason for the above is to support Python 3.11
* Compile wibo and put it in `PATH` on the container
* Download and patch `mwcc` on the container, no need to download it on build
* Set up Python 3.11 `venv` with our packages
* Greatly simplify our workflows

The container lives at [ribbanya/melee](https://hub.docker.com/repository/docker/ribbanya/melee/general) but if doldecomp owners want to make their own that's cool with me.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         makeflags: ["GENERATE_MAP=1", "NON_MATCHING=1"]
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


ribbanya (commented):
Drafting this for now. There are a lot of improvements to be made.

ribbanya:
Superseded by #751.

PR #745, opened 2023-02-26
Add Windows build job and improve speed of Ubuntu build

ribbanya:
1. Add Windows build
   * This build relies only on PowerShell, MinGW, Git for Windows, and Python 3; no more MSYS2 or devkitPro.
   * It mirrors the actions of the Ubuntu build except for uploading the map artifact.
2. Use a new compilers zip
   * Only contains binaries we actually use, including a pre-patched 1.1 linker, as well as `powerpc-eabi-as.exe` which is needed for the Windows build.
3. Improve speed of Ubuntu build
   * Download the `wibo` binary from their CI release instead of building it ourselves.
   * The sha1 of the 0.4.1 binary is hard-coded and was double-checked locally by me.
   * This lets us cut down on the packages installed.
4. Clean up the build script
   * Use version numbers everywhere instead of `latest` to avoid future breakages.
   * Prefer breaking up long lines for readability and git-friendliness.
5. Tweak Makefile to be able to run under Windows shells
   * Just need to use `make`'s `file` function.

---

Going to need to change the repo settings to require `build-ubuntu` and `build-windows` instead of `build`, under Branches > Branch protection rules > `master`.

---

Also need to allow `cache-apt-pkgs-action`. Settings > Actions > Allow specified actions and reusable workflows:
```
awalsh128/cache-apt-pkgs-action@*,
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,7 +79,6 @@ jobs:
       run:
         make WINDOWS=1 `
           AS=".\tools\mwcc_compiler\powerpc-eabi-as.exe" `
-          HOSTCC=gcc `
           -j "$env:NUMBER_OF_PROCESSORS" `
           ${{ matrix.makeflags }}
     - name: Calc progress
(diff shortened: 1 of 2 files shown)


r-burns (approved):
Looks great, thanks for setting this up. Will be really useful for me since I don't have a windows machine handy to test on.

PR #746, opened 2023-03-01
Fix editorconfig workflow

ribbanya:
Applies https://github.com/NixOS/nixpkgs/commit/31ce09b37790915618e839ed6a728f7b40157fdc.

### DIFF
--- a/.github/workflows/editorconfig.yml
+++ b/.github/workflows/editorconfig.yml
@@ -32,11 +32,9 @@ jobs:
         # nixpkgs commit is pinned so that it doesn't break
         # editorconfig-checker 2.4.0
         nix_path: nixpkgs=https://github.com/NixOS/nixpkgs/archive/c473cc8714710179df205b153f4e9fa007107ff9.tar.gz
-    - name: install editorconfig-checker
-      run: nix-env -iA editorconfig-checker -f '<nixpkgs>'
     - name: Checking EditorConfig
       run: |
-        cat "$HOME/changed_files" | xargs -r editorconfig-checker -disable-indent-size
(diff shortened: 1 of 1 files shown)


ribbanya:
Okay well that didn't work. I would really prefer to have a container so this doesn't happen again.

ribbanya:
This PR is running master's workflow so I don't actually know how to test this.

r-burns (approved):

ribbanya:
Going with #747.

r-burns:
Why? This change is tiny and will fix it.

ribbanya:
I don't want to rely on a package manager that randomly breaks stuff. The container image is the same hash every time.

r-burns:
If you'd rather pin to a specific version you can do it like this https://github.com/nix-community/home-manager/commit/9438e56a7b7df5e02091027dccdd8d86768f1a2d

ribbanya:
I'm just not going to use Nix for CI.

r-burns:
I think you are overreacting to a very small breakage.

ribbanya:
Possibly. But in this instance you had to use your preexisting Nix knowledge to find the patch whereas "download exe with hash and run it" is something everyone understands and it will work until GitHub shuts down.

r-burns:
I didn't use any nix knowledge, I just looked at upstream nixpkgs for the fix.

ribbanya:
Okay well I don't think this is going anywhere productive. Please choose one of the PRs and merge it.

ribbanya (commented):
Can we add the `GITHUB_STEP_SUMMARY` stuff from #747?

PR #747, opened 2023-03-01
Use container image for `check-editorconfig` [skip editorconfig]

ribbanya:
Closes #746.

Uses an official container image with a fixed tag for the `editorconfig-checker` binary, so no more future surprises.

Also cleans up the output and reduces the number of steps.

Tested in ribbanya#3 ([run](https://github.com/ribbanya/melee/actions/runs/4299328731)).


ribbanya:
Repurposed as a QoL fix ([run](https://github.com/ribbanya/melee/actions/runs/4299904658)).

* Use fixed Ubuntu version
* Remove hardcoded references to `doldecomp`
  * Now works on PRs to any repo
* Update `grep` exclude filter
* Run `editorconfig-checker` in verbose mode
* Add markdown step summary

PR #751, opened 2023-03-05
Publish packages for building Melee and generating documentation

ribbanya:
This PR restructures our workflows to use custom images, which are built automatically and uploaded to [ghcr.io](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry). The image (called "packages" by GitHub) are built every week and deployed to the host repo (meaning doldecomp when this PR is merged, but they will also deploy to forks, like [mine](https://github.com/ribbanya?tab=packages&repo_name=melee)). The packages are only deployed to the branches `master` or `main`, but they will be built if any of the files that generate them have changed.

Each package, when built, does a dry run of its entrypoint to check for success, meaning for example `build-linux` will try to build Melee. If the package fails its dry run, it will not be published to GHCR. Because of this automation, we can now use `latest` of everything upstream (devkitPro, WiBo, Ubuntu, Python, etc.) without worrying about whether it breaks. In the event that packages are failing to build, `build-melee` and `publish-pages` will simply pull from the latest working image.

The images used as base layers for each of the Dockerfiles are all cached by GitHub's runners on both [Ubuntu](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#cached-docker-images) and [Windows](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#cached-docker-images), meaning that despite being a few gigabytes each in size, the base layer takes roughly 0 seconds to pull.

Contributors who do not want to set up a working environment locally can now also use `docker` or `podman` to pull our GHCR images and build the project, whether they're working on Linux, Windows, or macOS.

# Packages

## `build-linux`

This package contains everything necessary to build Melee on Linux, and nothing else. It pulls the latest binaries from wibo, devkitPro, and our compilers from Discord.

## `gen-pages`

This package contains everything necessary to generate our docs, and it also sets up a Python venv and installs the packages in `requirements.txt`. This is necessary to run `m2ctx.py` but it also serves as a smoke test on our Python tooling.

## `build-windows`

This thing is incredibly slow, taking upwards of 12 minutes to finish, but it serves a very important purpose. It takes a fresh Windows Server 2022 image and, using PowerShell with [scoop](https://scoop.sh/), installs everything needed to build Melee, as well as our Python tooling. It uses PowerShell 5.1, which is installed by default on both Windows Server and Windows 10/11, so if this package builds successfully then that almost certainly means a Windows user can perform the same steps to get set up locally. It does not install devkitPro, and doesn't need to.

# Workflows

## `build-melee`

This workflow has been overhauled to use the pre-built images. It does not hardcode `doldecomp` anywhere, so it will work in a fully self-contained manner on any fork. The other notable change is that if `GENERATE_MAP=1` or `NON_MATCHING=1` fails, the other will continue anyway. However, if this workflow is running and a new push occurs, the Windows build (but not the Linux build) will be cancelled by the new pending workflow run.

The Linux build typically takes less than a minute now, so it should be pretty useful as a way to compile even even for someone who doesn't have a local setup.

## `publish-pages`

Now uses the pre-built page generator image to upload our documentation and `ctx.html` to GitHub Pages of the host repository. It will generate the pages no matter what, but will only upload to `master` or `main`. It's now included in pull request events to check whether the generation succeeds.

It also uploads a zipped version of the pages as an artifact to the run. Not to be confused with the GitHub Pages upload, which only occurs on the main branch, this artifact is the same as how `GENERATE_MAP=1` leaves `GALE01.map` as an artifact associated with every run.

## `publish-packages`

This builds the GHCR images. As stated above, this should work on any fork. It has a few rules that govern its operation:

1. The workflow only triggers if the files that generate the packages are changed, so `.github/workflows/publish-packages.yml` or anything inside `.github/packages`.
1. If this workflow is running, and a new push that triggers it occurs, it will stop and let the new run take over.
1. It will run on any branch that changes its files, but it will not push the packages to GHCR unless it's on the main branch.
1. On the main branch, it will run once a week on Monday at 5am EST.
1. Each package is tagged multiple times based on different criteria:
    * `latest` is the latest successful build.
    * `edge` is basically the same as `latest` in our case, but this tag is used by Alpine and others to refer to the `HEAD` of the main branch. We may at some point change `latest` to mean something more stable.
    * Every image is tagged with the branch name of the branch that produced it.
    * Every image is tagged with `sha-abcdefa` according to the sha of the commit that produced it, so this acts as a "permalink" to that image even when it's no longer tagged as `latest`.
    * The weekly builds are tagged with the date in the same form as devkitPro (`YYYYMMDD`).


ribbanya:
Builds are failing because the images don't exist on doldecomp. I will force merge this when it's approved.

---

~~Nevermind, there's not really a point to filtering the branch.~~

---

Nevermind again, it's not allowed.

ribbanya:
Superseded by #752.

PR #753, opened 2023-03-06
Clean up fighter vars union

r-burns:


ribbanya (commented):
Hmm I don't mind merging this because it is better but I'd rather start looking at `HSD_JOBJ`-style casts instead of unions.

r-burns:
It's a fixed region of the Fighter struct similar to the special attrs at x222C, so I think a union would be marginally safer than casting. But yeah we can revisit later. I'm just cleaning up the weird zero-sized array hack it's currently using to put lots of data at the same address.

ribbanya:
Well okay, for state vars I remember experimenting with a union of arrays. So instead of per-fighter it's like
```c
/// @sz I don't remember
union StateVars {
    u8 u8[0x100];
    s32 s32[0x100 / 4];
    u32 u32[0x100 / 4];
    f32 f32[0x100 / 4];
}
```
Which to me feels more sane than trying to reinvent it for every fighter. The "SpecialAttrs" are definitely their own structs, though, since they often contain GObj references and things. But how do you feel about that direction for state vars?

ribbanya:
The problem to me is that as it's designed right now, `ft/types.h` needs to include not only a vars union for every fighter, but within that a union for several states for each fighter.

If we could invert the direction of the headers I wouldn't mind the state vars being strongly typed.

ribbanya:
Like I would prefer to just cast because then `ftFox_SpecialSVars` could just be defined at the top of `ftFox_SpecialS.c` (since these structs are not reused anywhere else) and then cast that section of the fighter struct. A union seems like complete overkill because of how little code reuse it facilitates in this case.

ribbanya (approved):
I'm approving this for now but I'm going to submit a proof of concept PR for casting instead of unions.

ribbanya:
Turns out pointer casting doesn't work everywhere and I don't know if it can be made to. So we might be stuck with unions in some form forever.

r-burns:
Oh yeah, that's a good point. It also consumes stack which can mess up matches.

ribbanya:
Well I tried a lot of things on `ftNess_SpecialNHold_Anim`. The decrement statements just wouldn't match.

PR #762, opened 2023-03-17
Convert package names to lowercase

ribbanya:
I noticed @DRGN-DRC's [build was failing](https://github.com/DRGN-DRC/melee/actions/runs/4449315107/jobs/7813331863#step:5:1070) so this should fix it.


DRGN-DRC (approved):
I'm not super familiar with the workflow syntax, but I don't see any obvious mistakes. (^_^)b

PR #919, opened 2023-06-03
Pass `--passthru-magic-macros` to `pcpp`

ribbanya:


ribbanya:
Probably not the best solution, see #920.

PR #926, opened 2023-06-12
Hardcode image name

r-burns:
Otherwise when I'm pushing to my fork, the action tries to pull the image using my fork's owner/name which doesn't exist.

Drafting because if this is the correct solution, all the case/esac and tr stuff should be removed as well.

### DIFF
--- a/.github/workflows/build-melee.yml
+++ b/.github/workflows/build-melee.yml
@@ -6,6 +6,9 @@ on:
     paths: [ "**" ]
   pull_request:
 
+env:
+  IMAGE_REPO: doldecomp/melee
+
 jobs:
   build-linux:
     name: Linux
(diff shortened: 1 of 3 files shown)


ribbanya:
Hm, the same actions are working for [me](https://github.com/ribbanya/melee/actions/runs/5175808078/jobs/9323863610#step:9:101). For [Durgan](https://github.com/DRGN-DRC/melee/actions/runs/5110812902/jobs/9187119561#step:12:230), it failed two weeks ago but I'm not sure if that's consistent or just an outage. @DRGN-DRC could you re-run that job?

Anyway, the intended workflow is to push code to your own fork, which then pulls your own image to run actions, but I realize it's poorly documented on the melee side. I can't remember if there's something specific you need to enable.

If you want to do it this way and just not build packages on your fork, a better approach would be to set a [variable](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository) to `1` which tells the action to use doldecomp's packages instead. The package-per-repository approach is intended to support non-vanilla and other custom forks, as well as vanilla.

---

Could also make the variable's absence default to using doldecomp's packages, and setting it to 1 causes the local packages to build, if you think that makes more sense.

ribbanya (changes requested):
See https://github.com/doldecomp/melee/pull/926#issuecomment-1586620120.

ribbanya:
@r-burns Is this still a pain point for you? Did we ever decide on a good solution?

r-burns:
Yeah I still think it's a little weird that forks build off of the fork's master image rather than the upstream master image. It's not a problem for people who recently forked the repo, but for long-lived forks it's unintuitive to have to keep their master branch up to date with upstream when their master branch serves no purpose otherwise. Not really a big deal but I do want to try to think about this a little more.

ribbanya:
Okay, yeah, I understand your issues and also want to fix them. I need to rethink the CI chain in general as we move to dtk, assuming we want to do that (see Discord thread for discussion).

DRGN-DRC:
Is there anything you still want me to do for this?

ribbanya:
No, it's just something to look at again after dtk.

r-burns:
This is becoming a pain point for me again. The packages are no longer building on my fork and I'm getting spammed with failed action runs every time I push to my fork.

ribbanya:
Okay, I'll think of a stopgap fix and try to implement it today.

ribbanya (approved):
This should ideally disable the `publish-packages` workflow for forks as well, but only if you're not developing the workflow itself, so we'll have to revisit it again after dtk.

@r-burns, merge if this implementation suits your needs.

r-burns:
Thanks, let's give this a try

PR #931, opened 2023-06-21
Split out item logic tables, match itcapsule

r-burns:


src/melee/it/it_803F1418.c
```diff
@@ -0,0 +1,1204 @@
...
+void it_8029B2F4();
+void it_8029B328();
+void it_8029B348();
+bool it_8029B46C();
+bool it_8029B4AC();
+bool it_8029B4EC();
+bool it_8029B52C();
+bool it_8029B56C();
+void it_8029B670();
+void it_8029B6D8();
```
ribbanya, line 426: Are these only used in the tables, or are they called/defined elsewhere? Please use `(void)` instead of `()`.

ribbanya (changes requested):

src/melee/it/it_803F1418.c
r-burns: They are defined in each item's TU, and take a Item_GObj*. These are just temporary, they will be moved to the individual items' headers as those items are matched.

ribbanya (approved):

PR #941, opened 2023-06-27
Match ~79% of hammer item

ItzSwirlz:
src: https://decomp.me/scratch/nF6aS

### DIFF
--- a/obj_files.mk
+++ b/obj_files.mk
@@ -574,7 +574,6 @@ TEXT_O_FILES +=\
     $(BUILD_DIR)/asm/melee/it/items/itfflower.s.o\
     $(BUILD_DIR)/asm/melee/it/items/itkinoko.s.o\
     $(BUILD_DIR)/asm/melee/it/items/itdkinoko.s.o\
-    $(BUILD_DIR)/asm/melee/it/items/ithammer.s.o\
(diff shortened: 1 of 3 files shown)


ribbanya (changes requested):
This needs the asm for non-matching functions to be migrated to C and then the C surrounded in `#if` blocks. I can do it today or tomorrow probably.

jpeisach:
It includes everything, the non-matching functions/the ones that cause compilation errors have the bad lines commenteed

jpeisach:
I found a few things actually hold on

RevoSucks:
update on this?

jpeisach:
yeah it's down to a register swap

jpeisach:
https://decomp.me/scratch/Q1rJA very close

r-burns:
Unless I'm missing something I don't see any regswaps in that scratch. I do see some mis-scheduled epilogues, but I bet those will match with our new 1.2.5n compiler. So if that scratch is the whole file I think you have a 100% match, nice!

ribbanya:
> I bet those will match with our new 1.2.5n compiler

https://decomp.me/scratch/Xaf1U

PsiLupan:
Closing in favor of updated PR that I've created that also matches data.

PR #954, opened 2023-07-03
Remove itHit struct

r-burns:
A number of functions in itcoll pass `&Item->x5D4` to functions which expect a HitCapsule, and known areas of itHit/HitCapsule are identical, so we can merge the structs.

### DIFF
--- a/src/melee/it/types.h
+++ b/src/melee/it/types.h
@@ -275,47 +275,6 @@ struct Article {
     ItemDynamics* x14_dynamics;
 };
 
-struct itHit {
-    bool x0_toggle; // Toggles hitbox on/off.
-    s32 x4_unk;
-    s32 x8_damage;        // Projected damage
-    f32 xC_damage_staled; // Staled damage, actually applied
-    Vec3 x10_offset;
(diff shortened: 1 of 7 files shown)


ribbanya (approved):
Finally lol

src/melee/it/types.h
```diff
@@ -275,47 +275,6 @@ struct Article {
...
-    s32 x8_damage;        // Projected damage
-    f32 xC_damage_staled; // Staled damage, actually applied
-    Vec3 x10_offset;
-    f32 x1C_hitbox_size;
-    s32 x20_angle;
-    s32 x24_knockback_growth;
-    s32 x28_wdsk; // Weight-Dependent Set Knockback
-    s32 x2C_base_knockback;
-    s32 x30_element;       // Normal, fire, electric, etc.
-    s32 x34_shield_damage; // If hitbox damage + shield damage is less than 0
```
r-burns, line 290: It seems to me that this struct and all its named members were copied from m-ex, so I think it's in line with our current approach to remove and independently verify these.

PR #958, opened 2023-07-15
 Remove frank, build with 1.2.5n

r-burns:
ninji is my GOAT :100: :goat:


r-burns:
Oops, need to update the compiler zip + base docker image first.

PR #963, opened 2023-09-13
itHammer matching

PsiLupan:
Builds off previous draft PR to rebase, correct typing issues from struct usages, and matches data.


src/melee/it/items/ithammer.c
```diff
@@ -0,0 +1,244 @@
...
+                                     it_80294080,
+                                     it_8029424C,
+                                     it_8029427C,
+                                 },
+                                 {
+                                     -1,
+                                     it_80294304,
+                                     it_8029430C,
+                                     it_80294310,
+                                 } };
```
ribbanya, line 54: I think removing the commas at the end of each inner array will make the formatting less crazy.

ribbanya (approved):

src/melee/it/items/ithammer.c
PsiLupan: Yeah, I had been sticking with the precedent already set on all the other item files, they just weren't formatted.

PR #1054, opened 2024-01-11
Match most of melee/gr/grT-

seizefire:
- melee/gr/grtfox
- melee/gr/grtkirby
- melee/gr/grtlink
- melee/gr/grtseak
- melee/gr/grtsamus
- melee/gr/grtdrmario
- melee/gr/grtzelda
- melee/gr/grtmario


ribbanya:
Hi, thank you! Seems like nearly all of the errors are due to missing or incorrect function signatures. For example, changing `void grtDrMario_8022050C(void)` to `void grtDrMario_8022050C(bool)` resolves that error.

PR #1124, opened 2024-01-26
Itspycloak

wyatt-avilla:
# `itspycloak.c`

## Non-Matching

Currently, the only thing that doesn't match is the alignment of `it_804DCB70` in `.sdata2`. Objdiff shows the variable in the target aligned to 8 bytes, but mine is aligned to 4.

## Other Stuff

- The casts from `Item_GObj*` to `HSD_GObj*` in the calls to `Item_80268E5C` don't seem to be necessary, but I left them in there, as that's what decomp.me initially generated.
- I padded `it_803F6370` to match the size that was shown in the target in objdiff. The function pointers are guesses based on their declaration and the table in `itstar.c`.


ribbanya:
The casts are just a quirk of our m2c setup. You're correct that they're not necessary.

src/melee/it/items/itspycloak.c
```diff
@@ -0,0 +1,129 @@
...
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+};
```
ribbanya, line 127: You can trim this in `splits.txt`.

ribbanya: But if you look in `asm/melee/it/items/itspycloak.s`, it seems like it actually is longer. I think the function pointers aren't resolved in objdiff because the `.o` is unlinked? Not sure.

```s
.section .data
    .balign 8
.global it_803F6370
it_803F6370:
    .4byte NULL
    .4byte it_80295CB8
    .4byte it_80295CC0
    .4byte it_80295CC4
    .4byte NULL
    .4byte it_80295D2C
    .4byte it_80295D34
    .4byte it_80295D64
    .4byte NULL
    .4byte it_80295DB8
    .4byte NULL
    .4byte NULL
    .4byte NULL
    .4byte it_80295D2C
    .4byte it_80295DE8
    .4byte it_80295E18
    .4byte NULL
    .4byte it_80295E74
    .4byte it_80295E7C
    .4byte it_80295E80
```

ribbanya:
@wyatt-avilla Thank you for contributing! If you want to study my changes, check out 74b584c1b2e68c1f47d842059f35391b6d12b518.

PR #1133, opened 2024-01-26
Fix colored output of tools on Windows

BR-:
> If you’re on a recent version of Windows 10 or better, and your stdout/stderr are pointing to a Windows console, then this will flip the magic configuration switch to enable Windows’ built-in ANSI support.
> 
> If you’re on an older version of Windows, and your stdout/stderr are pointing to a Windows console, then this will wrap sys.stdout and/or sys.stderr in a magic file object that intercepts ANSI escape sequences and issues the appropriate Win32 calls to emulate them.
> 
> In all other circumstances, it does nothing whatsoever. Basically the idea is that this makes Windows act like Unix with respect to ANSI escape handling.

https://pypi.org/project/colorama/

### DIFF
--- a/tools/decomp.py
+++ b/tools/decomp.py
@@ -7,9 +7,6 @@ from pathlib import Path
 from sys import stderr
 from typing import Optional, cast
 
-import colorama
-colorama.just_fix_windows_console()
-
 from elftools.elf.elffile import ELFFile
 from elftools.elf.sections import SymbolTableSection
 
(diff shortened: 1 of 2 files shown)


ribbanya (commented):
What about `requirements.txt`? In those files, I import `pygments` only at the end when outputting. Could you move these statements down to there?

https://github.com/doldecomp/melee/blob/321851ff748770b29079447ebcfbd3c138f4db6f/tools/decomp.py#L139-L144

https://github.com/doldecomp/melee/blob/321851ff748770b29079447ebcfbd3c138f4db6f/tools/m2ctx/m2ctx.py#L195-L203

BR-:
`colorama` and `pygments` are both in `requirements.txt`, but if someone doesn't have them, `--colorize` could still work like this I suppose

PR #1134, opened 2024-01-26
Fix `ninja all_source` build

BR-:


ribbanya (commented):
I'll take a look locally but I don't think this is going to work until `sqrtf_accurate` is sorted out. It needs to go in `math.h` but still embed itself in `lbvector`.

PR #1161, opened 2024-01-31
Initial work on lbarchive

PsiLupan:
* Matches several lbarchive functions
* Adds partial matches for ones that I've got most of the way, but regswaps seem to be an issue
* Adjusts datatypes used in lbfile based on the call to HSD_DevComRequest being set to the proper prototype


ribbanya (approved):
Let me know if you want help resolving the errors.

ribbanya:
Oh, I should mention, don't bother removing symbols from `asm/**/*.s` anymore. Just use `src/**/*.c` as a WIP file until it can be fully linked.

ribbanya:
And you can use `tools/revert_asm.py src/.../file.c` to regen the asm, if for example the splits change.

PR #1164, opened 2024-01-31
Split plstale into a new "plattack" and match some of the functions, setup appropriate function prototypes

PsiLupan:


ribbanya:
[This thing](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) might make your life easier assuming you're using VSCode.

PR #1173, opened 2024-02-01
Match CObj setupTopHalfCamera, bring in existing matches for CObjEraseScreen and CObjLoad

PsiLupan:


ribbanya (commented):
Just move static functions from the `.h` to the top of the `.c`. Put the `static` keyword on the declaration. Remove it or keep it on the definition, doesn't matter.

PR #1192, opened 2024-02-06
Match 74.75% of `ittomato` (1.17 KB)

wyatt-avilla:
## Report of `melee/it/items/ittomato.h`
Function|Size|Score|Max|%
-|-|-|-|-
**File**|`1.17 KB`|`7,372`|`29,200`|`74.75%`
`it_80284324`|`52 bytes`|`1,935`|`1,300`|`0.00%`
`it_802841B4`|`216 bytes`|`4,622`|`5,400`|`14.41%`
`it_802842A4`|`128 bytes`|`815`|`3,200`|`74.53%`
`it_8028454C`|`112 bytes`|`0`|`2,800`|`100.00%`
`it_802844E4`|`92 bytes`|`0`|`2,300`|`100.00%`
`it_802843E4`|`60 bytes`|`0`|`1,500`|`100.00%`
`it_802845F0`|`52 bytes`|`0`|`1,300`|`100.00%`
`it_80284388`|`48 bytes`|`0`|`1,200`|`100.00%`
`it_80284488`|`48 bytes`|`0`|`1,200`|`100.00%`
`it_802843B8`|`44 bytes`|`0`|`1,100`|`100.00%`
`it_8028442C`|`44 bytes`|`0`|`1,100`|`100.00%`
`it_802844B8`|`44 bytes`|`0`|`1,100`|`100.00%`
`it_80284358`|`40 bytes`|`0`|`1,000`|`100.00%`
`it_80284458`|`40 bytes`|`0`|`1,000`|`100.00%`
`it_802845BC`|`40 bytes`|`0`|`1,000`|`100.00%`
`it_80284624`|`32 bytes`|`0`|`800`|`100.00%`
`it_8028428C`|`24 bytes`|`0`|`600`|`100.00%`
`it_80284380`|`8 bytes`|`0`|`200`|`100.00%`
`it_80284420`|`8 bytes`|`0`|`200`|`100.00%`
`it_80284480`|`8 bytes`|`0`|`200`|`100.00%`
`it_80284540`|`8 bytes`|`0`|`200`|`100.00%`
`it_802845E4`|`8 bytes`|`0`|`200`|`100.00%`
`it_80284428`|`4 bytes`|`0`|`100`|`100.00%`
`it_80284548`|`4 bytes`|`0`|`100`|`100.00%`
`it_802845EC`|`4 bytes`|`0`|`100`|`100.00%`

---
# `ittomato`

## Non-Matching

- `it_802841B4`
- `it_802842A4`
- `it_80284324`

### Contain Branch Arrows

- `it_802844E4`
- `it_8028454C`

## Draft

I've marked this as a draft as a bit more work needs to be done. Specifically, I wasn't sure whether the non-matching functions should've had their scopes modified in `symbols.txt` in b8521098ea40583353fc41879e29063e80d8216e. Also, if anything else needs changing, I'm happy to do it.


src/melee/it/items/ittomato.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+ItemStateTable it_803F5740[] = {
+    { 0xFFFFFFFF, it_80284380, it_80284388, it_802843B8 },
+    { 0xFFFFFFFF, it_80284420, it_80284428, it_8028442C },
+    { 0xFFFFFFFF, it_80284480, it_80284488, it_802844B8 },
+    { 0xFFFFFFFF, it_80284540, it_80284548, NULL },
+    { 0xFFFFFFFF, it_80284480, it_80284488, it_802844B8 },
+    { 0xFFFFFFFF, it_802845E4, it_802845EC, it_802845F0 },
+};
```
ribbanya, line 186: This should go at the top of the file, just as a convention.

src/melee/it/items/ittomato.c
```diff
@@ -0,0 +1,186 @@
...
+
+bool it_802845E4(Item_GObj* gobj)
+{
+    return false;
+}
+
+void it_802845EC(Item_GObj* gobj)
+{
+    return;
+}
```
ribbanya, line 166:
```suggestion
void it_802845EC(Item_GObj* gobj) {}
```

src/melee/it/items/ittomato.c
```diff
@@ -0,0 +1,186 @@
...
+
+#include "baselib/jobj.h"
+#include "it/inlines.h"
+#include "it/it_266F.h"
+#include "it/it_26B1.h"
+#include "it/item.h"
+
+#include <placeholder.h>
+
+extern M2C_UNK gm_80473A18;
```
ribbanya, line 11: This goes in its header file, `src/melee/gm/gm_1601.h`. You can use `UNK_T`, which expands to `void*` most of the time and when using M2C it expands to `M2C_UNK`.

src/melee/it/items/ittomato.c
```diff
@@ -0,0 +1,186 @@
...
+HSD_GObj* it_802841B4(s32 arg0, Vec3* arg1, s32 arg2)
+{
+    NOT_IMPLEMENTED;
+}
+
+void it_8028428C(Item_GObj* gobj)
+{
+    Item* ip;
+
+    ip = gobj->user_data;
```
ribbanya, line 42: Declarations can usually be combined without affecting codegen, and stylistically it's preferred.
```suggestion
    Item* ip = gobj->user_data;
```

If it doesn't break the match, this is also preferred:
```suggestion
    Item* ip = GET_ITEM(gobj);
```

ribbanya (dismissed):
> I wasn't sure whether the non-matching functions should've had their scopes modified in symbols.txt in https://github.com/doldecomp/melee/commit/b8521098ea40583353fc41879e29063e80d8216e.

Yeah, modifying the scopes is correct (assuming they're all `static`, I didn't check). And don't worry about it being incomplete, we can merge it anyway.

Take the review comments just as guidance. Merging in matching code is more important than anything, I just thought I'd give some stylistic pointers since you seem to be interested in contributing regularly.

src/melee/it/items/ittomato.c
ribbanya: Also I think the first fields should be expressed as `-1`.

src/melee/it/items/ittomato.c
ribbanya: Oh, and don't worry about the address comments if you don't feel like it. If you happen to use vim, this is my replacement for that:
```
'<,'>s`.\{-}80\(\x\{6\}\).\+`/* \1 */ \0`e | nohl
```

And then I sort with
```
'<,'>!tools/sort-declarations.sh
```

src/melee/it/items/ittomato.c
```diff
@@ -30,54 +32,7 @@ extern M2C_UNK gm_80473A18;
...
-        var_r3 = Item_80268B18(NULL);
-    }
-    if (var_r3 != NULL) {
-        temp_r5 = var_r3->user_data;
-        temp_r5->xDD4_itemVar.Egg.filler[4] |= 0x80;
-        temp_r5->xDD4_itemVar.star.x8 = M2C_BITWISE(f32, arg2);
-    }
-    return var_r3;
+    NOT_IMPLEMENTED;
 }
```
ribbanya, line 36: I didn't check what percent matching this was, but there's no need to do this. `NOT_IMPLEMENTED` is just to have _something_ there, so that `clang` doesn't complain about functions that don't exist.

As long as a function compiles without error, having some code there is better than nothing (or `NOT_IMPLEMENTED`).

wyatt-avilla:
> Yeah, modifying the scopes is correct (assuming they're all `static`, I didn't check). And don't worry about it being incomplete, we can merge it anyway.

Oh, is there a way to determine whether or not the functions should be static? I assumed it was based on whether or not they were in the header.

ribbanya:
Just whether it's referenced outside of the TU. I just use my editor's full-project text search (everything under git) to look for the symbol name in other files. You could also use ripgrep or something, I suppose.

It's definitely one of the areas our tooling is still lacking, though, so don't worry about it too much. Same with the scope in `symbols.txt`.

wyatt-avilla:
> Just whether it's referenced outside of the TU. I just use my editor's full-project text search (everything under git) to look for the symbol name in other files.

oo, I'll be using this. Also, thanks for taking the time to add all the comments and suggestions. I really appreciate it :)

PR #1218, opened 2024-02-14
Match `lbColl_80007DD8`

rjeli:
now just down to stack ordering / padding, did my best but couldn't hack it


rjeli:
not sure what's going on with `lbColl_JObjSetupMatrix`, it checks for null but not dirty. If I change to use `lbColl_JObjSetupMatrix`, it inlines it which is not wanted. seems like compiler doesn't support `__attribute__((noinline))`

ribbanya:
We have access to a few inline-related pragmas: `dont_inline on|off`, `inline_depth(n|smart)`, `always_inline on|off`. You can find most of the pragmas for our compiler on page 273 of [this manual](https://www.nxp.com/docs/en/reference-manual/CWMCUCFCMPREF.pdf).

That's for forcing a match, though. The correct way to fix this kind of issue is almost always by nesting some of the code in a helper function, or nesting that helper function in another function.

In your particular case, you can see the assert line (`0x478U`, or 1144), which is `HSD_JObjGetMtxPtr`:
https://github.com/doldecomp/melee/blob/d5284ac248fecc79f207a880c669421b11e8ce1e/src/melee/lb/lbcollision.c#L1600
https://github.com/doldecomp/melee/blob/16c91b8b5baf5d4fe3c88a367e1ca2191cba6cf2/src/sysdolphin/baselib/jobj.h#L449-L454

This will bring the match up to 90%:
```c
if (hit_transform != NULL) {
    PSMTXConcat(hit_transform, HSD_JObjGetMtxPtr(hit->bone), transformed_hit);
}
```

This is perfectly mergeable as-is, I'll just hold off in case you have any other questions.

ribbanya:
Nice work!

rjeli:
cool thanks, duh should've grepped for assert line number 🤦 100% now!

PR #1227, opened 2024-02-15
Match 49.92% of `lbmemory` (2.73 KB)

rjeli:
## Report of `melee/lb/lbmemory.c`
Function|Size|Score|Max|%
-|-|-|-|-
**File**|`2.73 KB`|`34,205`|`68,300`|`49.92%`
`lbMemory_8001564C`|`644 bytes`|`16,100`|`16,100`|`0.00%`
`lbMemory_80015320`|`412 bytes`|`10,300`|`10,300`|`0.00%`
`fn_80015184`|`280 bytes`|`7,000`|`7,000`|`0.00%`
`lbMemory_800150F0`|`148 bytes`|`615`|`3,700`|`83.38%`
`lbMemory_80014FC8`|`296 bytes`|`180`|`7,400`|`97.57%`
`lbMemory_800154BC`|`24 bytes`|`10`|`600`|`98.33%`
`lbMemory_800154D4`|`208 bytes`|`0`|`5,200`|`100.00%`
`lbMemory_80014E24`|`200 bytes`|`0`|`5,000`|`100.00%`
`lbMemory_800155A4`|`168 bytes`|`0`|`4,200`|`100.00%`
`lbMemory_80014EEC`|`144 bytes`|`0`|`3,600`|`100.00%`
`lbMemory_8001529C`|`132 bytes`|`0`|`3,300`|`100.00%`
`lbMemory_80014F7C`|`76 bytes`|`0`|`1,900`|`100.00%`

---

just some regallocs left.


ribbanya:
Wow, nice. I can clean it up from here.

rjeli:
i mean just regalloc left for what I have, there's still 2 more funcs

ribbanya:
Yeah, would recommend a new PR for the remaining functions.

ribbanya:
Please refer to 5d3bb4bb3cb34960d0aab32de05bfd77f1fcd903 for future changes. As you can see, `handle` is the name of a local variable that's being asserted as truthy.

src/Runtime/platform.h
```diff
@@ -150,6 +150,11 @@ typedef void (*Event)(void);
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
+#define STATIC_ASSERT(cond)                                                   \
+    struct {                                                                  \
+        int x[1 - 2 * !(cond)];                                               \
+    };
+
```
ribbanya, line 157: `clang` doesn't like this. Can you explain how it works?

ribbanya (changes requested):

src/Runtime/platform.h
ribbanya: I don't think we really need it, since the match checksum enforces that kind of thing anyway.

rjeli: oh, didn't realize clang was used too. in the dark ages before `_Static_assert` it was used to error at compile time for constant expressions: https://elixir.bootlin.com/linux/v4.20.17/source/include/linux/build_bug.h#L66

if `cond` is a compile-time expression that evaluates to `0`, it tries to create an array size of `-1` and errors.

declaring it as anonymous struct allows it to be used at toplevel.

rjeli: yeah I figured there were checks elsewhere, but it was just nice for my build-diff loop to immediately get errors when I miscalculate padding. feel free to remove and I'll just keep it in local stash

rjeli: if you really love it though, probably needs some cpp concat `__FILE__ __LINE__` gensym crap so it can create a unique struct name

rjeli: if you want to keep it, though, should be able to `ifdef` swap it for `_Static_assert`

ribbanya: `clang` is used for static analysis (our Issues CI check and editor language servers).

I'm having some internet issues so I'll have to continue this later.

ribbanya: After [way too much googling](https://stackoverflow.com/a/54993033) I found something that seems to work.

ribbanya (approved):

PR #1335, opened 2024-02-27
Support Nix environment

r-burns:
Build `main.dol` in the nix sandbox using `nix-build`.

Enter development environment using `nix-shell`. Unlike the makefile build, the configure.py script doesn't pick up environment variables, so you have to manually specify some locations when configuring. You can `cat $configurePhase` to see what you need to specify.


nix/default.nix
ribbanya, line 1: Please move everything to `.nix` (instead of `nix`). Environment-related utilities should be in hidden directories.

default.nix
ribbanya, line 1: Does this need to exist? Like when I source Python, I do:
```sh
. ./.venv/bin/activate
```

Or when I source vimrc I do
```vim
:source tools/vimc
```

Can this file be removed and your local command updated accordingly?

tools/project.py
```diff
@@ -214,7 +214,9 @@ def generate_build_ninja(
         description="TOOL $out",
     )
 
-    if config.build_dtk_path:
+    if config.build_dtk_path and os.path.isfile(config.build_dtk_path):
+        dtk = config.build_dtk_path
+    elif config.build_dtk_path:
```
ribbanya, line 219: I think this basic idea can work, I'll just clean it up a little.

ribbanya (commented):
I really think it would be simpler if your config phase directly leverages `tools/download_tool.py`, as `configure.py` does, to acquire the binaries. This is more future-proof against our config, and also greatly simplifies your scripts by not needing to build a bunch of random things.

`configure.py` can be given a new `dry-run` mode (it already has `configure`, the default, and `progress`, for printing progress):
```sh
❯ ./configure.py dry-run
{"dtk_tag": "v0.7.2", "compilers_tag": "20231018", "wibo_tag": "0.6.9", "sjiswrap_tag": "v1.1.1", "linker_version": "
GC/1.3.2", "version": "GALE01", "progress_code_fancy_item": "Trophies", "progress_data_fancy_item": "Event Matches"}
```

So the idea is for Nix to run that, parse the json, and then invoke `download_tool.py` itself:

```sh
❯ tools/download_tool.py --tag "v0.7.2" dtk build/tools/dtk        
Downloading https://github.com/encounter/decomp-toolkit/releases/download/v0.7.2/dtk-linux-x86_64 to build/tools/dtk

❯ tools/download_tool.py --tag "20231018" compilers build/compilers
Downloading https://files.decomp.dev/compilers_20231018.zip to build/compilers

# etc
```

Additionally, `configure.py` should be modified so that if a `configure.json` exists, and a particular argument is not passed directly to the script, the default is loaded from the json. `configure.json` would be gitignored and can be generated by `configure.py`. This allows Nix (or similar environments) to provide defaults without the user needing to pass them.

---

Thinking about it more, rather than `dry-run` emitting a json, it could be more specific and emit parameters pairs for `download_tool` to consume, like this:
```sh
❯ ./configure.py tools
dtk v0.6.2
compilers 20231018
```

ribbanya (changes requested):
See https://github.com/doldecomp/melee/pull/1335#pullrequestreview-1903326783 and let me know what you think. I'll handle refactoring the Python scripts if you can refactor the Nix scripts. The main thing would be implementing a nix function to invoke `download_tool.py`.

ribbanya (approved):

PR #1356, opened 2024-02-29
Restore ability to build with Make in nix shell

r-burns:
This only requires some env variables and the compilers symlink, so it's easy to keep support for this.


ribbanya:
At this point it's starting to feel a little silly that we're not just using Nix for all the CI, since there's so much duplicated work being done... Regardless, do you think we should add a Make matrix entry for Nix?

ribbanya (approved):

PR #1377, opened 2024-03-06
Link `itmew` (992 bytes)

wyatt-avilla:
# Notes
- `it_8027ADEC` can probably be declared as `void it_8027ADEC(enum_t, Item_GObj*, HSD_JObj*, f32);`, but I didn't want to change it based on its use in a single file
- `it_802D3E2C` matches, but I feel like it could be changed to be prettier, I'm not sure why there are 2 identical item pointers


wyatt-avilla:
I tried consolidating the 2 variables in `it_802D3E2C`, but no luck. It probably has a similar form to `it_802D3BE0`, but I don't feel like spending more time on something that's already matched.

PR #1380, opened 2024-03-15
`itcoin`

wyatt-avilla:
# notes

- changes in `Ground_801C58E0` due to it returning `it_802F2094`
- changes in `struct Item` due to the accesses in `it_802F1F48`
- `it_802F1630` and `it_802F1CE0` seem to be non-matching for the same reason
- `it_802F1910` is so close to matching, it's just the condition within the first `if` statement

If any matching changes are obvious, please lmk and I'll make them. Otherwise, I'll mark this for review.


src/melee/gr/grfigureget.h
```diff
@@ -1,4 +1,11 @@
 #ifndef GALE01_219530
 #define GALE01_219530
 
+#include <platform.h>
+
+#include <placeholder.h>
+
+s32 grFigureGet_80219C50(s32);
+M2C_UNK grFigureGet_80219C34(s32);
```
ribbanya, line 9: In general, please
* use `int` (or `bool`) instead of `s32`
* use `UNK_T` in place of `M2C_UNK`

ribbanya (changes requested):

src/melee/ty/tydisplay.h
ribbanya, line 1: Tried to do this in one review but GitHub didn't feel like it.

Anyway, this needs an include guard. `gen_header` (from sourcing `tools/functions.sh`) can create one for you, or you can just make one up, but not having one will cause m2ctx to break.

src/melee/ty/tydisplay.h
```diff
@@ -1 +1,10 @@
+#ifndef __TYDISPLAY_H
+#define __TYDISPLAY_H
 
+#include <placeholder.h>
+
+#endif
+
+#include <platform.h>
+
+void* un_8031C5E4(s32);
```
ribbanya, line 10:
```suggestion
#ifndef __TYDISPLAY_H
#define __TYDISPLAY_H

#include <platform.h>

void* un_8031C5E4(s32);

#endif
```

wyatt-avilla: oops, thanks for catching that

ribbanya (approved):

PR #1388, opened 2024-06-11
Match mnCharSel_803F0A48 and mnCharSel_803F0DFC

blueOkiris:
NOTE: This is my first time contributing to the decomp. So please don't hesitate to let me know if I've done anything "out of protocol" :)

## Justification

mnCharSel_803F0A48 is primarily concerned with icon and mode data while mnCharSel_803F0DFC is concerned with the doors at the bottom of the screen.

Both appear to be pieces of data heavily used by other functions in mncharsel, so matching this should accelerate the decompilation of the CSS.

These data structures are also well known and propagated among various mods, Gecko codes, tools, etc.

## Implementation

I created two data structs for each of the memory structures with as much info as I could find then filled in the defaults into a variable in .data (where objdiff says they're located).

## Testing

1. Objdiff shows 100% match for each and proper sizing
2. Through the use of static write mods, I have verified the different mappings to characters to ensure that the names used are accurate as well as offsets.

### DIFF
--- a/src/melee/mn/forward.h
+++ b/src/melee/mn/forward.h
@@ -6,6 +6,14 @@ typedef struct PlayerInitData PlayerInitData;
 typedef struct StartMeleeData StartMeleeData;
 typedef struct StartMeleeRules StartMeleeRules;
 typedef struct VsModeData VsModeData;
+typedef struct CSSModeInfo CSSModeInfo;
+typedef struct CSSIcon CSSIcon;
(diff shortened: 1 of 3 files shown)


blueOkiris:
Changes: I noticed I failed the first formatting check, so I fixed the formatting and redid the commits

ribbanya:
Thank you!

PR #1402, opened 2024-07-26
Partially match `grkongo`

thefoxcam:


thefoxcam:
Not sure what's causing the broken functions error; all but `grKongo_801D828C` haven't had any prior work done on them.

ribbanya:
It means they no longer match. Probably a struct change somewhere.

ribbanya:
Could also be a function signature change.

PR #1406, opened 2024-08-02
m2ctx: Pass `--quiet` to error output

thefoxcam:
Don't print the entire context file on an error if `--quiet` is specified, making it easier to actually see the error message

### DIFF
--- a/tools/m2ctx/m2ctx.py
+++ b/tools/m2ctx/m2ctx.py
@@ -79,17 +79,17 @@ def write_header(path: Path):
     path.write_text("\n".join(files), encoding="utf-8")
 
 
-def try_import(c_command: List[str], args_quiet):
+def try_import(c_command: List[str], quiet: bool):
     try:
         out_text = subprocess.check_output(c_command, cwd=root, encoding="utf8")
     except subprocess.CalledProcessError as err:
-        err_output = "" if args_quiet else err.output
         print(
             "Failed to preprocess input file, when running command:\n"
-            + " ".join(c_command)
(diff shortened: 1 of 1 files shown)


tools/m2ctx/m2ctx.py
```diff
@@ -79,14 +79,15 @@ def write_header(path: Path):
...
+def try_import(c_command: List[str], args_quiet):
     try:
         out_text = subprocess.check_output(c_command, cwd=root, encoding="utf8")
     except subprocess.CalledProcessError as err:
+        err_output = "" if args_quiet else err.output
         print(
             "Failed to preprocess input file, when running command:\n"
             + " ".join(c_command)
-            + f"\n\n{err.output}",
+            + f"\n\n{err_output}",
```
ribbanya, line 90: The trailing newlines are a bit messy. Could either construct an `output` variable, or just do more than one `print` statement with `if not quiet`.

tools/m2ctx/m2ctx.py
```diff
@@ -79,14 +79,15 @@ def write_header(path: Path):
     path.write_text("\n".join(files), encoding="utf-8")
 
 
-def try_import(c_command: List[str]):
+def try_import(c_command: List[str], args_quiet):
```
ribbanya, line 82: Please rename to `quiet: bool` and do the same in the other helper functions.

ribbanya (changes requested):

ribbanya (approved):

ribbanya:
Thanks again. Tooling PRs are really useful.

PR #1421, opened 2024-08-11
Match 34% of `sislib`

thefoxcam:
Will improve names for structs and variables as I work on files in `mn`.


ribbanya:
I haven't looked at it closely yet but I suspect the failed build can be fixed by using `tools/revert_asm.py`.

PR #1431, opened 2024-08-14
Various `mn` work, match 64% of `mnsound`

thefoxcam:


src/melee/mn/mn_2295.h
```diff
@@ -3,6 +3,37 @@
...
 
-/* 22F3D8 */ void mn_8022F3D8(void);
+#include "dolphin/mtx/types.h"
+
+#include <baselib/object.h>
+
+void mn_8022F3D8(void*, u8, HSD_TypeMask);
+void mn_8022ED6C(HSD_JObj*, Vec3*);
+u32 mn_80229624(u32);            // returns a bitfield of controller inputs
+void mn_80229894(s32, s32, s32); // go to a specific part of the menu?
```
ribbanya, line 13: If you change it to the following, it will show up in our [docs](https://doldecomp.github.io/melee/mncharsel_8c.html#a75670d84771edc8747d8d06c341c179a) and in `clangd` intellisense this way.
```suggestion
u32 mn_80229624(u32);            ///< @returns a bitfield of controller inputs
void mn_80229894(s32, s32, s32); ///< go to a specific part of the menu?
```

As a style preference, we would rather keep doc comments in the corresponding `*.dox` file (in this case `mn_2295.dox`), which is the same as .c or .h but only contains comments. But keeping a brief comment here for now is of course fine.

src/melee/mn/types.h
```diff
@@ -6,6 +6,42 @@
...
+    /*  +8 */ Menu_GObj* next;
+    /*  +C */ Menu_GObj* prev;
+    /* +10 */ Menu_GObj* next_gx;
+    /* +14 */ Menu_GObj* prev_gx;
+    /* +18 */ HSD_GObjProc* proc;
+    /* +1C */ void (*rendered)(Menu_GObj* gobj, s32 code);
+    /* +20 */ u64 gxlink_prios;
+    /* +28 */ HSD_JObj* hsd_obj;
+    /* +2C */ Menu* user_data;
+    /* +30 */ void (*user_data_remove_func)(Fighter* data);
```
ribbanya, line 38:
```suggestion
    /* +30 */ void (*user_data_remove_func)(Menu* data);
```

src/sysdolphin/baselib/sislib.h
```diff
@@ -39,19 +39,21 @@ typedef struct SIS {
...
-    f32 xC;
-    f32 x10;
+struct HSD_Text {
+    // these get passed to the text initializer HSD_SisLib_803A5ACC
+    f32 x0;  // global x position, 0,0 at center
+    f32 x4;  // global y position
+    f32 x8;  // scale that happens from the origin after the translation of x0,
+             // x4
+    f32 xC;  // scale that happens on the X axis from the left side of the text
+    f32 x10; // unknown, no effect observed so far
```
ribbanya, line 53: See above comment for explanation.
```suggestion
    f32 x0;  ///< global x position, 0,0 at center
    f32 x4;  ///< global y position
    f32 x8;  ///< scale that happens from the origin after the translation of x0,
             ///< x4
    f32 xC;  ///< scale that happens on the X axis from the left side of the text
    f32 x10; ///< unknown, no effect observed so far
```

ribbanya (approved):

src/melee/mn/mn_2295.h
ribbanya: Also, if you `:source tools/vimrc` in (Neo)vim, you should get Doxygen highlighting.

PR #1437, opened 2024-08-17
Match Dolphin's hio based on changes from Dec 2001 SDK

PsiLupan:


ribbanya (approved):
Not saying it needs to be done in this PR, but I just wanted to make sure you're aware of #1409.

Nice work.

PsiLupan:
> Not saying it needs to be done in this PR, but I just wanted to make sure you're aware of #1409.
> 
> Nice work.

Yeah, I'd been aware of it. Just figured I'd take the existing stuff that Revo had given me, since AFAIR, the HIO there didn't come close to matching as those are the May 2001 and Melee is for some reason closer to Dec 2001 (meaning there was probably a Nov 2001 or something).

PR #1442, opened 2024-09-04
Match mem_funcs from cam's work

PsiLupan:


ribbanya:
I tried ignoring the issues in `compile_flags.txt` but it's an actual clang error and not a warning so it can't be ignored. I also don't know how to exclude the file from the check at the moment.

I also tried refactoring the code to be valid C but wasn't quite able to match it that way.

Not sure what to do about it right now, sorry.

PR #1446, opened 2024-09-07
Update `dtk-template`

ribbanya:


.github/workflows/build.yml
```diff
@@ -119,12 +119,12 @@ jobs:
...
             expected="$root/expected-report.json"
-            actual="$root/report.json"
+            actual="$root/current-report.json"
             changes="$root/changes-report.json"
             head=$(git rev-parse HEAD)
 
-            ninja all_source
-            objdiff-cli report generate -o "$actual"
+            ninja all_source "$root/report.json"
+            cp "$root/report.json" "$actual"
```
ribbanya, line 127: Why the extra file?

encounter: So we can put it back for the artifact upload

ribbanya: @encounter Why can't `actual="$root/report.json`? I actually don't get it, sorry.

The script rolls back to the PR base commit (`master`) _first_ and then it checks out the PR head and builds. So the most recently generated `report.json` should be what we want as the artifact, no?

ribbanya: Also, why is [this error](https://github.com/doldecomp/melee/actions/runs/10904863096/job/30262330201?pr=1446#step:6:917) happening? Shouldn't ninja understand how to build the report in either case? Or did the path change?

encounter: The commit it's checking out doesn't have the updated dtk-template that supports the report.json rule. Since this is in the middle of adding support. May have to ignore that error temporarily and merge so that master branch gets the updated code

PR #1484, opened 2024-10-22
Decompile some camera functions

eperdew:
Decompile some functions in camera.c.

Sadly, a couple of them were only partial matches. That said, I think we made good progress on figuring out the struct layout of `Camera`.


ribbanya:
Looks good, thank you!

PR #1487, opened 2025-01-12
some `mn` work

thefoxcam:


ribbanya:
Nice, are those brand new math lib matches?

thefoxcam:
I didn't try to match them, was mainly interested in `powi` and `lb_8000D148` and decided to include the other funcs in the TU as well to see what they looked like. `lb_8000D008` looks like atan2, but I didn't analyze it long enough to see whether it's also doing something else.

PR #1488, opened 2025-01-13
Item files work

AaronB-Hub:


ribbanya (approved):
This is a lot of nice work!

Not sure why commenting out `it_3F14` is necessary but it can be fixed later.

AaronB-Hub:
Thanks! I wasn't sure if/how all the item tables should be split up across files, so I just put them all together in one file for now.

PR #1512, opened 2025-05-24
Switch Nix CI caching from magic-nix-cache to cachix

r-burns:


r-burns:
Initial run with cold cache succeeded in 8m30s, which was typical for the broken workflow without caching. Pushing to cache seems to have succeeded. Pushing a dummy commit to test out performance with a warmed up cache.

r-burns:
Cool, caching is working again.

Issue #1565, opened 2025-06-04
Remove usage of truncated includes


r-burns:
- [x] #3349 
- [ ] Remove extra include paths from build systems (mwcc, clang, native, m2ctx, doxygen)
- [ ] Remove fancy clang-format sorting (maybe keep some simple rules?)
- [ ] Enforce and/or automatically apply `tools/replace-includes.py`

---
We have a lot of usage of "truncated" include paths in our melee code, such as the following:

https://github.com/doldecomp/melee/blob/ea08f374971a0ce6d628ed97df2eba97e4acdb2b/src/melee/ft/ftparts.c#L3-L10

Because of this, we are including files relative to src/ as well as src/melee/, so both have to be added to the include path.

To be consistent with C coding standards, we should limit the number of include paths that need to be added to the compile-line - so just `src/` for melee code. We should only use quote-includes for paths relative to the current file, and use bracket-includes for fully-qualified includes relative to include paths.

I would propose that the above snippet could look like this:

```c
 #include "forward.h" 
  
 #include "ftparts.h" 
  
 #include "fighter.h" 
 #include "inlines.h" 
 #include "types.h" 
 #include <melee/lb/lbrefract.h> 
```

The choice to use quote-includes for files in the same directory is a matter of taste. One could also fully-qualify all includes, like this:

```c
 #include <melee/ft/forward.h>
  
 #include <melee/ft/ftparts.h>
  
 #include <melee/ft/fighter.h>
 #include <melee/ft/inlines.h>
 #include <melee/ft/types.h>
 #include <melee/lb/lbrefract.h> 
```

ribbanya:
[Twilight Princess](https://github.com/zeldaret/tp/blob/992e323ade82ee2deb2495ac0282f1c6bffc5a9f/configure.py#L232-L242) uses `<>` (system includes) for
- dolphin
- MSL
- MetroTRK
- Runtime

and then it just uses `""` relative to `src` for everything else.

PR #1589, opened 2025-06-11
Match 55% of `plbonuslib`

Avery-Brinkman:
Very new to decomp stuff and C. I'd love to get feedback, including nitpicks.


src/melee/pl/types.h
```diff
@@ -4,6 +4,8 @@
 #include <platform.h>
 #include <placeholder.h>
 
+#include "forward.h"
```
r-burns, line 7: One quirk of our compiler is that relative includes don't work properly, they're always relative to the .c file being compiled. So for example, if this gets included from melee/ft/fighter.c, it will actually include melee/ft/forward.h instead of melee/pl/player.h. I'd suggest fully qualifying this as <melee/pl/forward.h> instead. Not a big deal here but something to keep in mind.

r-burns (approved):
Looked over for nitpicks and everything is looking great. Nice work conforming to the project style and header structure. Thanks!

src/melee/pl/types.h
ribbanya: It is actually a big deal in the sense that it will break eventually, when it's included from a non-`pl` header.

Avery-Brinkman: so sounds like i just shouldn't use relative includes. they'll cause problems in headers, and at that point might as well not use them in source either

ribbanya: It's fine to use relative includes in C files, just not in headers.

ribbanya: But yeah, we're probably going to phase them out anyway (#1565).

PR #1605, opened 2025-06-17
Random work for item investigations

AaronB-Hub:
Not sure if small, scattered PR's like this are annoying


ribbanya:
They're not annoying, it's beneficial to merge them so that conflicts don't occur later.

AaronB-Hub:
Hmm, I'm stumped on fixing this error in mnstagesel.c. Probably missing something obvious. Not sure why it thinks the first call wants an 'int' return type.

src/melee/mn/mnstagesel.static.h
```diff
@@ -1,5 +1,5 @@
-#ifndef GALE01_2599EC
-#define GALE01_2599EC
+#ifndef __GALE01_2599EC
+#define __GALE01_2599EC
```
ribbanya, line 2: The guard was the same as `mnstagesel.h` so that file was being skipped, which was breaking the forward declaration.

PR #1606, opened 2025-06-19
Matching functions in ftSk_SpecialHi.c

Gelatart:
Used a scratch to figure out ftSk_SpecialAirHiStart_0_Phys, was able to find that kipcode66 had old scratches where they decompiled and matched several other functions not implemented in code yet, ftSk_SpecialAirHiStart_1_Coll fully matches in his scratch but doesn't seem to here yet, but it is getting over 50%

I made some changes to _ftSeakAttributes to get my scratch to work, and ftSk_SpecialAirHiStart_1_Coll seems to need further changes to get its results, and they don't seem to break anything anywhere else

### DIFF
--- a/src/melee/ft/chara/ftSeak/ftSk_SpecialHi.c
+++ b/src/melee/ft/chara/ftSeak/ftSk_SpecialHi.c
@@ -135,9 +135,22 @@ void ftSk_SpecialAirHi_Enter(HSD_GObj* gobj)
 
 /// #ftSk_SpecialAirHi_Enter
 
-/// #ftSk_SpecialHiStart_0_Anim
+// Gelatart's scratch at https://decomp.me/scratch/ysptk
+void ftSk_SpecialHiStart_0_Anim(HSD_GObj* gobj)
+{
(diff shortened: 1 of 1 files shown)


config/GALE01/build.sha1
```diff
@@ -1 +1 @@
-08e0bf20134dfcb260699671004527b2d6bb1a45  build/GALE01/main.dol
\ No newline at end of file
+08e0bf20134dfcb260699671004527b2d6bb1a45  orig/GALE01/sys/main.dol
```
ribbanya: Why is this path changed?

Gelatart: I think that was an oversight, I ran a dtk shasum command from a dtk-template guide again, and forgot to do the step afterward where I correct that

Gelatart: Cause there was a test complaining about how I set up this sha1 file that I didn't fully understand what it wanted

Gelatart: Gonna try to commit to fix this back

ribbanya:
Nice, thanks!

PR #1614, opened 2025-06-20
ftSk_SpecialAirHi Functions Implemented

Gelatart:
Implemented the rest of the ftSk_SpecialAirHi functions through decompiles, often comparing to similar functions for Zelda

### DIFF
--- a/src/melee/ft/chara/ftSeak/ftSk_SpecialHi.c
+++ b/src/melee/ft/chara/ftSeak/ftSk_SpecialHi.c
@@ -475,7 +475,26 @@ void ftSk_SpecialHi_Phys(HSD_GObj* gobj)
     ft_80084F3C((Fighter_GObj*) gobj);
 }
 
-/// #ftSk_SpecialAirHi_Phys
+// Gelatart's scratch at https://decomp.me/scratch/rQtEW
+void ftSk_SpecialAirHi_Phys(HSD_GObj* gobj) {
+    u8 _[8];
+
+    Fighter* fp = GET_FIGHTER(gobj);
+    ftSeakAttributes* attributes = fp->dat_attrs;
+    
+    if (fp->cmd_vars[0] != 0) {
+        ftCommon_8007D4B8(fp);
+        ftCommon_8007D440(fp, attributes->x4C * fp->co_attrs.air_drift_max);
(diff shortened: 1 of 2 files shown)


src/melee/ft/chara/ftSeak/ftSk_SpecialHi.c
```diff
@@ -22,7 +22,10 @@
 #include <math.h>
 #include <baselib/gobj.h>
 
-/* 08A2BC */ static void ft_8008A2BC(Fighter_GObj* gobj);
+/* 08A2BC */ void ft_8008A2BC(Fighter_GObj* gobj);
+/* 0D5CB0 */ void ftCo_800D5CB0(ftCo_GObj*, bool, float landing_lag);
+/* 096900 */ void ftCo_80096900(ftCo_GObj* gobj, int, int,
+                                bool allow_interrupt, float, float);
```
ribbanya: Please include the headers that these functions belong to, rather than declaring them here. For example, `ft_8008A2BC` is from `ft/ft_0892.h`.

src/melee/ft/chara/ftSeak/ftSk_SpecialHi.c
```diff
@@ -450,7 +453,19 @@ void ftSk_SpecialHi_Anim(HSD_GObj* gobj)
     }
 }
 
-/// #ftSk_SpecialAirHi_Anim
+// Gelatart's scratch at https://decomp.me/scratch/2S6Ll
```
ribbanya, line 454: Just letting you know, while some comments with scratch links have made it into the codebase, this is not a pattern that we encourage. If there is missing information in the scratch, the best way to add it to the code is through comments. Otherwise, there's not really a reason to link a scratch. If the concern is attribution, that's what git authorship is for.

Gelatart: Alright, just thought I'd err on the side of doing it to be safe, I'll refrain going forward

src/melee/ft/chara/ftSeak/ftSk_SpecialHi.c
Gelatart: Fixing that and pushing

PR #1619, opened 2025-06-21
Defined some Game and Watch items

djad99:
Made some initial strides towards contributing to decompiling items and have matched 23 functions and increased the overall match rate by 0.04% (according to two ninja reports -- one on my local and one on master).


r-burns (commented):
the Issues check is failing because of missing declarations - add these functions' declarations to their .h files (and include lb_00B0.h when you use lb_8000BA0C) and all the checks should then be passing

PR #1642, opened 2025-06-26
Match 100% itpeachtoad.c

darsilrod:


src/melee/it/items/itpeachtoad.c
```diff
@@ -1,25 +1,179 @@
...
+        item = GET_ITEM(item_gobj);
+        itpeachtoad_inline_1(item);
+        return true;
+    }
+    return false;
+}
+
+void it_802BE100(Item_GObj* item_gobj)
+{
+    HSD_JObj* hsd_jobj;
```
ribbanya, line 149: You can just call this `jobj`.

ribbanya:
Nice work, thanks for the contributions.

PR #1653, opened 2025-06-29
gmmainlib matches

r-burns:


src/melee/gm/gmmain_lib.c
```diff
@@ -5,6 +5,70 @@
...
+    0,  // mode
+    2,  // time limit
+    3,  // stock count
+    0,  // handicap
+    10, // damage ratio
+    0,
+    0,     // stock time limit
+    false, // friendly fire
+    true,  // pause
+    0,     // score display
```
ribbanya, line 28: Is there a better way to do this in C?

r-burns: do what?

ribbanya: The initializer comments.

r-burns: If we were using a compiler that wasn't from the stone age I'd use [designated initializers](https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html) but MWCC doesn't support those :smiling_face_with_tear:

PR #1655, opened 2025-06-30
More itmball.c matches

MikeBooon:
Here are some more matches for itmball.c.


src/melee/it/items/itmball.c
```diff
@@ -40,20 +54,52 @@ bool it_8029799C(void)
 
 void it_802979A4(void) {}
 
-/// #it_802979A8
+s32 it_802979A8(Item_GObj* arg0)
```
ribbanya: This needs to return `bool` to match the function typedef.

MikeBooon: I have updated the types. Build succeeds now. Let me know if anything else needs updating.

objdiff-macos-arm64
r-burns, line 1: oops, looks like this file got committed by accident, but I think that's the only reason the CI is failing at the moment

MikeBooon: Good catch, my bad. Kept it in the the repo dir while I was using it and committed it by mistake.

PR #1656, opened 2025-06-30
Some item matches.

darsilrod:


src/melee/it/items/it_27CF.h
```diff
@@ -70,7 +70,7 @@
 /* 2C6DB8 */ void it_802C6DB8(Item_GObj*);
 /* 2C6DD8 */ void it_802C6DD8(Item_GObj*);
 /* 2C6E50 */ void it_802C6E50(Item_GObj*);
-/* 2C720C */ HSD_GObj* it_802C720C(Item_GObj*, Vec3*, s32, float);
+/* 2C720C */ HSD_GObj* it_802C720C(Item_GObj*, Vec3*, Fighter_Part, float);
```
r-burns:
```suggestion
/* 2C720C */ HSD_GObj* it_802C720C(HSD_GObj*, Vec3*, Fighter_Part, float);
```

I think the docs build is failing because this Item_GObj argument is inconsistent with the HSD_GObj argument used in this function's definition in its .c file. I would assume the item's `parent` gobj is actually a game&watch Fighter_GObj, but if we're unsure we can just call it a generic HSD_GObj for now

PR #1659, opened 2025-07-01
Almost finish matching `ft_0C88.c`

Giga-Bowser:
I can't figure out how to fix the stack mismatch with `ftCo_800C92E4`


Giga-Bowser:
Now I'm just very confused, I'll take a look

Giga-Bowser:
I'll just keep working on the partial match of `ftCo_800C92E4` separately

ribbanya:
@Giga-Bowser One of the structs is wrong. It's possible all of the assumed field names in `lbColl_80008D30_arg1` are wrong. A lot of bad assumptions have been made about collisions in the codebase based on modder foreknowledge. In general, if new matches disagree with existing headers, the headers are wrong.

See e50848f7569898954c7517d87c19708e2d906910 for how I addressed it and feel free to do similar things in the future.

PR #1673, opened 2025-07-09
Minor changes

AaronB-Hub:


AaronB-Hub:
Seems to be complaining that I changed it_804D6D24, it_804D6D30, and it_804D6D38 from Article** to Article* []. Not sure why that would be (it_804A0F60 doesn't complain)

r-burns:
`Article*[]` is itself a contiguous region of `Article*` pointers, but `Article**` is a pointer to such a region. This means that `Article**` can be stored in sbss but `Article*[]` may be large enough that it needs to go in ordinary bss. So those two types are not ABI equivalent and are breaking matches.

PR #1690, opened 2025-07-14
Add translation unit support for decomp.py

thefoxcam:
Added support to decomp.py for reading relative translation unit paths (i.e. `melee/it/items/itheiho`).

Changes:

- Either `--function / -f` or `--translation-unit / -tu` must be supplied instead of `function` being the default argument, verified with an argparse mutually-exclusive group
- I didn't implement `--write` with a TU input, as I don't use the feature myself.
- I changed the `--format` shorthand  from `-f` to `-fm` to give priority to `--function` .


r-burns:
Can `--function` be the default behavior, since that's the way the script currently works? I would also imagine that the script could automatically know if you're providing a file path instead of a function name, since function names can't contain slashes.

thefoxcam:
I was worried that relying on checking for a slash would make it not possible to input a TU from the root of the source directory, (though there's probably no case that would be needed for Melee) but now I realize we can unambiguously determine whether a function or TU is intended since `has_function` checks the symbols smartly and we can simply rely on `find_obj` returning None for a TU input.

PR #1707, opened 2025-07-19
make grtest equivalent

MarioLover63:


r-burns (commented):
Looks like our CI is complaining about a few minor linting issues:

* tabs instead of spaces, and trailing whitespace (adding an [editorconfig](https://editorconfig.org) plugin to your editor will prevent this in the future)
* missing include in grtest.h for the Vec3 type
* missing include in grtest.c for some gobj utility functions
* parameter names can't be omitted, I just name the parameter "unused" in this case

Otherwise looks great functionalitywise, happy to take care of the linter issues myself if the CI errors are confusing

PR #1710, opened 2025-07-20
/ef work (eflib could still use some love)

AaronB-Hub:


AaronB-Hub:
Not sure how to fix these errors

AaronB-Hub:
Not sure how to fix remaining errors

AaronB-Hub:
Is there a way to force the FighterKind enum to be s8's instead of int's? That seems to be what's causing the issue. I could remove all of the type 'fixing', but feel like it shouldn't be necessary

r-burns:
Not that I'm aware of - I'm afraid it might be easiest to switch the FighterKind types back to s8 for now.

There is an `-enum min` option to mwcc which seems to correctly treat the FighterKind enum as an s8, but unfortunately it seems apply to all enums. It might even be correct for melee to use this option, but it would surely break a ton of matches at the moment and would be hard to switch over to.

ribbanya:
@r-burns I haven't looked into this case, but I remember lblanguage enums being treated as `s8` by the asm in certain cases but declared as `int`, could that be what's happening here?

PR #1715, opened 2025-07-21
Most ftYs_SpecialAirSLoop_IASA functions matched

Gelatart:
ftYs_SpecialAirSLoop_0_IASA still remaining but it seems to be a bigger one


src/melee/ft/chara/ftYoshi/ftYs_SpecialN.c
```diff
@@ -586,7 +587,21 @@ void ftYs_SpecialAirSStart_0_IASA(ftYs_GObj* gobj) {}
...
+void ftYs_SpecialAirSLoop_1_IASA(ftYs_GObj* gobj)
+{
+    Fighter* fp = GET_FIGHTER(gobj);
+    ftYoshiAttributes* attributes = fp->dat_attrs;
+
+    s32 x48 = attributes->x48;
+    s32 x4C = attributes->x4C;
+    s32 difference = x48 - x4C;
+
+    if ((fp->mv.ys.specialn.x0.x0 < difference) &&
```
r-burns: Based on the function names, shouldn't these be using `specials` attrs?

Gelatart: Yoshi doesn't seem to have specials attributes, should we make some? Or are they somewhere else?

PR #1717, opened 2025-07-22
gmtou matches

sadkellz:


src/melee/gm/gm_17C0.c
```diff
@@ -1,3 +1,4 @@
+#include "gm/gmtou.static.h"
 #include "gm_17C0.h"
```
r-burns: small point, but I like to have each .c file include its own .h file at the top before any other includes as a guarantee that its .h file works standalone without needing any other declarations

src/melee/gm/gm_17C0.h
```diff
@@ -6,6 +6,7 @@
 #include <melee/gm/forward.h>
 #include <melee/it/forward.h>
 #include <sysdolphin/baselib/forward.h>
+#include "gm/gmtou.static.h"
```
r-burns: Ah, this is the issue. .static.h headers should only be included from .c files because they contain static definitions - any files that include this header have these new static definitions, breaking their data/bss sections

PR #1742, opened 2025-07-26
gmcamera work + minor scene handler update

sadkellz:


src/melee/gm/gmscdata.c
```diff
@@ -31,315 +32,315 @@
...
-        MJ_DEBUG_SOUND_TEST,
+        MN_DEBUG_MENU,
         NULL,
         gm_801A632C_OnEnter,
         NULL,
         NULL,
     },
     {
-        MJ_HANYU_CSS,
+        MN_CSS,
```
r-burns, line 84: Good catch! I was wondering why these callbacks didn't make any sense, should have realized they are minor scene ids

PR #1749, opened 2025-07-27
Decompile 3 functions from Toy.s

ChaseHaymond:
Decomped 3 toy.s functions. This was my first time decomping so these were a good learning experience. I want to merge these 3 functions now because I'm not sure when ill be able to work on these again.

In the future ill probably just make new issues with decomp.me links.


src/sysdolphin/baselib/controller.c
```diff
@@ -43,6 +43,11 @@ s32 HSD_PadGetResetSwitch(void)
     return (p->reset_switch != 0) ? true : false;
 }
 
+float HSD_PadGetNmlStickY(u8 slot)
```
PsiLupan: Let's move this function out of controller.c into controller.h - similar to how we've done with inlines in jobj.h.

Reasoning being putting this function there that doesn't actually exist is shifting things and broke 100% matches, as noted by the bot.

ChaseHaymond: Moved that function from controller.c and moved it to an inline function in controller.h.

PsiLupan:
There's a few things called out by the Issues build that you'll want to address. If you need help, just ask in the discord.

PR #1767, opened 2025-07-29
itdrmariopill.c matches

lukechampine:
First 100% match lfg

### DIFF
--- a/src/melee/it/items/itdrmariopill.c
+++ b/src/melee/it/items/itdrmariopill.c
@@ -8,7 +8,9 @@
 #include "placeholder.h"
 #include "platform.h"
 
+#include "baselib/gobj.h"
 #include "baselib/jobj.h"
+#include "db/db.h"
 #include "dolphin/mtx.h"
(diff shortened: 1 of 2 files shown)


src/melee/it/items/itdrmariopill.c
```diff
@@ -285,6 +355,97 @@ static s32 it_802C0E48(Item_GObj* gobj)
...
+
+static void it_802C0DF8_no_inline2(Item_GObj* gobj)
+{
+    it_802C0DF8_no_inline(gobj);
+}
+
+static void it_802C0DF8_no_inline3(Item_GObj* gobj)
+{
+    it_802C0DF8_no_inline2(gobj);
+}
```
lukechampine, line 403: surely there is a less insane way to do this...?

ribbanya: Yeah, by finding the original call chain. Maybe it's calling something from Mario fireball or elsewhere, which is calling something else.

PR #1769, opened 2025-07-30
Item type cleanup

AaronB-Hub:


src/melee/it/items/itwhitebea.c
```diff
@@ -45,10 +45,12 @@ bool it_802E3588(void)
 
 /// #it_802E3784
 
+// This function only called by it_8028F2B4 from itfreeze, so using those item
+// vars
```
r-burns, line 49: Hmm, but that freeze function isn't calling this on the freeze item itself, it's calling this on the "linked" item. Are you sure this shouldn't be whitebear here?

AaronB-Hub: You're right that the linked item could be the bear. it_8028F2B4 is the "picked_up" callback for itfreeze so I was assuming that it would just involve the freeze item. But I guess without the spawn callback decomp'ed there's no way of knowing if there isn't some interaction between these items.

Something to note is that it looks like the whitebear file has some other items' functions in there as well (looks like Ottosea).

AaronB-Hub: Still think it's probably more likely to be the itfreeze item than whitebear

r-burns: Ah ok, let's just stick with this for now then, can always revisit with we decomp whitebear a bit more.

PR #1770, opened 2025-07-30
gmtou matches

sadkellz:


src/melee/gm/gm_18A5.h
```diff
@@ -2,9 +2,11 @@
 #define MELEE_GM_18A5_H
 
 #include <placeholder.h>
+#include <melee/ft/forward.h>
 #include <sysdolphin/baselib/forward.h>
 #include <melee/gm/forward.h>
 #include <melee/sc/forward.h>
+#include <melee/mn/types.h>
```
r-burns: Can we avoid including types.h from headers? I think that would just require moving CSSIconHud to a forward.h header here

sadkellz: o7

PR #1774, opened 2025-07-30
some work on un/

BR-:


BR-:
Actual regressions:
```
          un_802FE260 | fuzzy_match  |  72.73% ->  59.74%
        DevText_Setup | fuzzy_match  |  76.32% ->  10.14%
          fn_803020F8 | fuzzy_match  | 100.00% ->   0.00%
```
Not actually sure what happened to `un_802FE260`, `DevText_Setup` is inlining something it shouldn't (should be fixed with better text splits probably), third wasn't a real function and is deleted.

PR #1776, opened 2025-08-01
Infer function types from ItemStateTables

lukechampine:
Apologies for the big diff. 😅 I wrote a script that:

- Looks for declarations of `ItemStateTable`
- Finds the table data in `.s` files and parses out the function pointers
- Fixes the signatures of any functions that disagree with the type required by `ItemStateTable`

This revealed a few errors (e.g. `it_802A12FC` returning an `f32` instead of `void`), but nothing major.

Initially I also tried to emit the table itself and append it to the appropriate `.c` file, but this turned out to be more fraught than I expected. Often the table is referencing functions that are defined elsewhere, so I'd need to add the necessary includes too. It's definitely doable, but I figured I'd start with this.


src/melee/it/items/itnesspkthundertrail.h
r-burns, line 1: CI is telling me this file needs to include baselib/gobj.h now

lukechampine: fixed

PR #1784, opened 2025-08-03
Add tools/table-typer and infer more types

lukechampine:
Confession: I vibe-coded the script I used for #1776. It worked fine, but I figured there was a decent chance we'd want to extend or modify the script in the future, so I rewrote it the old-fashioned way. As a plus, it picked up more signatures that could be fixed, so I've included those as well.


lukechampine:
The linter is apparently very unhappy that I wrote a tool in Go instead of Python 😅

lukechampine:
done

ribbanya (approved):

PR #1795, opened 2025-08-05
Just about match ftCo_DamageIce and improve used names

AlexanderHarrison:


r-burns:
Fixed conflicts with https://github.com/doldecomp/melee/pull/1802 and fixed the regression in ftCo_80091854. Looks good to go, if you're done here.

ribbanya (approved):

PR #1810, opened 2025-08-08
Improve all ItemStateTable function names

lukechampine:
Added a bulk-renaming command to `table-typer` that takes a type name (in this case, `ItemStateTable`), finds all the functions that are fields of that type, and renames them based on a pattern (in this case, `it + [Capitalized filename] + _UnkMotion(N)_(Anim|Phys|Coll)`. Thus, `it_8027D110` becomes `itCapsule_UnkMotion0_Anim`, etc.

Now that the basic machinery works, it should be easy add new types. For now, I just did `ItemStateTable` because I figured we might want to bikeshed on the best way to name these functions.


r-burns (dismissed):
LGTM, wasn't really involved in the discussion on this so will defer to @ribbanya for final approval on this

src/melee/it/items/itkirbycutterbeam.h
```diff
@@ -9,9 +9,9 @@
 
 /* 29BAB8 */ UNK_RET it_8029BAB8(UNK_PARAMS);
 /* 29BB90 */ UNK_RET it_8029BB90(UNK_PARAMS);
-/* 29BC88 */ bool it_8029BC88(Item_GObj* gobj);
-/* 29BCA8 */ void it_8029BCA8(Item_GObj* gobj);
-/* 29BDD4 */ bool it_8029BDD4(Item_GObj* gobj);
+/* 29BC88 */ bool itKirbycutterbeam_UnkMotion_Anim(Item_GObj* gobj);
+/* 29BCA8 */ void itKirbycutterbeam_UnkMotion_Phys(Item_GObj* gobj);
+/* 29BDD4 */ bool itKirbycutterbeam_UnkMotion_Coll(Item_GObj* gobj);
```
ribbanya: Is this a bug with the typer?

src/melee/it/items/itlizardon.h
```diff
@@ -38,9 +38,9 @@
 /* 2CC5C4 */ bool it_802CC5C4(Item_GObj*);
 /* 2CC5CC */ bool it_802CC5CC(Item_GObj*);
 /* 2CC5D4 */ UNK_RET it_802CC5D4(UNK_PARAMS);
-/* 2CC650 */ bool it_802CC650(Item_GObj* gobj);
-/* 2CC684 */ void it_802CC684(Item_GObj* gobj);
-/* 2CC6C4 */ bool it_802CC6C4(Item_GObj* gobj);
+/* 2CC650 */ bool itLizardon_UnkMotion_Anim(Item_GObj* gobj);
+/* 2CC684 */ void itLizardon_UnkMotion_Phys(Item_GObj* gobj);
+/* 2CC6C4 */ bool itLizardon_UnkMotion_Coll(Item_GObj* gobj);
```
ribbanya: Here too.

src/melee/gr/types.h
```diff
@@ -149,8 +149,8 @@ typedef struct StageData {
     char* data1;
     void (*callback0)(void);
     void (*callback1)(int);
-    void (*callback2)(void); ///< on load callback
-    void (*callback3)(void); ///< on GO! callback
+    void (*OnLoad)(void);
+    void (*OnStart)(void);
```
ribbanya, line 153: These should be `on_load` and `on_start` because they are struct fields.

ribbanya (commented):
Overall really good. Mostly just wondering why some states don't have a number.

A lot of files need to tweak their prefix to be camel case (`itClimbersice` -> `itClimbersIce`, etc.) but I don't expect the typer to handle this and it can easily be done in bulk by hand later.

ribbanya (changes requested):
Waiting to resolve the missing state numbers.

src/melee/gr/types.h
ribbanya: Actually it looks like this strange precedent was set elsewhere (by @r-burns?), so we can address it later.

lukechampine:
I added number suffix only when necessary to disambiguate, i.e. if there's only one set of callbacks in the table, we don't add a suffix. I'll change it to add the suffix unconditionally; it would simplify the code a bit anyway.

Another wrinkle is handling multiple tables within the same file. Currently we just skip both, leaving them unchanged, but maybe this could be improved somehow.

lukechampine:
Also, there are a few other mass-renames I could do; would it be preferable for each to be a separate PR, or to group them all together?

lukechampine:
Decided to knock out the other renames while I'm at it

src/melee/gm/gmmovieend.c
```diff
@@ -21,7 +21,7 @@
 #include <melee/mn/types.h>
 #include <melee/un/un_2FC9.h>
 
-void gm_801A6254(MinorScene* arg0)
+void gmTiny_UnkScene12_Decide(MinorScene* arg0)
```
r-burns: This isn't a gmtiny function, it's a callback used as many minor scenes' vs mode exit callback. I think you need to take a look at how these gm callbacks are being named, most of the gmTiny prefixes look incorrect.

src/melee/gm/gmclassic.c
```diff
@@ -4,308 +4,308 @@
...
-        gm_801B3A34,
-        gm_801B3B40,
+        gmClassic_UnkScene42_Prep,
+        gmClassic_UnkScene42_Decide,
         {
             MN_VS,
-            &gm_80472AF8,
-            &gm_8047086C,
+            &gmClassic_UnkScene43_Prep,
+            &gmClassic_UnkScene43_Decide,
```
r-burns: * These aren't unknown, they're for the VS mode minor scene.
* I don't think Prep/Decide are the correct names here, it should be enter/leave, load/unload, something like that.

r-burns:
> Would it be preferable for each to be a separate PR, or to group them all together?

It seems like the item and stage tables are looking good but we'll need to iterate on the Scene handling tables some more, maybe better to split that into a separate PR

lukechampine:
ok, removed the MinorScene commit, easy to revisit later

ribbanya:
I suspect that if a file has multiple tables, it's split improperly. So no need to address it for now.

ribbanya (dismissed):

ribbanya:
Going to need some tweaking later on. For example, I think `*_UnkStage0_*` is fake, but it's not a big deal for now.

r-burns:
> I suspect that if a file has multiple tables, it's split improperly.

Could very well be true for item / stage code, but do note that gmmultiman is correctly split at the moment and contains multiple MinorScene tables for 10-man, 100-man, etc.

PR #1831, opened 2025-08-12
gmcamera work

lukechampine:
Working my way through the remaining tables with UNK_RET functions, and decomping a few functions as I go. 28 left!


src/melee/gm/gmcamera.h
```diff
@@ -35,22 +42,44 @@ typedef struct _gmCameraUnkStruct {
...
-    /*0x44*/ u32 x44;
-    /*0x48*/ HSD_Text* x48;
-    /*0x4C*/ u8 pad_x4C[0x54 - 0x4C];
+    /*0x44*/ s32 x44;
+    /*0x48*/ HSD_Text* x48[3];
     /*0x54*/ u32 x54;
 } gmCameraUnkStruct;
 
-typedef struct _gmCameraUnkStructPtrs {
+typedef struct _gmCameraUnkStruct2Ptrs {
```
ribbanya, line 50: I know you didn't set this precedent, but just so you don't propagate it further, we don't prefix struct names with underscores anymore. So it would be
```c
typedef struct gmCameraUnkStruct2Ptrs {
    // ...
} gmCameraUnkStruct2Ptrs;
```

But, because we also place struct definitions in `types.h` and declarations in `forward.h`, the correct way of declaring would be:
```c
// gm/types.h
struct gmCameraUnkStruct2Ptrs {
    // ...
};

// gm/forward.h
typedef struct gmCameraUnkStruct2Ptrs gmCameraUnkStruct2Ptrs;
```

It is a bit tedious but we do it to comply with doxygen, to avoid circular header dependencies, and to avoid pulling in full struct definitions just to declare a function signature.

Don't worry about it for this commit, just please follow the above practice in the future.

PR #1832, opened 2025-08-12
grcastle work

lukechampine:
Two more tables down, and a few small functions.

### DIFF
--- a/src/melee/gr/grcastle.h
+++ b/src/melee/gr/grcastle.h
@@ -1,13 +1,22 @@
 #ifndef GALE01_1CD338
 #define GALE01_1CD338
 
+
+#include "lb/forward.h"
+#include "gr/forward.h"
 #include <placeholder.h>
-
-#include "it/inlines.h"
-
+#include <dolphin/mtx.h>
 #include <baselib/forward.h>
 
-#include <dolphin/mtx.h>
+typedef struct unkCastle {
+    /* 0x000 */ u8 _pad[0x10C];
+    /* 0x10C */ HSD_GObj* x10C[5];
+    /* 0x120 */ u8 _pad2[0x134-0x120];
+    /* 0x134 */ u8 x134[5];
+} unkCastle;
+
+typedef void (*unkCastleCallback)(UNK_T, unkCastle*);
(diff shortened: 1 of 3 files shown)


src/melee/gr/grcastle.h
```diff
@@ -1,13 +1,22 @@
...
-#include <dolphin/mtx.h>
+typedef struct unkCastle {
+    /* 0x000 */ u8 _pad[0x10C];
+    /* 0x10C */ HSD_GObj* x10C[5];
+    /* 0x120 */ u8 _pad2[0x134-0x120];
+    /* 0x134 */ u8 x134[5];
+} unkCastle;
+
+typedef void (*unkCastleCallback)(UNK_T, unkCastle*);
+typedef void (*unkCastleCallback2)(UNK_T, unkCastle*, Ground_GObj*);
```
lukechampine: not sure if this is the right place to define these structs/functions...?

ribbanya: `gr/forward.h`, and forward-declare `unkCastle` there.

src/melee/gr/grcastle.h
```diff
@@ -8,15 +8,7 @@
...
-typedef struct unkCastle {
-    /* 0x000 */ u8 _pad[0x10C];
-    /* 0x10C */ HSD_GObj* x10C[5];
-    /* 0x120 */ u8 _pad2[0x134-0x120];
-    /* 0x134 */ u8 x134[5];
-} unkCastle;
-
-typedef void (*unkCastleCallback)(UNK_T, unkCastle*);
-typedef void (*unkCastleCallback2)(UNK_T, unkCastle*, Ground_GObj*);
+typedef struct unkCastle unkCastle;
```
ribbanya: Other way around; this goes in `forward` and the struct goes here.

ribbanya (changes requested):

src/melee/gr/grcastle.h
lukechampine: lol yup, just realized

ribbanya (approved):

PR #1834, opened 2025-08-13
ifstatus work

lukechampine:


src/melee/if/ifstatus.c
```diff
@@ -538,16 +551,51 @@ void ifStatus_802F6E3C(s32 player_num)
...
+    lbArchive_LoadSections(*archive, (void**) &desc, "ScInfCnt_scene_models", 0);
+
+    ifStatus_803F9628[0].x14 = M2C_FIELD(&desc, s32**, 0)[0];
+    ifStatus_803F9628[1].x14 = M2C_FIELD(&desc, s32**, 0)[1];
+    ifStatus_803F9628[2].x14 = M2C_FIELD(&desc, s32**, 0)[2];
+    ifStatus_803F9628[3].x14 = M2C_FIELD(&desc, s32**, 0)[3];
+    ifStatus_803F9628[4].x14 = M2C_FIELD(&desc, s32**, 0)[4];
+    ifStatus_803F9628[5].x14 = M2C_FIELD(&desc, s32**, 0)[5];
+    ifStatus_803F9628[6].x14 = M2C_FIELD(&desc, s32**, 0)[6];
+    ifStatus_803F9628[7].x14 = M2C_FIELD(&desc, s32**, 0)[7];
```
lukechampine: not sure what's going on here 🤔

src/melee/if/ifstatus.c
```diff
@@ -538,16 +551,51 @@ void ifStatus_802F6E3C(s32 player_num)
 
 /// #ifStatus_802F7034
 
-/// #ifStatus_802F7134
+void ifStatus_802F7134(void)
+{
+    u32 _pad;
+    HSD_Archive** archive;
+    SceneDesc* desc;
+    PAD_STACK(20);
```
lukechampine: 20 is a ton of extra stack. `SceneDesc` is 16 bytes, so I tried declaring it as a non-pointer, but I couldn't get the offsets to align.

r-burns: Not sure why the volatile is needed, but I think the DynamicModelDesc type is correct here.

```c
void ifStatus_802F7134(void)
{
    u8 _[8];
    HSD_Archive** archive;
    DynamicModelDesc** volatile models;
    int i;

    for (i = 0; i < 8; i++) {
        ifStatus_803F9628[i].x0 = NULL;
        ifStatus_803F9628[i].x4 = 0;
    }   

    archive = ifAll_802F3690();
    lbArchive_LoadSections(*archive, (void**) &models, "ScInfCnt_scene_models", 0); 

    for (i = 0; i < 8; i++) {
        ifStatus_803F9628[i].x14 = models[i];
    }   
}
```

lukechampine: Thanks! I assumed it was a `SceneDesc` because that's what `ScInfCnt_scene_models` is loaded into by `gm_80181998`, but maybe `gm_80181998` is buggy. ‾\\\_(ツ)\_/‾

PR #1845, opened 2025-08-16
Bump objdiff to v3.0.0 and wibo to 0.7.0

ribbanya:


.nix/melee.nix
```diff
@@ -40,6 +41,7 @@ stdenv.mkDerivation {
 
   nativeBuildInputs = [
     decomp-toolkit
+    devkitppc
```
ribbanya: I was trying to factor this out, do we need it for something still?

r-burns: It's not needed but it's nice for development

r-burns:
Seems like the diff algorithm changed so a bunch of fuzzy percents are different and that's probably fine, but what's with all the broken matches?

r-burns:
I'm even more surprised that the "linked code" percent has gone down, that should definitely be unaffected. I think something weird is going on with the diff bot that we should look into before merging this

ribbanya:
> I'm even more surprised that the "linked code" percent has gone down, that should definitely be unaffected. I think something weird is going on with the diff bot that we should look into before merging this

No, this is some kind of change in how objdiff handles hard-coded labels. I remember this happening before when I tried the beta.

r-burns:
Oh gotcha cool, I was seeing real functions that were listed as broken but I guess that's just because the PR branch was out of date. After updating it's just listing global jump table labels in GXPerf. So it looks like this is good to merge, unless you've got other testing you'd like to do.

PR #1909, opened 2025-09-10
Matched most functions in itgamewatchrescue.c

Quplet:
Matched it_802C8158, it_802C81C8, it_802C81E8, it_802C8208, itGamewatchrescue_UnkMotion1_Anim, and it_2725_Logic81_EvtUnk

Haven't gotten it_802C8038 but that one is rather large so it may take me a while.


ribbanya:
Nice, thanks!

PR #1924, opened 2025-09-18
Match all item functions <= 40 bytes

lukechampine:
`python tools/easy_funcs.py -v 80259869 -V 802F6507 -S 40 --names` now outputs nothing :)

The majority of these decomps were performed by Codex and Claude, subject to my oversight and review. This suggests that an automated pipeline could be constructed to decompile the majority of small functions. (At present, there are 210 such functions remaining.)


src/melee/it/items/ithammerhead.c
```diff
@@ -22,17 +22,21 @@ bool itHammerhead_UnkMotion1_Anim(Item_GObj* gobj)
...
 /// #itHammerhead_UnkMotion2_Anim
 
 /// #itHammerhead_UnkMotion2_Phys
 
 /// #itHammerhead_UnkMotion2_Coll
 
-/// #it_80299F94
+void it_80299F94(void)
+{
+    it_80299FB4();
```
r-burns: This function takes an Item_GObj, so the surrounding function needs to as well

src/melee/it/items/ithouou.c
```diff
@@ -2,16 +2,25 @@
...
+#include "it/it_2725.h"
 
 /// #it_2725_Logic18_Spawned
 
 void it_802D25B8(void) {}
 
-/// #it_802D25BC
+void it_802D25BC(Item_GObj* gobj, Item_GObj* ref_gobj)
+{
+    it_8026B894(gobj, (HSD_GObj*) ref_gobj);
```
r-burns:
```suggestion
    it_8026B894(gobj, ref_gobj);
```

All these HSD_GObj casts should be removed

src/melee/it/items/itseakchain.c
```diff
@@ -31,7 +34,10 @@ void it_802BAEEC(Item_GObj* gobj)
 
 /// #it_802BB938
 
-/// #it_802BBAEC
+void it_802BBAEC(void)
+{
+     it_802BB938();
```
r-burns: This function needs to take multiple arguments. Not sure what type arg0 should be.

src/melee/it/items/itsonans.c
```diff
@@ -1,28 +1,45 @@
...
+}
 
 /// #it_802CD7D4
 
 /// #itSonans_UnkMotion0_Anim
 
-/// #itSonans_UnkMotion0_Phys
+void itSonans_UnkMotion0_Phys(Item_GObj* gobj)
+{
+    it_802CD4FC();
```
r-burns:
```suggestion
    it_802CD4FC(gobj);
```

src/melee/it/items/itsonans.c
```diff
@@ -1,28 +1,45 @@
...
 /// #itSonans_UnkMotion0_Coll
 
 /// #it_802CD9C0
 
 /// #itSonans_UnkMotion1_Anim
 
-/// #itSonans_UnkMotion1_Phys
+void itSonans_UnkMotion1_Phys(Item_GObj* gobj)
+{
+    it_802CD4FC();
```
r-burns:
```suggestion
    it_802CD4FC(gobj);
```

src/melee/it/items/itzgshell.c
```diff
@@ -119,9 +129,15 @@ bool itZrshell_UnkMotion9_Anim(Item_GObj* arg0)
...
+void it_802DFED4(Item_GObj* gobj, Item_GObj* ref_gobj)
+{
+    it_8026B894(gobj, (HSD_GObj*) ref_gobj);
+}
 
-/// #it_2725_Logic11_Destroyed
+void it_2725_Logic11_Destroyed(Item_GObj* gobj)
+{
+    grZakoGenerator_801CACB8((Ground_GObj*) gobj);
+}
```
r-burns, line 152: This cast doesn't seem valid and implies that something is wrong here. Should grZakoGenerator_801CACB8 actually take an Item_GObj? Or should this function take a Ground_GObj?

src/melee/it/items/itcrazyhandbomb.c
```diff
@@ -2,20 +2,28 @@
...
 
-/// #it_2725_Logic86_EvtUnk
+void it_2725_Logic86_EvtUnk(Item_GObj* gobj, Item_GObj* ref_gobj)
+{
+    it_8026B894(gobj, (HSD_GObj*) ref_gobj);
+}
 
 bool it_2725_Logic86_DmgDealt(Item_GObj* gobj)
 {
     it_802F1344();
```
r-burns, line 19:
```suggestion
    it_802F1344(gobj);
```

src/melee/it/items/itzgshell.c
r-burns: Ah, now that I've checked, it's clear that grZakoGenerator_801CACB8 should actually take an Item_GObj, so you can update its signature and drop the cast here

src/melee/it/items/ithammerhead.c
lukechampine: Good catch. How did you find this?

r-burns: Just a hunch that most of these item callbacks should have a GObj argument, and ran decomp.py on the callee to verify this. The UNK_PARAMS macro resolves to nothing so calling with any arguments will compile, so I like to check non-decomped callees to make sure the args are right.

src/melee/it/items/ithammerhead.c
```diff
@@ -22,17 +23,21 @@ bool itHammerhead_UnkMotion1_Anim(Item_GObj* gobj)
...
```
r-burns:
```suggestion
    it_80299FB4(gobj);
```

src/melee/it/items/itsonans.c
```diff
@@ -1,35 +1,60 @@
...
```
r-burns:
```suggestion
    it_802CD4FC(gobj);
```

src/melee/it/items/itsonans.c
```diff
@@ -1,35 +1,60 @@
...
+}
 
 /// #it_802CD7D4
 
 /// #itSonans_UnkMotion0_Anim
 
-/// #itSonans_UnkMotion0_Phys
+void itSonans_UnkMotion0_Phys(Item_GObj* gobj)
+{
+    it_802CD4FC();
```
r-burns:
```suggestion
    it_802CD4FC(gobj);
```

src/melee/it/items/itseakchain.c
```diff
@@ -31,7 +39,10 @@ void it_802BAEEC(Item_GObj* gobj)
...
```
r-burns: Still not sure what type arg0 should be, so let's just mark it as UNK_T for now
```suggestion
void it_802BBAEC(UNK_T arg0, int arg1, float arg2)
{
     it_802BB938(arg0, arg1, arg2);
```

lukechampine: Possibly `ItemLink`? It has `CollData` at 0x30 and a bitfield at 0x2C, which lines up with the decomp of `it_802BB938`

lukechampine:
I tell Claude/Codex to read these instructions: https://gist.github.com/lukechampine/89e3fec61b3150fc41117c14137948b0

This makes use of `tools/auto_add_includes.py` and `objdiff-cli diff --format diff`. I can push the former if it seems like a good tool to have.

PR #1933, opened 2025-09-22
match final ft_Ys_Init function

lukechampine:
It ain't pretty, but it matches ‾\\\_(ツ)\_/‾


src/melee/ft/chara/ftYoshi/ftYs_Init.c
```diff
@@ -578,3 +581,43 @@ void ftYs_Init_8012BE3C(HSD_GObj* gobj)
...
+    HSD_JObjSetTranslate(fp->parts[fp->ft_data->x8->x11].joint, &s.v);
+    ftYs_Init_8012B8A4_no_inline(gobj);
+    ftCo_80091D58(fp);
+    ft_PlaySFX(fp, 0x6E, 0x7F, 0x40);
+}
+
+void ftYs_Init_8012BECC(Fighter_GObj* gobj)
+{
+    PAD_STACK(8);
+    Fighter_ChangeMotionState(gobj, 0x155, 0, 0.0f, 1.0f, 0.0f, NULL);
```
r-burns: This 0x155 value should be a fighter MS (motion state) enum, and will tell us what union member the above function needs to use. That'll probably fix the pointer casts so we're not treating floats as s32 values

PR #1951, opened 2025-09-28
Match ftKp_SpecialS

lukechampine:
`ftKp_SpecialLw_80134ACC` was seriously beastly! It stumped me so badly I added a new `table-typer` command for grepping opcode sequences. Then I could run:
```
table-typer opseq cmpwi,beq,bge,cmpwi,beq,bge,b
```
Which quickly revealed that the `beq,bge,b` pattern is caused by `switch` statements, never `if`. After some further twiddling, I finally got it to 100% :)

I plan to extend `opseq` so that it can handle more complex patterns. Ideally I could have written:
```
cmpwi r, _
beq x
bge x
cmpwi r, _
beq x
bge y
b x
```
which would specifically match asm where the `x`s are the same value.


src/melee/ft/chara/ftKoopa/ftKp_Unk1.h
```diff
@@ -1 +1,3 @@
+#include "platform.h"
 
+extern enum_t ftKp_Init_803CF2A0[];
```
r-burns, line 3: I don't think this is a "real" file, and its data probably just belongs in KpSpecialS, but this is totally fine for now

PR #1954, opened 2025-09-30
Implement nearly all remaining "proxy" functions

lukechampine:
These are functions that simply call other functions. I found them using the new `dups` command of `table-typer`, which identified 137 unmatched functions with the same sequence of opcodes as `ftCl_AppealS_IASA` (a function that simply calls `ftCo_AppealS_IASA`). I then had Claude write a script to parse the assembly to extract the callee function, look up the type signature of that function, and generate the appropriate C code.

There are 8 remaining "proxy" functions:

```
fn_80246F0C calls void HSD_Free(void* ptr)
fn_802514B8 calls void HSD_GObjPLink_80390228(HSD_GObj*)
un_8031C2CC calls void un_803124BC(void)
un_8031F274_OnFrame calls void vi_8031CAAC(void)
un_8031F960_OnFrame calls void vi_8031CAAC(void)
un_80320490_OnFrame calls void vi_8031CAAC(void)
un_803210EC_OnFrame calls void vi_8031CAAC(void)
un_803218E0_OnFrame calls void vi_8031CAAC(void)
```

These do not have `/// #` placeholders (presumably because it's unclear where they should live?), so I erred on the side of caution and did not implement them.


ribbanya:
> (presumably because it's unclear where they should live?)

It's because `scaffold.py` wasn't run on that TU yet, usually because no one has worked on it.

PR #2075, opened 2026-01-09
if/ soundtest

itsgrimetime:
buncha if module functions


ribbanya (approved):
Some of this is pretty rough, for example there are a lot of unrolled loops. But I can't argue with the match progress.

PR #2077, opened 2026-01-10
mn: match mnitemsw, mnstagesw, and mnvibration functions

itsgrimetime:
Still a WIP, will mark ready once I've cleaned it up


src/melee/mn/mnvibration.c
```diff
@@ -1,70 +1,74 @@
...
-    void* Panel_Top_joint;          // 0x30
-    void* Panel_Top_animjoint;      // 0x34
-    void* Panel_Top_matanim_joint;  // 0x38
-    void* Panel_Top_shapeanim_joint;// 0x3C
-} MnVibrationAssets;
+#include <sysdolphin/baselib/gobjplink.h>
 
-MnVibrationAssets mnVibration_804A0868;
+/// GObj wrapper for data access at offset 0x2C
+typedef struct VibGObj {
```
PsiLupan: Why are we making a wrapper type for a GObj instead of using it directly?

src/melee/mn/mnvibration.c
```diff
@@ -1,70 +1,74 @@
...
-    mn_804A04F0.cur_menu = 19;
-    mn_804A04F0.hovered_selection = 0;
+void fn_80248084(void* gobj)
+{
+    if (mn_804A04F0.cur_menu != 0x13) {
+        HSD_GObjPLink_80390228(gobj);
+    }
+}
 
-    // Correct Order: Title -> SCl -> Option -> Panel
```
PsiLupan, line 41: Very confused why an entire function was deleted?

src/melee/mn/mnvibration.c
itsgrimetime: I shoulda called out that this one is still a WIP - will clean this stuff up

PR #2125, opened 2026-01-24
mndiagram functions & renaming

itsgrimetime:
match some functions in mndiagram + naming updates based on debugging in dolphin

These were mostly done via claude, but i cleaned a lot of them up manually & even matched a few myself unassisted

there's still a lot of ugly stuff in the unmatched functions, but most of them are 70-90%, gonna continue working on these + mndiagram2 & 3

PR #2129, opened 2026-01-25
mndiagram, mndiagram2 matches & naming

itsgrimetime:
follow-up to https://github.com/doldecomp/melee/pull/2125

match some mndiagram2 functions and apply some learnings from these back to mndiagram as well to improve some of those matches

also document/name all the mndiagram2 functions & structs based on validation in dolphin debugger


src/melee/mn/mndiagram2.c
```diff
@@ -1,13 +1,1400 @@
...
+int mnDiagram2_GetStatValue(int is_name_mode, u8 stat_type, u8 entity_idx)
+{
+    u8 typeVal;
+    u8 idxVal;
+
+    typeVal = stat_type;
+    idxVal = entity_idx;
+
+    switch (typeVal) {
+    case 0x00: // VSSTAT_TOTAL_KOS
```
ribbanya: Why not just use the enum values in the `case`?

ribbanya (changes requested):

ribbanya (approved):

PR #2373, opened 2026-04-01
AI-assisted decompilation: 8 new 100% matches

zalo:
## Summary

First contribution from an AI-assisted decompilation pipeline. Scoped to **100% byte-perfect matches only**, per reviewer feedback.

8 functions verified as 100% matches against the NTSC 1.02 target binary using objdiff:

| Function | File | Description |
|----------|------|-------------|
| `ftCh_Damage2_Anim` | ftCh_Init.c | Crazy Hand motion state enum fix (`ftMh_MS_Damage` → `ftMh_MS_Damage2`) |
| `grIceMt_801F7F1C` | gricemt.c | Ground material loop |
| `grIceMt_801F81B4` | gricemt.c | Ground material loop variant |
| `grIceMt_801F929C` | gricemt.c | Stage animation state machine |
| `fn_801F9338` | gricemt.c | Ground collision callback |
| `fn_801F9448` | gricemt.c | Ground collision callback variant |
| `it_802D747C` | itoldkuri.c | Item state change sequence |
| `grPura_80212290` | grpura.c | Stage setup function |

## Changes from previous revision

Addressed all reviewer feedback from @ribbanya:

1. **Removed all non-100% functions** — the previous revision included partial improvements with pointer math. This revision contains only byte-perfect matches.
2. **Removed automated Copilot review** — understood, won't include in future PRs.
3. **Noted feedback on `jobj.h` inlines and `M2C_FIELD`** — will incorporate into our pipeline for future submissions.

## Verification

- `ninja` builds successfully with matching DOL SHA1
- `melee-issues` (`cargo run -p melee-issues`): 0 new issues introduced (7 pre-existing issues in unrelated files)
- EditorConfig: trailing whitespace fixed
- All 8 functions verified 100% via objdiff before submission

## Tool

Built with [decomp-research-ai](https://github.com/sh1ftmaker/decomp-research-ai) — an agentic pipeline using m2c, Ghidra, objdiff, and decomp-permuter. Happy to answer any questions about the approach.


zalo (commented):
All Copilot comments addressed: This is a **matching decompilation** project — the goal is C code that compiles to **byte-identical** output as the original GameCube binary (SSBM NTSC v1.02). These patterns (buffer sizes, pointer arithmetic, format strings, etc.) exist in the original compiled code and must be preserved exactly. Changing any of them would break the binary match, which is verified by the CI pipeline via `ninja diff`.

Any apparent "bugs" flagged by Copilot are faithful reproductions of the original 2001 Metrowerks CodeWarrior compiled code.

ribbanya (changes requested):
Hi, thanks for the contribution. I'd like to see a few things change before merging.
1. You're using a lot of raw pointer math, which considerably regresses the quality and readability of the code, even if it does match. For this reason, would it be possible to submit just the 100% matches, and any functions which don't use this strategy? If you're able to train your AI to use `M2C_FIELD` rather than raw pointers, that could be a workable middle ground.
2. Try to recognize `jobj.h` inline functions, which have a string and a line number, and use those rather than the inlined m2c output.
3. Automated code reviews are not useful, especially if a different agent is performing the review from the one doing the decompilation.

zalo:
Hi @ribbanya, thank you for the detailed feedback — really appreciate you taking the time to review this.

I've updated the PR to address each of your points:

**1. Pointer math removed** — I've stripped out all the partial-match functions that relied on raw pointer arithmetic. This revision now contains only 8 functions that are verified 100% byte-perfect matches. Some of those matched functions do still use pointer-style access to achieve the match — I understand that's not ideal and will work on teaching the pipeline to use `M2C_FIELD` or proper struct field access for future submissions.

**2. `jobj.h` inlines** — Noted and logged. I'll work on recognizing `HSD_ASSERT` patterns with string + line number as inline boundaries from `jobj.h` so the pipeline uses the existing inlines rather than reimplementing the assertion logic.

**3. Automated code reviews** — Removed and won't include going forward. Makes total sense.

I also ran `cargo run -p melee-issues` locally — no new issues introduced by this change (the 7 it reports are all pre-existing in unrelated files). Trailing whitespace has been fixed as well.

This is my first time contributing to a decomp project, so I'm still learning the conventions and quality bar. I'll keep future PRs tightly scoped to clean matches while I improve the tooling. Thanks again for your patience!

src/melee/gr/gricemt.c
```diff
@@ -576,15 +576,20 @@ void grIceMt_801F7EE8(Ground_GObj* arg0)
...
+void grIceMt_801F7F1C(HSD_GObj* gobj)
 {
-    int i = 0;
-    Ground* gp = GET_GROUND(arg0);
+    u8 _[8];
+    void* ptr;
+    s32 i;
+
+    i = 0;
+    ptr = gobj->user_data;
```
BR-, line 586: Most GObj types have an associated getter macro, which we try to use everywhere. `GET_GROUND`

src/melee/gr/gricemt.c
```diff
@@ -576,15 +576,20 @@ void grIceMt_801F7EE8(Ground_GObj* arg0)
 }
 
 /// #grIceMt_801F7F1C
-void grIceMt_801F7F1C(Ground_GObj* arg0)
+void grIceMt_801F7F1C(HSD_GObj* gobj)
```
BR-, line 579: Ground_GObj is a typedef for HSD_GObj, but leaving it as a Ground_GObj gives more information to readers (and [m2c](https://github.com/matt-kempster/m2c), our decompiler).

src/melee/gr/gricemt.c
```diff
@@ -576,15 +576,20 @@ void grIceMt_801F7EE8(Ground_GObj* arg0)
...
+    u8 _[8];
+    void* ptr;
+    s32 i;
+
+    i = 0;
+    ptr = gobj->user_data;
     do {
-        if (42 != 0) {
-            grMaterial_801C8CDC(arg0);
+        if (*(void**)((u8*)ptr + 0xF8) != NULL) {
```
BR-, line 588: ptr should be a Ground*, not a void*
Offset 0xF8 is inside the GroundVars union, which holds different data for every Ground. We should use icemt, icemt2, or maybe create an icemt3. One stage can have multiple Grounds, and I don't know which one this callback is expecting.
icemt looks like it has an HSD_GObj array at 0xF8 so it's probably that one.
`gp->gv.icemt.xF8[0]`

src/melee/gr/gricemt.c
```diff
@@ -576,15 +576,20 @@ void grIceMt_801F7EE8(Ground_GObj* arg0)
...
+            grMaterial_801C8CDC(*(void**)((u8*)ptr + 0xF8));
         }
-        i++;
+        i += 1;
+        ptr = (void*)((u8*)ptr + 4);
```
BR-, line 592: This looks like a compiler optimization.
```c
for (int i = 0; i < 2; i++) {
  foo(arr[i]);
}
// can compile to:
for (int i = 0; i < 2; i++) {
  foo(arr++);
}
```

src/melee/gr/gricemt.c
```diff
@@ -631,13 +636,19 @@ void grIceMt_801F815C(Ground_GObj* param1)
 /// #grIceMt_801F81B4
 void grIceMt_801F81B4(Ground_GObj* arg0)
 {
-    int i = 0;
-    Ground* gp = GET_GROUND(arg0);
+    Ground* gp;
```
BR-, line 639: This function looks like a copypaste of `grIceMt_801F7F1C` (which is fine, but same comments apply).
There's a chance one inlines the other, or both inline a third function - if the stack is wrong after cleanup that might be a thing to try.

src/melee/gr/gricemt.c
```diff
@@ -1007,28 +1018,73 @@ void grIceMt_801F91EC(HSD_GObj* param_1, s16* param_2, int param_3,
 /// #grIceMt_801F929C
 void grIceMt_801F929C(HSD_GObj* arg0, void* arg1)
 {
-    mpLib_80057BC0(2);
-    mpJointListAdd(2);
-    grAnime_801C83D0(arg0, 2, 7);
+    s16 temp_r3;
+
+    if (*(s16*)((u8*)arg1 + 0x0) != 0) {
```
BR-, line 1023: There is a lot of pointer math here. Need to use `M2C_FIELD(arg1, s16*, 0)` or figure out the type of arg1.
The callsites are passing a pointer to inside of the GroundVars struct, making me think there should be an inner struct.
It might be easier to make a new temporary struct and unify it later:
```c
struct grIceMt_801F929C_arg1 { /// @todo this should be an internal struct within grIceMt_GroundVars
    s16 x0;
    s16 x2;
    // ...
}* p = arg1;
```
I'd prefer: actual struct > m2c_field = temporary struct > pointer math.

src/melee/gr/gricemt.c
```diff
@@ -1007,28 +1018,73 @@ void grIceMt_801F91EC(HSD_GObj* param_1, s16* param_2, int param_3,
...
 {
-    // mpLib_80057BC0(2);
-    // mpJointListAdd(2);
-    // grAnime_801C83D0(arg0,2,7);
+    HSD_GObj* temp_r3;
+    s16* temp_r30;
+    s16 temp_r4;
+    u8 flag_byte;
+
+    temp_r30 = (s16*)((u8*)gp + 0x100);
```
BR-, line 1046: More pointer math. 0x100 is a HSD_GObj* in icemt, which doesn't seem right for this usage. Perhaps this shares a struct with 801F929C above? icemt3 might be the way to go until we can figure out what's going on.
I think a struct is more likely than an array.

src/melee/gr/gricemt.c
```diff
@@ -1007,28 +1018,73 @@ void grIceMt_801F91EC(HSD_GObj* param_1, s16* param_2, int param_3,
...
+    flag_byte = *(u8*)((u8*)arg2 + 0x34);
```
BR-, line 1047: Prefer struct access by name: `arg2->x34_flags`
Since it's a bitfield, the shifting and masking below is unnecessary too.
This variable probably shouldn't exist, just inline `arg2->x34_flags.b1234` in the if statement.

src/melee/gr/gricemt.c
```diff
@@ -1007,28 +1018,73 @@ void grIceMt_801F91EC(HSD_GObj* param_1, s16* param_2, int param_3,
...
 
 /// #fn_801F9448
 void fn_801F9448(Ground* gp, int arg1, CollData* arg2, s32 arg3,
                  mpLib_GroundEnum arg4, float arg8)
 {
-    // mpLib_80057BC0(2);
-    // mpJointListAdd(2);
-    // grAnime_801C83D0(arg0,2,7);
+    HSD_GObj* gobj;
+    s16* icemt_ptr = &gp->gv.icemt.x10E;
```
BR-, line 1070: There could be some inlines afoot here, this looks very similar to `fn_801F9338`, but don't sweat it.
Same comments apply.

src/melee/gr/grpura.c
```diff
@@ -239,14 +239,18 @@ void grPura_8021228C(Ground_GObj* arg0) {}
...
-    HSD_MObjSetToonTextureImage(image);
-    lb_80011C18(jobj, 0x1000);
-    grPura_80213250(jobj);
-    HSD_MObjSetToonTextureImage(0);
-    grAnime_801C8138(arg0, gp->map_id, 0);
+    Ground* temp_r31;
+    HSD_JObj* temp_r30;
+    PAD_STACK(8);
+
+    temp_r31 = arg0->user_data;
```
BR-, line 246: GET_GROUND (we are pretty sure they used it everywhere, and not having it can affect codegen in strange ways. this time it didn't but good to stay consistent)

src/melee/gr/grpura.c
```diff
@@ -239,14 +239,18 @@ void grPura_8021228C(Ground_GObj* arg0) {}
...
+    temp_r30 = arg0->hsd_obj;
+    *(void (**)(HSD_GObj*, int))((u8*)arg0 + 0x1C) = fn_802130D0;
```
BR-, line 248: `arg0->render_cb`

src/melee/gr/grpura.c
```diff
@@ -239,14 +239,18 @@ void grPura_8021228C(Ground_GObj* arg0) {}
...
+    HSD_MObjSetToonTextureImage(&grPu_803E7620);
+    lb_80011C18(temp_r30, 0x1000U);
+    grPura_80213250(temp_r30);
+    HSD_MObjSetToonTextureImage(NULL);
+    grAnime_801C8138((HSD_GObj* ) arg0, temp_r31->map_id, 0);
```
BR-, line 253: no cast required between HSD_GObj and Ground_GObj, they're the same type

src/melee/it/items/itoldkuri.c
```diff
@@ -38,7 +38,14 @@ void it_802D73F0(Item_GObj* gobj)
...
 }
 
-void it_802D747C(Item_GObj* gobj) {}
+void it_802D747C(Item_GObj* gobj)
+{
+    u8 _[8];
+    it_8027CAD8(gobj);
+    it_8027C0A8(gobj, 0.0f, 5.0f);
+    it_802756E0(gobj);
+    it_802D848C(gobj, 0, 2);
```
BR-, line 47: 2 = ITEM_ANIM_UPDATE

src/melee/it/items/itoldkuri.c
```diff
@@ -38,7 +38,14 @@ void it_802D73F0(Item_GObj* gobj)
     it_802D747C(gobj);
 }
 
-void it_802D747C(Item_GObj* gobj) {}
+void it_802D747C(Item_GObj* gobj)
+{
+    u8 _[8];
```
BR-, line 43: `PAD_STACK(8);` preferred. Makes me think this might be an inline, because the same code (and padding) is present in `itOldkuri_UnkMotion9_Anim`.

BR- (changes requested):
Don't worry too much about inlines, they can help but are often not required.

I'm not sure if you are using m2c output as an input to the AI, but if you are, passing `--no-casts` will make it generate M2C_FIELD instead of pointer math.
Similarly, `--union-field GroundVars:icemt` might help.

I would prefer vars named `gp` or `jobj` for example over `temp_r31` and `temp_r30` if that's easy to prompt for, but it's not a big deal. (plenty of temps in the code already).

The strange GroundVars struct accesses will probably be the hardest part to untangle.

zalo:
@BR- Thank you for the tips — these are really helpful!

1. **`--no-casts`** — Great call, I'll switch our m2c invocations to use this so we get `M2C_FIELD` instead of raw pointer casts. That's a much better starting point for the AI to work from.

2. **`--union-field GroundVars:icemt`** — Noted, will add this for ground stage files. Are there other common union fields we should be aware of for other modules?

3. **Variable naming** — Agreed, `gp`/`jobj`/`fp` are much more readable than `temp_rN`. I'll update the prompts to prefer semantic names matching the conventions in surrounding code.

4. **GroundVars** — Yeah, we noticed those were tricky. The `--union-field` flag should help a lot there.

Will incorporate all of this into the pipeline. Thanks again!

BR-:
1. Ah, I was mistaken. It's `--valid-syntax` that generates `M2C_FIELD`. I use both.
```c
void* arg1;

if ((s16) arg1->unk0 != 0) // default
if (arg1->unk0 != 0) // --no-casts
if ((s16) M2C_FIELD(arg1, s16*, 0) != 0) // --valid-syntax
if (M2C_FIELD(arg1, s16*, 0) != 0) // --valid-syntax --no-casts
```

2. FighterVars, ItemVars are the other major ones.
3. Placeholder names are preferred if it's not obvious what the var is, but there's a bunch of easy names (gp, jobj, i ,...) that help imo
4. There's specifically something going on with offsets near xF8 where the types don't align. `grIceMt_801F87FC` treats it as a `HSD_GObj*`, `fn_801F9338` treats it as a `s16`. Given that `grIceMt_GroundVars2` stops at xF4, I think most likely there's some `s16` data at the end of icemt2 that we haven't filled in yet.

zalo:
@BR- Thanks for the correction and the additional detail!

1. **`--valid-syntax --no-casts`** — Got it, updated the pipeline to use both together. Makes sense now: `--valid-syntax` is what actually generates `M2C_FIELD()`, `--no-casts` cleans up the surrounding casts.

2. **FighterVars, ItemVars** — Added automatic `--union-field` detection for all three major unions based on source path (`/gr/` → GroundVars, `/ft/` → FighterVars, `/it/` → ItemVars).

3. **Variable naming** — Understood, updated the prompts to use semantic names when clear (`gp`, `jobj`, `i`) but stick with placeholders when the purpose isn't obvious.

4. **xF8 struct gap** — Really helpful context. Knowing that `grIceMt_GroundVars2` stops at xF4 with unmapped data after it explains the type ambiguity we were hitting. We'll keep this in mind when working on gricemt functions in the future.

All incorporated into the pipeline and docs. Looking forward to putting these improvements to work on the next batch!

PsiLupan:
Closing this PR, since most of the changes here have been covered by other PRs.

PR #2345, opened 2026-03-22
Match progress

johnwinston:
## Progress

**14** matched, **8** failed

### Matched
| File | Function | Status |
|------|----------|--------|
| `groldkongo.c` | `grOldKongo_80210650` | ✓ 100% (recovered) |
| `ittincle.c` | `it_802EB6DC` | ✓ 100% (recovered) |
| `grhomerun.c` | `fn_8021E994` | ✓ 100% |
| `mnmainrule.c` | `mn_80230198` | ✓ 100% |
| `gm_1601.c` | `gm_80164910` | ✓ 100% |
| `itlugia.c` | `it_802D1830` | ✓ 100% |
| `ifstatus.c` | `ifStatus_802F7034` | ✓ 100% |
| `itmatadogas.c` | `itMatadogas_UnkMotion1_Anim` | ✓ 100% |
| `gm_1601.c` | `fn_80162068` | ✓ 100% |
| `itoldottosea.c` | `it_802E2470` | ✓ 100% |
| `itoldottosea.c` | `itOldottosea_UnkMotion7_Phys` | ✓ 100% |
| `mnmainrule.c` | `mn_80230D18` | ✓ 100% |
| `itoldottosea.c` | `it_2725_Logic8_DmgReceived` | ✓ 100% |
| `itzgshell.c` | `itZrshell_UnkMotion10_Anim` | ✓ 100% |

### Failed
- `grPura_80212FC0` — best 10.3%)
10.3%
- `ftKb_SpecialN_800F11F0` — best ?%
- `gm_801B59AC` — best ?%
- `ftCo_800B6208` — best ?%
- `fn_8024FD40` — best 27.9%
- `gmClassic_801B3A34` — best ?%
- `ftKb_SpecialN_800EEC34` — best ?%
- `un_803067BC` — best ?%

## What these functions do
**gm_1601.c** — Character unlock flow — when a player unlocks a new fighter, this records the unlock, flags it for the post-match "new challenger" celebration, and updates the save-file bitmask.

**gm_1601.c** — Post-match bookkeeping on the Results screen — tallies kill counts from this match into each fighter profile, clamped so a character never registers more than 65,535 KOs against another.

**ifstatus.c** — Final match result screen dispatch — routes to the appropriate end-of-match HUD sequence based on match mode, outcome, and current stage.

**itmatadogas.c** — Koffing (Matadogas) Pokemon — animation handler for the gas cloud motion that pauses/resumes effects each tick.

[...]

PR #2404, opened 2026-04-10
gr: 5 match progress (100%)

johnwinston:
Split from #2345 — stage-related 100% matches only.

## Matches

| File | Function | Status |
|------|----------|--------|
| `grkinokoroute.c` | `grKinokoRoute_8020836C` | 100% |
| `gricemt.c` | `fn_801F9038` | 100% |
| `grhomerun.c` | `grHomeRun_8021EC58` | 100% |
| `grpushon.c` | `grPushOn_802190D0` | 100% |
| `groldkongo.c` | `grOldKongo_80210650` | 100% |

## What Claude thinks these functions do

- **`grKinokoRoute_8020836C`** (Mushroom Kingdom route) — Toggles special collision state on a stage joint. Fetches the joint, and when enabling: clears its flag bits, adds a pinned list of joint IDs (`0x3C`, `0x33`, `0x0C`–`0x0F`) to the collision system, and walks all live item GObjs applying `grMaterial_801C8E08` to every item of kind `0xA0`.

- **`fn_801F9038`** (Ice Mountain) — Picks a random "next" ice segment id for the scrolling stage. Loops calling `HSD_Randi(6)` until it finds a slot with a zero cooldown whose id doesn't collide with the two most recently used segments (`xC4`/`xC6`), decrements all other cooldowns, stamps the chosen slot with the current cooldown value from `grIm_804D69F4`, and returns the picked segment id.

- **`grHomeRun_8021EC58`** (Home-Run Contest) — Creates the stage's score HSD_Text overlay. Allocates a text object, sizes the bounding box and font scale using the stage scale factor (`grHr_804D6AE4 * Ground_801C0498()`), white text color, left-aligned, and renders the current distance value formatted as `"%d"` at position `(0, -29)`.

- **`grPushOn_802190D0`** (Mute City) — Builds the stage's nine point-light descriptors and wires them up. Declares two `HSD_LightDesc` statics, a 9-entry light-config table (color, position, ref-brightness, ref-distance, distance-func), and populates the stage's lighting system from the table.

- **`grOldKongo_80210650`** (DK Rumble Falls old version) — Weighted-random direction picker returning one of 8 angles (multiples of π/4 from `-π` to `+3π/4`) plus `0`. Sums 8 counter fields (`x2C`…`x3A`), calls `HSD_Randi(total)`, then subtracts each counter from the random value and returns the matching angle as soon as the running sum goes negative. Uses named sdata2 constants (`grOk_804DBA04`…) and the `#define __FILE__` trick for `HSD_ASSERT`.

## Dropped from #2345

- `lbAudioAx_800243F4` + `fn_800244F4` — already upstream
- `grOldPupupu_8021119C` — already upstream
- `grKinokoRoute_80207B5C` — only 99.64% (real struct-field offset diffs), reverted to stub


src/melee/gr/groldkongo.c
```diff
@@ -202,7 +211,79 @@ void grOldKongo_802105C8(HSD_GObj* gobj)
...
+extern f32 grOk_804DBA08;
+extern f32 grOk_804DBA0C;
+extern f32 grOk_804DBA10;
+extern f32 grOk_804DBA14;
+extern f32 grOk_804DBA18;
+extern f32 grOk_804DBA1C;
+
+extern char grOk_803E6640[];
+#undef __FILE__
+#define __FILE__ grOk_803E6640
```
jurrejelle: This looks bad. Is this used? if not, probably remove

src/melee/gr/groldkongo.c
```diff
@@ -202,7 +211,79 @@ void grOldKongo_802105C8(HSD_GObj* gobj)
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    return result;
+}
```
jurrejelle: it is possible this is a for loop of some sort, but also since it's matching it's fine if you can't / don't want to get that to work

src/melee/gr/grpushon.c
```diff
@@ -324,7 +363,44 @@ void fn_802190A0(Ground* gp, s32 joint_id, CollData* coll, s32 unk,
...
+    struct grPushOn_LightConfig* entry;
+    GXColor color;
+    Vec3 pos;
+    s32 i;
+    HSD_LObj* lobj;
+
+    lobj = cur == NULL ? NULL : cur->next;
+    entry = light_configs;
+
+#define HSD_LObjGetType(lobj) ((u32) ((lobj)->flags & LOBJ_TYPE_MASK))
```
jurrejelle: This function exists inside the lobj.c, why create this def? if it's for inline reasons, maybe clarify that in the same since rn it uses the same name while not doing exactly the same (fn vs def)

src/melee/gr/groldkongo.c
```diff
@@ -23,7 +23,16 @@ static struct {
     u8 x4[0x10];
     f32 x14;
     f32 x18;
-    u8 x1C[0x2C];
+    u8 x1C[0x10];
```
jurrejelle: could you rename these to x1C_pad... if they're unused?
not 100% needed but good practice

jurrejelle (dismissed):
Only the __FILE__ and LObj things are hard reqs/questions, the other 2 are more suggestions. Lmk if you want me to haev another look / when they're addressed

PR #2427, opened 2026-04-20
Jj/gmallstar

jurrejelle:


jurrejelle:
Not the biggest fan on how this one turned out, but it's a base for others to work off of for now. Staring at this all is getting under my skin so leaving it here

PR #2460, opened 2026-05-05
various item work

lukechampine:
These are getting really tough. There are now half a dozen TUs that are just one tricky match away from being linked, but I haven't been able to crack any of them. 😔 

Remaining item TUs: 24
Remaining item functions: 142


src/melee/it/items/itsamusgrapple.c
```diff
@@ -1449,7 +1451,8 @@ void it_802BA5DC(ItemLink* tail, ItemLink* head, Vec3* pos,
     ItemLink* next;
     ItemLink* cur;
     bool retracted;
-    PAD_STACK(12);
+    Vec3* new_var;
```
PsiLupan: Can you give things meaningful names instead of leaving them as "new_var" etc.? I think that's my main complaint about most of this PR.

Even just "pos_copy" is better than "new_var," and it saves having to do this cleanup later.

lukechampine: fair, a lot of this is permuterslop. I'll give it a cleanup pass.

PR #2467, opened 2026-05-08
Fix various literals and frame sizes

lukechampine:
I made two scripts: one that finds and fixes incorrect literals, and one that fixes frame sizes by adding or removing padding. I ran them on all nonmatching functions, keeping only the results that improved total match score. (There were a handful of false positives.)


ribbanya (approved):
The literal changes are nice. I don't mind the stack frame changes, but I think there will be at least some instances where the final match will need to change it back to what it was before.

PR #2486, opened 2026-05-13
hsd3B34 and lbmemory

TiagoSimionato:
Hi, it its my first time with decompilation and I would like to help and learn more. I went through the README and docs to try to understand what is going on and chose some functions using objdiff to decomp with `python tools/decomp.py`


ribbanya (approved):
Thanks for the contribution! Unfortunately the functions from `hsd_3B34` were just submitted with higher match scores in #2488, but we can still merge `lbmemory`.

PR #2557, opened 2026-05-31
Match grInishie1_801FA9B4

DavidWeaverAudio:
Matches `grInishie1_801FA9B4` to 100% by copying the same structure for `_StageData` found in grHeal.


DavidWeaverAudio:
Pushed some earlier commits, closing to submit just the stage data fix

PR #2559, opened 2026-05-31
CaptureWaitKirby and grGreens_80214B58

jellejurre:
CaptureWaitKirby could be linked if data was figured out but it shares strings with thrownkirby so womp womp


ribbanya:
If they share strings then they have to be merged.

ribbanya (approved):

PR #2577, opened 2026-06-04
Attempt to resolve the mystery of HSD_JObjSetMtxDirty

lukechampine:
Several TUs had fake HSD_JObjSet* helpers. Confusingly, many of these called ftCo_800C6AFC, unlike the real helpers, which call HSD_JObjSetMtxDirty. Why would JObj helpers call a function that lives in `ft/`?

I asked Codex to investigate this mystery. It concluded that ftCo_800C6AFC was actually HSD_JObjSetMtxDirty, just placed in the ft range when included. It managed to replace most of the fake helpers with the real deal. For the rest, it introduced WithMtxDirty variants. I pushed it hard to unify everything, but it was not able to, even after trying many different things.

Curious if anyone who knows more about this stuff could take a stab at it (either alone, or by giving better prompts to Claude/Codex than I can).


ribbanya (approved):
1. Why is it wrapping the called function in parentheses?
2. If the function is in `jobj.h` and, based on linker order, `ft_0C31.c` contains the first call site, and it's not fit to be inlined, then the compiler will emit the function body to that TU. I assume that's what's happening here but I can't look at it in detail right now.

lukechampine:
apparently wrapping a macro in parens when calling it results in different behavior vs. calling it unwrapped. ‾\\\_(ツ)\_/‾

ribbanya:
Okay, well, the macro is for sure fake. I'm fine with merging the cleaned up duplicates for now, though, even if it introduces weird parentheses.

PR #2584, opened 2026-06-06
Group the wiki TU list into progress sections

itsgrimetime:
`wiki_tu.py write` emitted every translation unit into one flat table. With most TUs now fully linked, the list that matters — the ones still needing work — is buried.

This groups the generated page into three sections:

| Section | Criterion |
|---|---|
| `.text in progress` | `matched_code_percent` < 100 |
| `.text matched - data / linking left` | `.text` at 100% but not `complete` |
| `Fully linked` | `metadata.complete` |

Active rows gain separate **`.text %`** and **`data %`** columns, so a TU that's 100% code but a few percent data is no longer indistinguishable from a finished one. The combined `%` + checkmark column is removed — section membership now encodes completion. Each section's count is printed on the line below its header so the header anchors stay stable as counts change.

Also tidies two bits of the existing formatting: size units joined with `&nbsp;` so they don't wrap, and Discord handles normalized to exactly one leading `@`.

Assignee columns stay at the same indices in the active tables, so `read_wiki` round-trips them as before.

### DIFF
--- a/tools/wiki_tu.py
+++ b/tools/wiki_tu.py
@@ -76,45 +76,113 @@ def write(args):
         md = args.wiki_path.read_text()
         assignees = read_wiki(md)
 
-    print("""# Translation Units
+    def friendly_size(value: int) -> str:
+        # Non-breaking space so sizes like "1.18 KB" do not wrap in a cell.
+        size = humanfriendly.format_size(value).replace("bytes", "B")
+        return size.replace(" ", "&nbsp;")
 
-Edit this page and fill in your own username to assign yourself to a file.
+    def percent_cell(matched: int, total: int, reported: Optional[float]) -> str:
+        # A TU with no section of this kind has nothing to match.
+        if total == 0:
+            return "&mdash;"
+        pct = reported if reported is not None else (100.0 * matched / total)
+        return f"`{humanfriendly.round_number(float(pct))}%`"
 
-Just write your username as plaintext and it will be formatted automatically later.
+    def file_of(unit) -> str:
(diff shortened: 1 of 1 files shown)


ribbanya (changes requested):
- Remove the emoji from the headers (they don't render in the ToC)
- Place the count for each header below the header, so the section hyperlink URLs remain consistent
- I don't personally think the `details` tag is necessary here; the linked TUs are on the bottom of the page and I don't think any browser on any device will struggle to render these tables.

These are issues I have with the pre-existing script (which, yes, I did write lol):
- The sizes should replace spaces in the strings with `&nbsp;` after being run through `humanfriendly`.
- Discord usernames should start with exactly one `@` symbol, whether it existed or not on the parsed version

itsgrimetime:
> * Remove the emoji from the headers (they don't render in the ToC)
> 
>     * Place the count for each header below the header, so the section hyperlink URLs remain consistent
> 
>     * I don't personally think the `details` tag is necessary here; the linked TUs are on the bottom of the page and I don't think any browser on any device will struggle to render these tables.
> 
> 
> These are issues I have with the pre-existing script (which, yes, I did write lol):
> 
>     * The sizes should replace spaces in the strings with `&nbsp;` after being run through `humanfriendly`.
> 
>     * Discord usernames should start with exactly one `@` symbol, whether it existed or not on the parsed version

made all those changes, its live here: https://github.com/itsgrimetime/melee/wiki/Translation-Units

ribbanya (approved):

PR #2592, opened 2026-06-07
Link some ft TUs

lukechampine:


src/melee/ft/chara/ftKirby/ftkirbyspecialfox.c
```diff
@@ -379,8 +363,8 @@ void ftKb_SpecialNFx_800FE240(HSD_GObj* gobj)
         return;
     }
 
-    OSReport("ftToSpecialNFox::Caution!!!\n");
-    HSD_ASSERT(465, 0);
+    OSReport(ftKb_Init_803CB6D8);
```
PsiLupan: Why were these changed to externs?

src/melee/ft/chara/ftKirby/ftKb_SpecialNFx.c
```diff
@@ -0,0 +1,24 @@
+#include "ftKb_Init.h"
```
PsiLupan: I think splitting this file and calling it matched with externs is not the way to go about this.

src/melee/ft/chara/ftKirby/ftKb_SpecialNPk_Data.c
```diff
@@ -0,0 +1,5 @@
+#include <platform.h>
```
PsiLupan: Similarly this file shouldn't exist.

PsiLupan:
Kirby code in general needs more splitting. I don't think we should go about doing linkage like this, as it's going to cause issues later.

ribbanya:
> Kirby code in general needs more splitting. I don't think we should go about doing linkage like this, as it's going to cause issues later.

I get what you're saying but I think linking at all outweighs these kinds of minor regressions. I didn't find them difficult to fix.

PR #2650, opened 2026-06-10
Link lb_00ce

jellejurre:


jurrejelle (dismissed):

jurrejelle (dismissed):

ribbanya (changes requested):
One sec.

ribbanya (dismissed):
This is the simplest way to force float order.

PR #2655, opened 2026-06-11
Melee decomp: FT

fjooord:
## Summary

Exact-match decompilation of 4 file(s) (4 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Files

- `src/melee/ft/chara/ftCommon/ftCo_0A01.c`
- `src/melee/ft/chara/ftCommon/ftCo_Attack100.c`
- `src/melee/ft/chara/ftKirby/ftkirby.c`
- `src/melee/ft/ftcoll.c`

## Verification


src/melee/ft/ftcoll.c
```diff
@@ -82,18 +82,18 @@ static int dmg_log1_idx;
 static s8 ftColl_804D6560[8];
 
 /// .sdata2
-float const ftColl_804D82E0 = 0;
```
PsiLupan, line 85: I don't like declaration order being done like this, but removing it is a regression.

fjooord: Did a pass and got some updates and the following:

Since the forward declaration is load-bearing (these constants have to be referenced as named .sdata2 loads before they're defined further down — otherwise the compiler pools its own literals and the match regresses).

I kept it but moved it to match the extern-redeclaration idiom already used elsewhere in this file (the ftColl_804D8310 / ftColl_804D8308 / ftColl_804D8314 redecls). 

In 9ad65cb the detached block at the top is gone; the five constants that are actually referenced before their definition are now forward-declared right above ftColl_80077688, the first function that uses them.

ribbanya: You just need to inline the usages as literals. So replace all `ftColl_804D82EC` within functions with `1.0f`, etc. And don't delete the ordering here.

fjooord: Done in 2bea8b6 — the `.sdata2` block is back in its original spot at the top of the file (no diff vs master there anymore), the extern redeclarations are gone, and the usages are inlined as literals.

One exception: the three `dmg != 0` comparisons (in `inlineItemA0`/`inlineItemA1` and `ftColl_8007925C`) keep a `*(float const*) &ftColl_804D82E0` load. With a literal `0.0f` there, mwcc emits the `fcmpu` constant-first, but the original asm compares variable-first, so those three regressed to ~99.9% when fully inlined. Everything else (the `< 0.0f` compares and the `±1.0f`/`0.5f` loads) is a plain literal.

Verified with per-function objdiff reports before/after: all 76 functions in the TU identical to the previous head (`ftColl_80077970` and `ftColl_8007925C` still 100%).

ribbanya: Fixed in 46cd7cf1bfb1cd7daf99ad91a9be7989a9537d62.

fjooord: Verified 1db8fd1 locally: the `if (dmg)` truthiness form compiles var-first `fcmpu` (unlike an explicit `!= 0.0f`, which mwcc canonicalizes constant-first) — so the extern casts were indeed unnecessary. Per-function objdiff report on the merged head shows zero regressions across the branch: `ftColl_80077970`/`ftColl_8007925C` still 100%, `fn_800D8EC8` 100% (master's version), `fn_800D9930` unchanged at 98.8%, and the 17 newly-matched upstream functions intact. Full ninja build green.

PR #2656, opened 2026-06-11
Melee decomp: GM

fjooord:
## Summary

Exact-match decompilation of 7 file(s) (9 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Files

- `src/melee/gm/gm_1601.c`
- `src/melee/gm/gm_1832.c`
- `src/melee/gm/gm_18A5.c`
- `src/melee/gm/gm_1A4C.c`
- `src/melee/gm/gm_1BA8.c`
- `src/melee/gm/gmregclear.c`
- `src/melee/gm/gmresult.c`

## Verification


src/melee/gm/gm_1832.c
```diff
@@ -1912,6 +1916,7 @@ bool gm_8018841C(void)
 }
 
 static TrainingModeState lbl_80473700;
+extern const f32 lbl_804DA60C;
```
PsiLupan: This is a regression that we've talked about before. Using an extern to make a function match is just due to data ordering.

src/melee/gm/gm_1832.c
```diff
@@ -2378,7 +2384,9 @@ void fn_801891F4(void)
                 lbAudioAx_80024030(8);
                 item = *(s16*) &((s32*) lbl_803D9828)[sub->menu_values[1]];
                 jobj = Player_GetEntity(0)->hsd_obj;
-                HSD_ASSERT(979, jobj);
+                if (jobj == NULL) {
```
PsiLupan: This is a regression and should be a call to something like `HSD_JObjGetTranslation2` since it's only has a single assert here.

src/melee/gm/gm_18A5.c
```diff
@@ -6292,16 +6293,16 @@ void fn_801981A0(HSD_GObj* gobj)
...
-    if (*(&lbl_804799D8.x44[pnum]) == 7) {
-        fn_8019044C(jobj, (f32) lbl_804799D8.x21[pnum]);
+    if (state_ptr[pnum + 0x44] == 7) {
+        fn_8019044C(jobj, (f32) state_ptr[pnum + 0x21]);
         return;
     }
 
-    fn_8019044C(jobj, (f32) (lbl_804799D8.x21[pnum] + 0x14));
+    fn_8019044C(jobj, (f32) (state_ptr[pnum + 0x21] + 0x14));
 }
```
jurrejelle: Introducing pointer math here is a regression, this shoudl b rthe field at offset 0x14 of the pointer indexed at pnum

src/melee/gm/gm_1BA8.c
```diff
@@ -581,11 +581,12 @@ void gm_801BAD70(GameScene* arg0)
 
 void gm_801BB758(GameScene* arg0)
 {
-    MatchExitInfo* exit = gm_801A4284(arg0);
+    u8* data = (u8*) &gm_80497758;
```
jurrejelle: Same here, don't introduce new u8*'s if possible, use struct accesses etc

src/melee/gm/gmregclear.c
```diff
@@ -3153,44 +3155,47 @@ int fn_80181BFC(int* arg0)
...
         {
-            var_r29++;
+            var_r30++;
         } else {
-            sp38 = var_r30;
+            sp38 = var_r29;
         }
-    }
+        var_r29++;
+    } while (var_r29 < 6);
```
jurrejelle: This is most likely a for loop. If you need the initializaton in that order, do it in the for loop start

jurrejelle (changes requested):

src/melee/gm/gm_18A5.c
fjooord: Fixed in b6b9840 — replaced the raw u8* offsets with a typed pointer to the state struct: `state->x2A[pnum].state`, `state->x21[pnum]`, and `state->x44[pnum]`. (The `+ 0x14` in the last `fn_8019044C` call is a value offset added to the animation frame, not a field offset, so it stays.) The compiled gm_18A5.o is byte-identical to the previous build.

src/melee/gm/gm_1BA8.c
fjooord: Fixed in b6b9840 — dropped the u8* alias. The offsets actually resolve inside `gm_804979D8` (a `MatchExitInfo`), not `gm_80497758`, so they're now proper field accesses: `&gm_804979D8.match_end`, `player_standings[0].stocks` / `.percent`, and the existing `(u8*) exit + 0x10/0x14/0x6C` casts became `exit->match_end.result` / `.frame_count` / `.player_standings[0].stocks`. Instruction bytes are unchanged (only the relocation symbol moved to the containing object) and the linked main.dol is still byte-identical to retail.

src/melee/gm/gmregclear.c
fjooord: Fixed in b6b9840 — converted back to a for loop with the ordered initialization in the for statement: `for (var_r29 = 1, var_r30 = 0, sp10 = data->xC; var_r29 < 6; var_r29++)`. The compiled gmregclear.o is byte-identical to the previous build.

jurrejelle (approved):

PR #2657, opened 2026-06-11
Melee decomp: GR

fjooord:
## Summary

Exact-match decompilation of 5 file(s) (4 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Files

- `src/melee/gr/grcastle.c`
- `src/melee/gr/grgreatbay.c`
- `src/melee/gr/gricemt.c`
- `src/melee/gr/grkongo.c`
- `src/melee/gr/grvenom.c`

## Verification


src/melee/gr/gricemt.c
```diff
@@ -1479,9 +1479,13 @@ int grIceMt_801F9ACC(Ground_GObj* gobj, float y, GrIceMtSegmentLookup ev,
     }
     f = grIceMt_801F993C(seg[0], seg[1]);
     mgobj = Ground_801C2BA4(seg[1]);
-    HSD_ASSERT(0xAB9, mgobj);
+    if (mgobj == NULL) {
```
PsiLupan: Regression. Keep the original HSD_ASSERT.

src/melee/gr/grkongo.c
```diff
@@ -659,7 +661,7 @@ void grKongo_801D828C(HSD_GObj* gobj)
         return;
     }
     if (gp->gv.kongo.u.taru.keep == NULL) {
-        __assert(grKg_803E1858, 1719, "gp->u.taru.keep");
```
PsiLupan, line 662: Don't change this to the extern char. It was already declared and it should be removed in the other place using it.

src/melee/gr/grkongo.c
```diff
@@ -1555,27 +1558,26 @@ void grKongo_801D7BBC(HSD_GObj* gobj)
...
         f32 r = HSD_Randf();
-        gp->gv.kongo.xE4 =
-            (s16) ((grKg_804D6980->unk4 - grKg_804D6980->unk0) * r +
-                   grKg_804D6980->unk0);
+        f32 delta = grKg_804D6980->unk4 - grKg_804D6980->unk0;
+        gp->gv.kongo.xE4 = (s16) (delta * r + grKg_804D6980->unk0);
     }
     var_f31 -= grKg_804D6980->unk8;
-    if (var_f31 < 0.0f) {
+    if (var_f31 < grKg_804DAFA0) {
```
PsiLupan: As mentioned on other PRs, this is a regression. It's trying to fix sdata2 ordering.

src/melee/gr/gricemt.c
```diff
@@ -1135,23 +1135,23 @@ s32 fn_801F8E58(Ground_GObj* arg0, s32* out)
...
     s16 b;
     s32 d;
 
     max = 0;
     gp = arg0->user_data;
-    timer = &gp->gv.icemt.xDC;
 
     {
         Ground* g = gp;
+        s32* p = &list[max];
```
jurrejelle, line 1147: can you try turning this into a for loop of 0..i..12?

src/melee/gr/gricemt.c
jurrejelle: You marked these as resolved but they are still ___asserts :P

src/melee/gr/gricemt.c
```diff
@@ -1754,12 +1770,17 @@ int fn_801FA4CC(int num)
...
+    return gobj->hsd_obj;
+}
+
 int grIceMt_801FA500(HSD_GObj* arg0, HSD_JObj* arg1)
 {
     s32 count = 0;
     HSD_JObj* jobj;
 
-    jobj = arg0->hsd_obj;
+    jobj = grIceMt_801FA500_get_jobj(arg0);
```
jurrejelle: can this use GET_JOBJ

src/melee/gr/grkongo.c
```diff
@@ -659,7 +661,7 @@ void grKongo_801D828C(HSD_GObj* gobj)
         return;
     }
     if (gp->gv.kongo.u.taru.keep == NULL) {
-        __assert(grKg_803E1858, 1719, "gp->u.taru.keep");
+        __assert(grKg_803E1858, 1719, grKg_803E1A00);
```
jurrejelle: Not solved either

jurrejelle (changes requested):
EDIT: I was looking at an older copy, oops

src/melee/gr/grkongo.c
fjooord: I believe this was removed from the PR as I could not resolve fully

src/melee/gr/gricemt.c
fjooord: Done in 363fcfe — converted to `for (i = 0; i < 12; i++)`. Kept the walking `Ground* g` / `s32* p` pointers since indexing `(&gp->gv.icemt.xDC)[i]` shifts register allocation and breaks the match. Object is byte-identical.

src/melee/gr/gricemt.c
fjooord: Done in 363fcfe — dropped the one-off helper and used `HSD_GObjGetHSDObj` directly (same accessor `GET_JOBJ` wraps). The `GET_JOBJ` macro itself doesn't match here: its explicit `(HSD_JObj*)` cast on the inline's `void*` result makes mwcc allocate an extra stack temp (frame grows 32→40) and reshuffles registers. The bare inline call with implicit pointer conversion is byte-identical.

jurrejelle (approved):

PR #2658, opened 2026-06-11
Melee decomp: IF + MN + TY

fjooord:
## Summary

Exact-match decompilation of 6 file(s) (8 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Files

- `src/melee/if/textlib.c`
- `src/melee/mn/mncount.c`
- `src/melee/mn/mnevent.c`
- `src/melee/mn/mnmain.c`
- `src/melee/ty/tydisplay.c`
- `src/melee/ty/tylist.c`

## Verification


src/melee/mn/mncount.c
```diff
@@ -768,36 +779,77 @@ static void mnCount_InitUserData_noinline(MnCountData* userdata)
     mnCount_InitUserData(userdata);
 }
 
+static char lbl_803EFB60[0xA8] =
```
PsiLupan: This is a regression. These strings should not be access this way.

src/melee/ty/tydisplay.c
```diff
@@ -997,7 +1002,7 @@ void un_80319EF0(void)
             }
         }
         if (scale > 2.1474836e9f || scale < -2.1474836e9f) {
-            OSReport("*** tyDisplay Table Scale Irregul!\n");
```
PsiLupan, line 1000: Similar regression to the previous. Matching because of externs is not correct. It means data ordering isn't finished.

src/melee/ty/tydisplay.c
```diff
@@ -1982,21 +1986,20 @@ void un_8031BF34(s32 arg0)
...
-    un_80308250(base->x38, (s16) arg0, 0);
+    un_80308250(base->x38, (s16) (u16) arg0, 0);
     un_804D6F2C = un_803087F4(base->x38);
 
     HSD_JObjClearFlagsAll(anim->jobj[0], 0x10);
     HSD_JObjSetFlagsAll(anim->jobj[1], 0x10);
     HSD_JObjClearFlagsAll(anim->jobj[0], 0x10);
 
-    jobj = (HSD_JObj*) un_804D6F2C->hsd_obj;
+    jobj = un_8031BF34_inline();
```
jurrejelle, line 2002: can you use GET_JOBJ here? if that doesn't work this is also fine but wanted to check

src/melee/ty/tylist.c
```diff
@@ -292,40 +293,46 @@ void un_80312904(void* arg0, s8 arg1)
...
-    row->text0->pos_z = f31;
-    row->text0->font_size.x = un_804DDE30;
-    row->text0->font_size.y = un_804DDE34;
+    text = row->text0;
+    text->pos_x = f30;
+    text->pos_y = f29;
+    text->pos_z = f31;
+    text = row->text0;
+    text->font_size.x = un_804DDE30;
+    text->font_size.y = un_804DDE34;
```
jurrejelle: if I had to guess, these are floats, can you inline them?

jurrejelle (changes requested):

src/melee/ty/tydisplay.c
fjooord: Tried it — `GET_JOBJ(un_804D6F2C)` drops the match on `un_8031BF34` (98.17% → 97.09% fuzzy), the arg evaluation at the call site changes the codegen vs the no-arg inline. Keeping the inline helper.

src/melee/ty/tylist.c
fjooord: Good guess — they're this TU's own `.sdata2` literal pool. Inlined as `0.028f`/`0.029f` (byte-exact vs the original: `3ce56042`/`3ced9168`) in a2169de; function match state is unchanged.

jurrejelle (approved):

PR #2659, opened 2026-06-11
Melee decomp: SysDolphin

fjooord:
## Summary

Exact-match decompilation of 4 file(s) (2 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Files

- `src/sysdolphin/baselib/axdriver.c`
- `src/sysdolphin/baselib/cobj.c`
- `src/sysdolphin/baselib/particle.c`
- `src/sysdolphin/baselib/sislib.c`

## Verification


src/sysdolphin/baselib/particle.c
```diff
@@ -1016,6 +1016,32 @@ s32 fn_80392CCC(s32 channel)
     return 0;
 }
 
+static char lbl_8040A540[0x268] =
```
PsiLupan: You submitted this change before and it was called out as a regression. Please make sure you're reviewing these PRs a bit more closely before submitting, as I'd prefer not to be calling out the same issues again and again.

src/sysdolphin/baselib/particle.c
fjooord, line 1: Mentioned in discord, these are new pragmas, which is recommended against

jurrejelle (approved):

PR #2661, opened 2026-06-12
Fix MatToQuat to 100% match

kolt-mcb:
Fix `MatToQuat` in `src/sysdolphin/baselib/quatlib.c` to achieve 100% binary match.

**Problem:** The stack frame was 8 bytes too large.

**Solution:** Removed 2 unnecessary local variables:
- `f32 tr` — the trace expression was stored in a local then reused; inlined directly into the `if` condition and `sqrtf(1.0F + ...)` call
- `int k` — was just `nxt[j]` assigned to a temp; replaced all uses with `nxt[j]` inline

**Result:** The entire `quatlib.c` unit is now 100% matched across all 6 functions (MatToQuat, HSD_QuatLib_8037EB28, HSD_QuatLib_8037EC4C, HSD_QuatLib_8037ECE0, EulerToQuat, HSD_QuatLib_8037EF28).

Hi this is my first contribution, I have read the contribution guidelines but if I missed something I am happy to update this.

### DIFF
--- a/src/sysdolphin/baselib/quatlib.c
+++ b/src/sysdolphin/baselib/quatlib.c
@@ -15,6 +15,7 @@ s32 MatToQuat(Mtx m, Quaternion* q)
     f32 scale;
     int i;
     int j;
+    int k;
 
     lenCol[0] =
         sqrtf(m[0][0] * m[0][0] + m[1][0] * m[1][0] + m[2][0] * m[2][0]);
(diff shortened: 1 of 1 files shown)


jurrejelle:
Superseded by https://github.com/doldecomp/melee/pull/2662, thanks for putting this TU on my radar :D You had the right idea but sometimes stuff that can get inlined already is and thus doesn't actually decrease stack usage. Re-using a variable is one of the ways to solve this. Thanks again!

PR #2668, opened 2026-06-12
Improve mn menu declaration-order matches

itsgrimetime:
Summary
- Improve declaration/local ordering in several mn menu functions to better match MWCC register allocation.
- Improve mnSnap_80253184 from 92.61% to 95.07% and mnSnap_80254298 from 94.60% to 95.73%.
- Improve additional mninfo, mnmainrule, mnsound, and mnstagesw functions via pure local declaration reordering.


jurrejelle:
tbf I'm usually against these small-improvement-but-not-match-prs but honestly that stack stuff can get tedious to figure out so if it's actually improvements idm this PR

itsgrimetime:
> tbf I'm usually against these small-improvement-but-not-match-prs but honestly that stack stuff can get tedious to figure out so if it's actually improvements idm this PR

same - an agent opened it (I thought there was at least 100% match included) but as soon as I noticed it had none I moved it to draft. Can close if we prefer, but figured it's more signal for others to know what others are working on currently

ribbanya (approved):
I think we're starting really scrape the bottom of the barrel in terms of matches, so we should take what improvements we can get, especially if it's a uniform methodology like this PR.

PR #2669, opened 2026-06-12
Disable MWCC warning errors

### LINKED ISSUE #2667: Disable `-warn iserror` to allow UB matches
- [ ] Change `-warn iserror` to `-warn off` in [`melee-mwcc`](https://github.com/doldecomp/melee/blob/e8153ee4d4d041470a8edce647588fde01161f1c/.github/packages/build-linux/melee-mwcc)
- [ ] Fix [`publish-packages` workflow](https://github.com/doldecomp/melee/actions/runs/27143048951/job/80113178252#step:7:14) by addressing the issues it generates. Most of these seem like actual problems, like casting the wrong function type, and the rest can be disabled or worked around, like for example not using binary literals.
- [ ] Remove the `--warn-error` flag from `configure.py` (I don't think it's being used anywhere anyway)
- [x] Update the [decomp.me preset](https://decomp.me/preset/63) to use `-warn off`

itsgrimetime:
Resolves #2667.

## Summary
- stop adding `-warn iserror` from `configure.py` and switch the package `melee-mwcc` wrapper to `-warn off`
- keep the package issue checker green by ignoring intentional decomp callback cast warnings and fixing the remaining C99/prototype/enum diagnostics it reported
- replace binary literals with C99-compatible constants

## Verification
- `DYLD_LIBRARY_PATH=/opt/homebrew/Cellar/llvm/19.1.7_1/lib cargo run --manifest-path tools/issues/Cargo.toml --quiet` -> `Issues: OK`
- `python configure.py && ninja` -> `build/GALE01/main.dol: OK`
- `ninja`

## Note
The repo does not include a decomp.me preset definition, so the production preset still needs to be updated separately to use `-warn off`.


compile_flags.txt
```diff
@@ -24,6 +24,7 @@
 -Werror=typedef-redefinition
 -Wno-bitfield-constant-conversion
 -Wno-builtin-macro-redefined
+-Wno-cast-function-type-mismatch
```
ribbanya: I was hoping for it to actually address these.

src/melee/if/ifstatus.c
```diff
@@ -896,8 +897,8 @@ void ifStatus_802F66A4(void)
     DynamicModelDesc** mrk;
     DynamicModelDesc** num;
     HSD_Archive** arch;
-    s32 reset = 0; // = 0 should go away, but that requires configure to be ran
-                   // with --no-warn-error
+    s32 reset = 0; // = 0 should go away after the uninitialized-use warning is
+                   // resolved.
```
ribbanya:
```suggestion
    s32 reset;
```

ribbanya (changes requested):

compile_flags.txt
itsgrimetime: working on it now

ribbanya: This does get the job done, though. I think we should just merge this and open a new PR to reintroduce the warning, cause it does seem helpful.

ribbanya: ... Or not, that was fast.

ribbanya (approved):
A lot of these double casts are arguably worse than how it was before, but it's fine. At least they look ugly.

ribbanya:
Thank you for doing this. The last successful `public-packages` build was [10 months ago](https://github.com/doldecomp/melee/actions/runs/17037773430), so I don't have to worry about that anymore.

PR #2670, opened 2026-06-12
Implement style check script and CI

itsgrimetime:
## Summary

Tooling for the style-check series. Both source sweeps merged separately (#2675 jobj `0x10`→`JOBJ_HIDDEN`, #2676 assert macros), so this PR is **tooling only**.

Adds `tools/check/` — a small, extensible style checker:

- two checks, both enabled by default and disabled with `--no-<check>` (mirrors clang `-Wno-`): `jobj-flags` (raw `0x10` where `JOBJ_HIDDEN` is meant) and `assert-macros` (direct `__assert` calls expressible with the HSD assert macros; `@todo`-annotated sites exempt)
- `--msg-style std` (default) mirrors `melee-issues` (`tools/issues`): a `severity > category > [option] > file` hierarchy and an `Issues: OK` / `Issues: N` summary, with the same exit codes (0 / 65 `EX_DATAERR` / 74 `EX_IOERR`). `--msg-style actions` emits GitHub Actions annotations.
- comment/string-aware scanning that handles calls split across lines
- fixtures per check under `tools/check/tests/` (run: `python3 -m unittest discover -s tools/check/tests -t tools -p '*.py'`)

CI (`style-check.yml`) lints `src` on push and pull_request — findings fail the job with inline annotations, mirroring `melee-issues`. The unit tests are a development tool and are not run in CI.

Incorporates @ribbanya's review (module layout; opt-out per-check flags; `--msg-style`; `std` mirrors `melee-issues` + exit codes; type annotations; line-break handling; precompiled regex; tests as a separate module; lint src in CI, not the unit tests).


itsgrimetime:
~~actually gonna move it into the issues checker/repo tools script as @ribbanya suggested in discord~~
Actually just a separate python script because editing a script sitting inline in yaml sucks

tools/jobj_flags_check.py
ribbanya, line 1: Try passing this along to the LLM:
1. Rename the script to `check.py`
2. Create CLI flag for each of the checks (currently just the one for jobj flags). It should be opt-out instead of opt-in (`--no-jobj-flags`, etc.). This mirrors the behavior of `clang`.
3. Refactor the GH actions output option to be `--msg-style actions`. This mirrors the behavior of `configure.py` and `mwcc`
4. Make sure everything is properly type-annotated and assume Python 3.12
5. Handle line breaks in the function call scanning
6. Use regex instead of `LITERALS`. Precompile all regular expressions. This will later need to be refactored into parsing arbitrary flag sets and resolving them to defines.
7. Remove the self-test (refactor to separate local implementation)

... That's all for now.

ribbanya (changes requested):

tools/jobj_flags_check.py
itsgrimetime: All done in the latest push (force-pushed; this branch is the base of #2671, which I restructured the same way):

1. Renamed to `tools/check.py`.
2. Each check is opt-out via `--no-<check>` (auto-generated from the check registry — just `--no-jobj-flags` for now).
3. GitHub output is `--msg-style actions` now (default `std`), matching `configure.py`/`mwcc`.
4. Fully type-annotated against Python 3.12 (builtin generics, `X | None`).
5. Multi-line handled: the scanner blanks comments/strings, then paren-walks, so calls split across lines (including between the name and `(`) are caught — fixture added.
6. The literal set is now a precompiled regex matching the integer forms whose value is 16 (hex/decimal/octal + optional suffix). This immediately caught a `16U` site (`it_2725.c`) the old set missed — folded into the sweep, so it's 505 sites now, not 504. Structured so the later "arbitrary flag set → defines" generalization drops in (noted with a TODO).
7. Self-test removed; fixtures moved to `tools/test_check.py` (`unittest`, like `tools/asm-differ/test.py`).

One judgment call: I left a CI step that runs `test_check.py` before scanning — cheap insurance against the checker silently breaking, but happy to make it local-only if you'd prefer. #2671 adds a second check (`assert-macros`) to the same file as a worked example of the registry.

ribbanya: 1. Might as well just start a module and place both files in there. As more checks are added, they can be split into separate files. Rename from `check.py` to `check/main.py` (alongside `__init__.py`). Rename from `test_check.py` to `check/tests/jobj_flags.py`.
2. Remove the CI check, it's for development of the tool only.
3. `std` output should mirror `melee-issues`, the source code for which is in `tools/issues`. Also copy the exit codes.

I'm going to submit a separate PR with just the src changes, and then we can merge this with `master`.

ribbanya: There was a typo in my previous instructions, `json_flags` &rarr; `jobj_flags`.

tools/check/main.py
ribbanya, line 1: Cool, now incorporate the checks from #2671 and then we can merge.

ribbanya (changes requested):

tools/check/main.py
ribbanya: Also @itsgrimetime I appreciate the work and I'm sorry for using you as a liaison to Claude...

tools/jobj_flags_check.py
itsgrimetime: yep np i figured it was a typo. just pushed those changes, lmk if that matches what you were thinking

ribbanya (approved):

ribbanya (changes requested):
Did the CI check get lost?

tools/check/main.py
itsgrimetime: lmao its fine happy to help out here

itsgrimetime:
Both checks are already in — `tools/check/main.py` registers `jobj-flags` and `assert-macros`, with fixtures in `tools/check/tests/jobj_flags.py` and `tools/check/tests/assert_macros.py`. Folded the #2671 check in during the restructure, so this should be ready to merge?

Or maybe I misunderstood what you meant by "Remove the CI check, it's for development of the tool only."

ribbanya:
Oh I see, the bot interpreted "don't run the self-tests during CI" as "remove the source checks from CI." It should definitely be linting src on pull requests. Just not the unit tests.

itsgrimetime:
> Oh I see, the bot interpreted "don't run the self-tests during CI" as "remove the source checks from CI." It should definitely be linting src on pull requests. Just not the unit tests.

Ah gotchya - that was misunderstanding on both claude & I's part lol

.github/workflows/style-check.yml
```diff
@@ -0,0 +1,24 @@
+name: "Style Check"
+
+on: [push, pull_request]
```
ribbanya: You can add a `src/**` path to the condition.

.github/workflows/style-check.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.12"
+
+    # Lint the source tree with tools/check. Findings exit non-zero (65),
+    # failing the job, with inline annotations from --msg-style actions; an
+    # internal error exits 74. The unit tests (tools/check/tests) are a
+    # development tool and are not run here.
+    - name: Lint src
```
ribbanya:
```suggestion
    - name: Run checks
```

itsgrimetime:
okay should be good now? lmk if you want anything else adjusted

.github/workflows/style-check.yml
```diff
@@ -0,0 +1,24 @@
+name: "Style Check"
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  style-check:
+    name: Lint src
```
ribbanya:
```suggestion
    name: Check source for common style issues
```

ribbanya (changes requested):

ribbanya:
It would be useful to also have a CLI option that applies the fixes (to be run locally), instead of relying on an LLM to interpret and implement them. But that can be done later, or never, up to you.

.github/workflows/style-check.yml
```diff
@@ -0,0 +1,24 @@
+name: "Style Check"
```
ribbanya:
```suggestion
name: "Check code style"
```

ribbanya (changes requested):

ribbanya (approved):

itsgrimetime:
> It would be useful to also have a CLI option that applies the fixes (to be run locally), instead of relying on an LLM to interpret and implement them. But that can be done later, or never, up to you.

love that idea - im gonna be continuing to do some tooling improvements like this so will fold that in in some upcoming PRs. thanks for the help on this one

PR #2660, opened 2026-06-11
Improve mnDiagram input and init matching

itsgrimetime:
## Summary

Improves VS Records diagram matching by reshaping the main input handlers and diagram init paths.

- improves `mnDiagram_InputProc` and `mnDiagram2_HandleInput` source shape
- matches `mnDiagram_80243434` and `mnDiagram_802437E8`
- refines diagram asset/table overlays, type signatures, and visible-entry helper use
- keeps `mnDiagram_CursorProc` at the upstream match level while preserving the new improvements

## Matching

Compared against current `upstream/master` (`0b15e713`):

- overall fuzzy: `97.50852%` -> `97.520096%`
- overall matched code: `2,800,380` -> `2,801,924`
- overall matched functions: `18,791` -> `18,793`
- `src/melee/mn/mndiagram.c`: `96.47131%` -> `98.06187%`, matched functions `33` -> `35`
- `src/melee/mn/mndiagram2.c`: `91.13237%` -> `92.5753%`

Notable function changes:

- `mnDiagram_InputProc`: `93.74395%` -> `98.672615%`
- `mnDiagram2_HandleInput`: `89.3421%` -> `97.460526%`
- `mnDiagram_80243434`: `96.49789%` -> `100%`
- `mnDiagram_802437E8`: `84.41611%` -> `100%`
- `mnDiagram_802427B4`: `93.82222%` -> `98.844444%`

## Verification

- `git diff --check upstream/master..HEAD`
- `python configure.py && ninja build/GALE01/main.dol`
- `shasum -a 1 -c config/GALE01/build.sha1`
- `ninja build/GALE01/report.json`

PR #2671, opened 2026-06-12
Use HSD assert macros for hand-expanded asserts

itsgrimetime:
## Summary

Second class in the convention sweep series (after #2670, which this branch is stacked on — review the last two commits; will rebase once #2670 lands).

Converts direct `__assert(...)` calls to the `HSD_ASSERT` / `HSD_ASSERTMSG` / `HSD_ASSERTREPORT` macros (per #2660 review feedback to keep the assert macros), for the sites where that is expressible: first argument is `__FILE__` or the file's own basename string. Asserts passing a *different* filename (e.g. `"ftcolanimlist.c"` inside `ftcolanim.c`, asserts inlined from headers) cannot use the macros — they hard-code `__FILE__` — and are untouched.

- **102 of 107 in-scope sites converted** across ~50 files
- where the assert string preserves an original local's name, the local is renamed back to it (`synth.c` `ptr`/`buf` → `p`; `mnmain.c`/`mndeflicker.c` `data`/`menu` → `user_data`) so `#cond` stringification reproduces the original string exactly
- **5 sites keep the expanded form with a `@todo`**: under the macro the object bytes change (`ftchangeparam.c` float-compare polarity; `mnhyaku.c`, `id.c` — both inside inline functions) or a structural change is needed first (`tobj.c`, `plbonuslib.c`). The `@todo` records that a byte-matching form is still wanted.
- adds the `assert-macros` check to `tools/check.py` (the checker introduced in #2670): direct `__assert` calls expressible with the macros, with `@todo`-annotated sites exempt (the annotation is the visible exemption). Enabled by default; `--no-assert-macros` to disable. Fixtures in `tools/test_check.py`.

## Verification

- `python configure.py && ninja` before and after: **all 1046 compiled objects byte-identical** (per-object sha1), `shasum -a 1 -c config/GALE01/build.sha1` passes — every conversion that changed bytes was reverted and `@todo`'d instead
- `python3 tools/test_check.py`: assert fixtures cover the `__FILE__` form, own-basename form, foreign-filename non-match, `@todo` exemption, macro-definition regions, and calls split across lines; pre-sweep tree → all in-scope sites, swept tree → 0
- changed lines formatted with clang-format


tools/assert_macro_check.py
ribbanya, line 1: This needs to be a monolithic tool, see https://github.com/doldecomp/melee/pull/2670#discussion_r3406593997. If you want to create a Python module with multiple source files, that's fine, but it needs to have a singular entrypoint.

ribbanya (changes requested):

ribbanya (commented):
Might make sense to merge the script and the refactors in separate PRs.

ribbanya:
Let's close this, split out the src changes, and integrate the checks into #2670.

PR #2686, opened 2026-06-13
Melee decomp: GM menu and result matches

fjooord:
## Summary

- Exact-match decompilation for Game-mode menu/result translation units.
- Adds 11 newly exact function(s) across 7 file(s).

## PR Shape

- Scope: Game-mode menu/result translation units
- Review risk: Medium; one subsystem, several files with decomp-shape warnings.
- Header/naming churn: none

## Reviewer Notes

- This file-level slice also improves 36 still-unmatched function(s); no fuzzy or metric regressions were introduced.
- Files:
  - `src/melee/gm/gm_1601.c`
  - `src/melee/gm/gm_16F1.c`
  - `src/melee/gm/gm_18A5.c`
  - `src/melee/gm/gm_1A4C.c`
  - `src/melee/gm/gm_1B03.c`
  - `src/melee/gm/gm_1BA8.c`
  - `src/melee/gm/gmresultplayer.c`

## Verification

- Slice regression report: 11 new match(es), 0 broken matches, 0 fuzzy regressions, 0 metric regressions.


src/melee/gm/gm_18A5.c
```diff
@@ -6904,9 +6965,9 @@ void fn_80199AF0(void)
     }
 
     fn_8019035C(0, lbl_804D6670->models[0], mode, 0x1A, 3, 1, fn_801985D4,
-                0.0f);
+                lbl_804DA808);
```
fjooord: possible regression here, removing inline values

src/melee/gm/gm_18A5.c
```diff
@@ -7944,13 +8000,15 @@ void fn_8019B81C(s32* state)
...
 
+#pragma dont_inline on
 void fn_8019B860(TmData* tm)
 {
     fn_8019A158();
     fn_80199AF0();
     fn_80198BA0();
     fn_8018F888();
-    fn_8018E618(tm->entrants, 4.5f, tm->x2C);
+    fn_8018E618(tm->entrants, lbl_804DA810, tm->x2C);
```
fjooord: possible regression here, adding pragmas and removing inline

src/melee/gm/gm_1A4C.c
```diff
@@ -238,6 +239,42 @@ void gm_801A8114(HSD_JObj* arg0, int arg1)
     HSD_JObjSetScaleZ(transJobj, scale);
 }
 
+static char lbl_803DB49C[][0xC] = {
```
fjooord, line 241: possible issue here, should these strings be accessed like this?

src/melee/gm/gm_18A5.c
fjooord: Follow-up after checking this against the literal/data ownership standard: I do not think this is a regression. `lbl_804DA808` is already the existing file-local float label for `0.0f`, and this same TU already uses it for the same `fn_8019035C` animation-parameter slot in nearby setup paths like `fn_80198D18` and `fn_80198EBC`. So this is supported reuse of known data, not a new data anchor replacing an ordinary literal without evidence. The PR's slice/full regression reports are also clean.

src/melee/gm/gm_18A5.c
fjooord: Follow-up after checking the pragma/matching-tactic standards: I do not think this is a regression. The `4.5f` change uses the existing `lbl_804DA810` float label, already declared and used in this TU, so it has data-ownership evidence. The new `dont_inline` is also scoped only to this small wrapper and reset immediately; that fits the narrow-exception pattern already present in this file. With the PR's clean isolated and full regression reports, this looks like an evidence-backed codegen boundary rather than a broad pragma regression.

src/melee/gm/gm_1A4C.c
fjooord: Follow-up after checking the string/data ownership side: I do not think this is a standards issue. This is not replacing an ordinary source string literal at a use site with a generated symbol; it is modeling an existing `gm_1A4C.c`-owned `.data` string block. The split owns `.data` from `0x803DB430` to `0x803DBF78`, and `symbols.txt` already has the contiguous `lbl_803DB49C` through `lbl_803DB61C` strings at `0xC` stride, with the next symbol at `0x803DB628` (`33 * 0xC` bytes later). So `char[][0xC]` matches the observed fixed-width table layout rather than creating a fake string anchor. The PR regression reports are clean as well.

src/melee/gm/gm_1601.c
```diff
@@ -3488,42 +3542,59 @@ s32 fn_80167638(s32 arg0, Vec3* arg1, Vec3* arg2)
...
+                    p += 0xE;
+                    if (p[0x4E] == 0) {
+                        idx = 5;
+                    } else {
+                        idx = 0;
+                    }
+                }
+            }
+        }
     }
```
jurrejelle, line 3558: 1. this pointer math is a fakematch, 2. this is probably a for loop

src/melee/gm/gm_16F1.c
```diff
@@ -2186,6 +2163,7 @@ lbl_8046DBD8_t* gm_801736DC(void)
     return &lbl_8046DBD8;
 }
 
+#pragma pool_data off
```
jurrejelle: remove this pragma

src/melee/gm/gm_18A5.c
jurrejelle: BS, replace this

src/melee/gm/gm_18A5.c
```diff
@@ -6904,9 +6965,9 @@ void fn_80199AF0(void)
     }
 
     fn_8019035C(0, lbl_804D6670->models[0], mode, 0x1A, 3, 1, fn_801985D4,
-                0.0f);
+                lbl_804DA808);
     gobj = fn_8019035C(0, lbl_804D6670->models[2], 0, 0x1A, 3, 1, fn_80198824,
-                       0.0f);
+                       lbl_804DA808);
```
jurrejelle: also here and in any other places

src/melee/gm/gm_18A5.c
jurrejelle: ditto, bs for both

src/melee/gm/gm_1A4C.c
```diff
@@ -399,13 +436,9 @@ void gm_801A9094(void)
...
             HSD_JObjSetScaleZ(root, 1.8f);
-            if (joint == NULL) {
-                __assert("jobj.h", 0x2F5U, "joint");
-            }
+            HSD_ASSERT(0x2F5, joint);
             child = HSD_JObjLoadJoint(joint);
-            if (child == NULL) {
-                __assert("jobj.h", 0x2F7U, "jobj");
-            }
+            HSD_ASSERTMSG(0x2F7, child, "jobj");
```
jurrejelle: This isn't in jobj.h, so the asserts should be called there. Probably just calling a function that's there.

src/melee/gm/gm_1BA8.c
```diff
@@ -502,24 +573,24 @@ void gm_801BAD70(GameScene* arg0)
     if (level == 0x31) {
         r3b[0x8A] = 1;
         r3b[0xAE] = (u8) -1;
-        r3b[0x91] |= 0x80;
-        r3b[0xB5] |= 0x80;
-        r3b[0x91] |= 0x20;
-        r3b[0xB5] |= 0x20;
+        ((PlayerInitData*) (r3b + 0x84))->xD_b0 = 1;
+        ((PlayerInitData*) (r3b + 0xA8))->xD_b0 = 1;
+        ((PlayerInitData*) (r3b + 0x84))->xD_b2 = 1;
```
jurrejelle: please try if it's possible to anti-pointer-math this part, idaelly by figuring out the struct/etc of r3b. if it doesn't work that's fine but worth a shot

src/melee/gm/gmresultplayer.c
```diff
@@ -826,65 +805,113 @@ void fn_80178BB4(HSD_GObj* gobj)
...
-                HSD_JObjSetFlagsAll(data->player_data[i].jobjs[7],
+                HSD_JObjSetFlagsAll(data->player_data[(*i)].jobjs[7],
                                     JOBJ_HIDDEN);
             }
         }
 
-        i++;
-    } while (i < 6);
+        (*i)++;
+    } while ((*i) < 6);
```
jurrejelle: this is probably a for loop

jurrejelle (changes requested):

src/melee/gm/gm_18A5.c
fjooord: Resolved in the current head: the `fn_80199AF0` `fn_8019035C` call now passes the inline `0.0f` literal instead of `lbl_804DA808`. I also checked the nearby setup calls that were touched by this PR, and the matching `0.0f`/`80.0f` spots are literals again. Verified with full `ninja` and regenerated report.

src/melee/gm/gm_18A5.c
fjooord: Resolved in `fb025f2e9`: removed the review-introduced `lbl_804DA810` use here and restored the inline `4.5f` literal. The earlier `dont_inline` wrapper is already gone in current head. `fn_8019B860` reports 100.0% after regenerating `build/GALE01/report.json`.

src/melee/gm/gm_1601.c
fjooord: Resolved in current head: this is now a normal `for` loop over `ARRAY_SIZE(info->FighterMatchInfo)` with `idx` defaulting to 0, instead of the byte-pointer walk. Full `ninja` still passes; regenerated report has `fn_80167638` at 96.061226%.

src/melee/gm/gm_16F1.c
fjooord: Resolved in current head: the `#pragma pool_data off/on` pair around `gm_801736E8` was removed. Full `ninja` passes.

src/melee/gm/gm_18A5.c
fjooord: Resolved in current head: the second `fn_80199AF0` `fn_8019035C` call also now passes inline `0.0f`. I checked the PR-introduced `lbl_804DA808` substitutions in this setup path and they have been restored to literals.

src/melee/gm/gm_1A4C.c
fjooord: Resolved in current head: restored the explicit `__assert("jobj.h", ...)` calls around `HSD_JObjLoadJoint` instead of using the local `HSD_ASSERT` wrappers. Full `ninja` passes.

src/melee/gm/gm_1BA8.c
fjooord: Resolved in current head: the level `0x31` block now uses the structured `md->players[1/2]` fields for `spawn_dir`, `xD_b0`, and `xD_b2` instead of raw `r3b` offsets. I checked this in the regenerated report; `gm_801BAD70` remains an improvement at 90.99527%.

src/melee/gm/gmresultplayer.c
fjooord: Resolved in current head: the player init pass is now a `for (; *i < 6; (*i)++)` loop in `fn_80178BB4_init_players`, replacing the prior `do/while`. Regenerated report has `fn_80178BB4` at 100.0%.

jurrejelle (approved):

PR #2687, opened 2026-06-13
Melee decomp: Ground, Kongo, and Zebes stage matches

fjooord:
## Summary

- Exact-match decompilation for Ground manager plus Kongo/Zebes stage files.
- Adds 6 newly exact function(s) across 3 file(s).

## PR Shape

- Scope: Ground manager plus Kongo/Zebes stage files
- Review risk: Medium; stage-local files with no shared header churn.
- Header/naming churn: none

## Reviewer Notes

- This file-level slice also improves 11 still-unmatched function(s); no fuzzy or metric regressions were introduced.
- Files:
  - `src/melee/gr/ground.c`
  - `src/melee/gr/grkongo.c`
  - `src/melee/gr/grzebes.c`

## Verification

- Slice regression report: 6 new match(es), 0 broken matches, 0 fuzzy regressions, 0 metric regressions.

### DIFF
--- a/src/melee/gr/grkongo.c
+++ b/src/melee/gr/grkongo.c
@@ -94,10 +94,18 @@ grKg_StageData grKg_803E1800 = {
 };
 
 char grKg_803E1858[] = "grkongo.c";
+char grKg_803E1A00[];
+static inline char* grKongo_801D828C_TaruKeepMsg(void)
+{
+    return &grKg_803E1A00[0];
+}
 static const lbColl_80008D30_arg1 grKg_803B7FB0 = {
     1, 1, 361, 0, 0, 180, 0, 0, 0,
 };
-extern char grKg_803E1A00[];
+static const f32 grKg_804DAFA0 = 0.0f;
+static const f32 grKg_804DAFA4 = 1.0f;
+static const f32 grKg_804DAFCC = -99999.0f;
+static const f32 grKg_804DAFD0 = 3.4028235e38f;
 
 void grKongo_801D5490(Ground_GObj* arg0)
 {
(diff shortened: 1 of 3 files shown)


src/melee/gr/grzebes.c
```diff
@@ -1164,9 +1174,8 @@ void grZebes_801DA254(Ground_GObj* gobj, f32 level)
         result.g = (u8) (t * (f32) (c1.g - c2.g) + (f32) c2.g);
         result.b = (u8) (t * (f32) (c1.b - c2.b) + (f32) c2.b);
         result.a = 0xFF;
-        HSD_LObjSetColor(lobj, result);
+        grZebes_801DA254_inline2(lobj, result);
```
fjooord, line 1177: possible regression removing used of HSD

src/melee/gr/ground.c
```diff
@@ -3184,18 +3211,23 @@ s32 Ground_801C5840(void)
     return stage_info.x6E4[i];
 }
 
+#pragma push
```
fjooord: New pragmas

src/melee/gr/grzebes.c
fjooord: Follow-up after checking this against the decomp standards: this does not remove the HSD operation. The wrapper still bottoms out in `HSD_LObjSetColor(lobj, color)`, so the source-level behavior and runtime call target stay the same.

The change is better understood as a narrow codegen/stack-shape tactic: the previous direct call plus trailing `PAD_STACK(8)` became an inline wrapper path that preserves the same HSD helper call. Since the concern here was a possible HSD-use regression, I do not think this thread needs a source change.

src/melee/gr/ground.c
fjooord: Follow-up after checking this against the pragma standard: this is a new `global_optimizer off`, so it is only acceptable as an exception with tight scope and match evidence. This one is wrapped in `#pragma push` / `#pragma pop` around only `Ground_801C5878`, so it does not leak into `Ground_801C58E0` or later code.

The PR report also shows `Ground_801C5878` moving from 75.88% to 100.00% with no regression entry for the unit. Given that scope and evidence, I do not see this as a regression or an actionable issue unless someone finds an equivalent match without the pragma.

src/melee/gr/grkongo.c
```diff
@@ -652,21 +637,24 @@ void grKongo_801D6378(Ground_GObj* arg)
     return;
 }
 
+#undef __FILE__
+#define __FILE__ (&grKg_803E1858[0])
```
jurrejelle: Remove this, the file is already called that. Also, remove the grkongo.c mention up top while you're at it

src/melee/gr/grkongo.c
```diff
@@ -707,33 +695,52 @@ void grKongo_801D6518(Ground_GObj* arg)
...
+    s32 a;
+    s32 b;
+} grKg_801D651C_pair;
+
+typedef struct grKg_801D651C_pair_list {
+    grKg_801D651C_pair pairs[3];
+    s32 terminator;
+} grKg_801D651C_pair_list;
+
+extern const grKg_801D651C_pair_list grKg_803B7FD4;
```
jurrejelle: change extern to static

src/melee/gr/grkongo.c
```diff
@@ -865,8 +864,21 @@ static struct _struct_grKg_803E188C_0x18 grKg_803E188C[0xF] = {
...
+#define __FILE__ "jobj.h"
+static inline void grKongo_801D7E78_GetTranslation(HSD_JObj* jobj,
+                                                   Vec3* translate)
+{
+    HSD_ASSERT(979, jobj);
+    HSD_ASSERTMSG(980, translate, &grKg_803E19F4[0]);
+    *translate = jobj->translate;
+}
+#undef __FILE__
+#define __FILE__ "grkongo.c"
```
jurrejelle: This is a call to jobj's stuff

src/melee/gr/ground.c
jurrejelle: bad, remove it

src/melee/gr/grzebes.c
jurrejelle: This is fine yeah, it's just an line wrapper to prevent inlining

jurrejelle (changes requested):

jurrejelle (approved):

PR #2691, opened 2026-06-13
Melee decomp: Interface and trophy UI matches

fjooord:
## Summary

- Exact-match decompilation for IF text/HUD helpers and TY Figupon.
- Adds 7 newly exact function(s) across 4 file(s).

## PR Shape

- Scope: IF text/HUD helpers and TY Figupon
- Review risk: Low; UI-facing implementation files with no shared declaration churn.
- Header/naming churn: none

## Reviewer Notes

- This file-level slice also improves 4 still-unmatched function(s); no fuzzy or metric regressions were introduced.
- Files:
  - `src/melee/if/ifcoget.c`
  - `src/melee/if/ifstock.c`
  - `src/melee/if/textlib.c`
  - `src/melee/ty/tyfigupon.c`

## Verification

- Slice regression report: 7 new match(es), 0 broken matches, 0 fuzzy regressions, 0 metric regressions.

### DIFF
--- a/src/melee/if/textlib.c
+++ b/src/melee/if/textlib.c
@@ -32,8 +32,15 @@
 #include <MSL/string.h>
 
 /// ?
-/* 4D6E18 */ extern DevText* devtext_drawlist;
-/* 4D6E38 */ extern DevText* devtext_poolhead;
+/* 4D6E18 */ extern DevText* un_804D6E18;
+/* 4D6E38 */ extern DevText* un_804D6E38;
+/* 4DDC88 */ extern GXColor un_804DDC88;
+/* 4DDC8C */ extern GXColor un_804DDC8C;
+/* 4DDC90 */ extern GXColor un_804DDC90;
+/* 4DDC94 */ extern GXColor un_804DDC94;
+/* 4DDC98 */ extern GXColor un_804DDC98;
+/* 4DDC9C */ extern f32 un_804DDC9C;
+/* 4DDCA0 */ extern f32 un_804DDCA0;
 unsigned short un_804A26B8[1000];
 unsigned short un_804A284C[1000];
 short* un_804D6EB4;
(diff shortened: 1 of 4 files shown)


src/melee/if/textlib.c
fjooord, line 1: I checked this concern. The extern-looking additions fall into two cases: existing `.sdata2` literal-pool objects needed for the exact `DevText_Create` match, and existing shared `.sbss`/`.sdata` objects being referenced by their canonical symbol-map names. I do not think this is a new-extern regression; the line comments below call out why these are written this way.

src/melee/if/textlib.c
```diff
@@ -32,8 +32,15 @@
 #include <MSL/string.h>
 
 /// ?
-/* 4D6E18 */ extern DevText* devtext_drawlist;
-/* 4D6E38 */ extern DevText* devtext_poolhead;
+/* 4D6E18 */ extern DevText* un_804D6E18;
+/* 4D6E38 */ extern DevText* un_804D6E38;
+/* 4DDC88 */ extern GXColor un_804DDC88;
```
fjooord, line 37: I checked the base symbols for this block: `un_804DDC88` through `un_804DDCA0` were already existing `.sdata2` objects before this PR, so this is not adding new data state. `DevText_Create` is using the original binary literal-pool loads for the default background/text colors and `10.0f`/`16.0f`; keeping these as inline/local literals leaves the function just short of exact, which matches the bot report moving `DevText_Create` to 100% here.

src/melee/if/textlib.c
```diff
@@ -57,15 +64,15 @@ GXColor red = { 0xFF, 0x80, 0x80, 0xFF };
 GXColor green = { 0x80, 0xFF, 0x80, 0xFF };
 GXColor blue = { 0x80, 0x80, 0xFF, 0xFF };
 
-GXColor color_08 = { 0x40, 0x50, 0x80, 0x80 };
-GXColor color_0C = { 0xE2, 0xE2, 0xE2, 0xFF };
-GXColor color_10 = { 0xFF, 0x80, 0x20, 0xFF };
-GXColor color_14 = { 0xA0, 0xA0, 0xFF, 0xFF };
+GXColor un_804D5A08 = { 0x40, 0x50, 0x80, 0x80 };
```
fjooord, line 67: These definitions were already real `.sdata` objects at `804D5A08` through `804D5A14`; the change is an address-name rename of the existing globals, not new extern state. The address names make the later references line up with `symbols.txt`, though a future cleanup could add friendlier aliases once the surrounding functions are better understood.

src/melee/if/textlib.c
```diff
@@ -57,15 +64,15 @@ GXColor red = { 0xFF, 0x80, 0x80, 0xFF };
...
+GXColor un_804D5A0C = { 0xE2, 0xE2, 0xE2, 0xFF };
+GXColor un_804D5A10 = { 0xFF, 0x80, 0x20, 0xFF };
+GXColor un_804D5A14 = { 0xA0, 0xA0, 0xFF, 0xFF };
 
 static inline DevText* find_by_id(char id)
 {
     DevText* text;
-    for (text = devtext_drawlist; text != NULL; text = text->next) {
+    for (text = un_804D6E18; text != NULL; text = text->next) {
```
fjooord, line 75: These two `.sbss` slots are the same draw-list and pool-head state that `textdraw.c` locally aliases as `devtext_drawlist` and `devtext_poolhead`. Referencing the canonical `symbols.txt` names here keeps the cross-TU reference tied to the existing objects; if we want the friendlier names in both files, I would make those aliases shared instead of having a separate extern name drift from the symbol map.

fjooord (commented):
Checked the `textlib.c` extern/name concern. I do not think this is a new-extern regression; the added notes explain which symbols are existing `.sdata2` literal-pool objects for the exact `DevText_Create` match and which are existing shared `.sbss`/`.sdata` objects being referenced by canonical symbol-map names.

jurrejelle (approved):

Issue #2693, opened 2026-06-14
Tracking: tools/check style-check roadmap


itsgrimetime:
Following #2670 (`tools/check` with `jobj-flags` + `assert-macros`), collecting
proposed checks and tooling so we can agree on shape before implementing each.
Comments on scope/shape welcome before any PR.

## Tooling
- [ ] `--fix` mode — apply fixes locally (not CI). `jobj-flags` is byte-safe;
      `assert-macros` is best-effort (dev builds + verifies, `@todo`s stragglers).
      Dry-run diff by default. (suggested by @ribbanya)
- [ ] Generalize the flag check to config-driven `(function name, parameter index)
      → flag set`, resolving raw literals (incl. composite masks, via bit-shift
      decomposition like `tools/flags`) to the named flags. Subsumes `jobj-flags`.

## Proposed checks
- [ ] Arbitrary JObj flags — any raw literal in `HSD_JObj*Flags*` resolvable to
      named `JOBJ_*` flags (generalizes the current `0x10`→`JOBJ_HIDDEN` check).
- [ ] Fighter motion-state flags in `Fighter_ChangeMotionState` calls — raw
      literals → named flags.
- [ ] Extern data that should be literal (floats, strings, …). :warning: needs a rule
      to distinguish de-inlining anti-patterns from legit cross-TU pool-sharing
      externs (e.g. #2666 `extern f32 ftCo_804D90D0`), or it false-positives.
- [ ] Named identifier where a literal belongs (the use-site side of the above);
      same contextual caveat.

## References
- `tools/flags` — decomposes a literal into its bit-shifted flag set.

## Open questions for maintainers
- Which checks should be blocking CI vs advisory?
- Preferred config format for `(function, param) → flagset`?
- For the extern/literal pair: how do we reliably separate anti-pattern from
  intentional cross-TU externs?

ribbanya:
> - [ ] Generalize the flag check to config-driven `(function name, parameter index)
      → flag set`, resolving raw literals (incl. composite masks, via bit-shift
      decomposition like `tools/flags`) to the named flags. Subsumes `jobj-flags`.

Not config, just reusing the same code for multiple flag sets.

> - [ ] Extern data that should be literal (floats, strings, …). :warning: needs a rule
      to distinguish de-inlining anti-patterns from legit cross-TU pool-sharing
      externs (e.g. #2666 `extern f32 ftCo_804D90D0`), or it false-positives.

[...]

PR #2694, opened 2026-06-14
tools/check: add jobj-flags --fix

itsgrimetime:
Adds a local `--fix` mode to `tools/check` (#2693), so the mechanical fix can be applied directly instead of by hand (suggested by @ribbanya).

## Behaviour
- `python3 tools/check/main.py --fix [paths]` rewrites source in place; `--fix --dry-run` prints a unified diff and writes nothing.
- Only **jobj-flags** is auto-fixed: the raw hidden-flag literal is replaced with `JOBJ_HIDDEN`. That swap is purely mechanical and exact.
- **assert-macros stays detection-only.** Converting it needs judgement (negate the `if`-guard, match it against the assert text, and not every site is byte-identical), so those are reported, not rewritten. After a `--fix` run they're listed as "left for manual review."
- Fixes change only tokens, not layout — run clang-format afterward.

(Earlier revisions of this branch included a best-effort assert rewriter; dropped it — that conversion is judgement-laden (negate the `if`-guard, match it against the assert text, and not every site is byte-identical), so it's left to manual conversion or a future clang-tidy fixer. See #2693.)

## Verification
- 15 unit tests in `tools/check/tests`.
- **Replay:** running `--fix` on the pre-sweep tree reproduces #2675 with **zero content differences across all 73 files** (only clang-format line-wrapping differs, due to local clang-format version).


tools/check/main.py
ribbanya, line 1: Why does it insist on doing this crazy procedural parsing? Can't it do pure regex, or parser combinators?

itsgrimetime: I think regex can't capture all the cases for the assert stuff that exist in the project, but yeah i hate this too (hence not marking it ready for review yet lol). Was gonna look at clang query closer after giga bowser mentioned it in discord

itsgrimetime: parser combinators might be better tho, just an extra python dep

itsgrimetime: ok i tested out clang-query and clang-tidy locally but wondering if you have a preference: should the checks live alongside tools/issues, or as standalone clang-query matchers in CI?

ribbanya: If you want to build it into `tools/issues` that's fine. I'm not sure what the limitations are on standalone matchers.

ribbanya (approved):
My only issue is that this seems incredibly annoying to maintain as written. Would prefer to see one of the options referenced in [#2694 (comment)](https://github.com/doldecomp/melee/pull/2694/changes#r3408970564).

PR #2695, opened 2026-06-14
grbigblueroute: match C85C, DED4, DF80 sizes; improve CD20 and fix sdata2 ordering

kolt-mcb:
## Summary

Match function sizes and fix .sdata2 constant ordering in `grbigblueroute.c`.

### Function sizes

| Function | Target | Ours | Status |
|---|---|---|---|
| C85C | 1220 | 1220 | ✅ 100% |
| CD20 | 3452 | 3448 | ⚠️ 99.9% (4B gap) |
| DED4 | 164 | 164 | ✅ 100% |
| DF80 | 92 | 92 | ✅ 100% |

### .sdata2 constants

All 23 constants \[0-22\] match exactly. Target has a trailing `0.0` at \[23\] (no relocations, no symbol table entry — dead constant, likely MWCC padding artifact).

### Known gap: CD20 (4 bytes)

The 4-byte gap is from MWCC allocating 1 extra GPR in the prologue (`stmw r24` vs target `stmw r25`). This shifts register allocation by 1 throughout the function:

- Stack frame: 592B (match)
- FPR spills: f16-f31 (match)  
- GPR spills: r24-r31 (8) vs r25-r31 (7) in target

Extensive testing of declaration order, block scoping, volatile variations, ternary vs if-else patterns, and 90k+ decomp-permuter iterations all failed to reduce register pressure.

### Changes

**C85C:** Invert comparison pattern to match target size.

**CD20:** 
- Reorder case 0: move atan2f/position calculation before xC adjustment
- Change jobj ternary to if-else to match target control flow
- Compute 45.0F\*Ground_801C0498() inline in case 0 (discard result) to fix sdata2 constant ordering \[11-14\] without variable live range
- Compute 45.0F\*Ground_801C0498() inline in cases 1-3

**DED4:** Restructure to use `Ground_801C0498()` + `Ground_801C38BC()` pattern.

**DF80:** Change to `if (b.y > -3.0f)` and add `Ground_801C38BC()` in else branch.

### DIFF
--- a/src/melee/gr/grbigblueroute.c
+++ b/src/melee/gr/grbigblueroute.c
@@ -303,7 +303,8 @@ void grBigBlueRoute_8020BC68(Ground_GObj* gobj)
         gp->gv.bigblueroute2.tracks[0].jobj = jobj;
         if (jobj != NULL) {
             jobj = gp->gv.bigblueroute2.tracks[0].jobj;
-            HSD_JObjGetTranslation(jobj, &gp->gv.bigblueroute2.tracks[0].offset);
+            HSD_JObjGetTranslation(jobj,
+                                   &gp->gv.bigblueroute2.tracks[0].offset);
             lbVector_Sub(&gp->gv.bigblueroute2.tracks[0].offset, &origin);
         }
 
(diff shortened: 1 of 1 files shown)


kolt-mcb:
sorry my agent made this pr before it was ready.

PR #2707, opened 2026-06-15
Match grIceMt_801F686C

sc00ppp:
## Summary

- 1 matched function in `gricemt.c` (Icicle Mountain)

### Stage code (`gr/`)

**gricemt.c — `grIceMt_801F686C`** — Icicle Mountain background-platform
setup. Captures the active ground state (`Ground_801C49F8`) and primes the
stage flags, then (for stage id 76) runs three successive scans over the
terrain-row table `grIm_803E4068`, each picking a row id distinct from the
rows already in use (`xAC[0]`/`xAC[1]` and the previously selected ids). It
accumulates a per-platform Y offset through `grIceMt_801F993C` and positions
the three background GObjs (`grIm_804D69E8` / `grIm_804D69EC` /
`grIm_804D69F0`) via the inlined `HSD_JObjSetTranslateY`. Clears the three
GObj slots on the fallback path when the stage id doesn't match.

## Verification

- `wsl -d Ubuntu -- bash -lc "cd … && ninja"` → `build-linux/GALE01/main.dol: OK` (SHA1 match)
- Strict build (`--max-errors 0 --require-protos --map`) links cleanly
- `objdiff` reports 100.0% for `grIceMt_801F686C` (was 95.33%)
- No fuzzy regressions in adjacent functions/TUs


src/melee/gr/gricemt.c
```diff
@@ -263,40 +288,45 @@ void grIceMt_801F686C(void)
...
-        bg_gobj = grIceMt_801F71E8(grIm_803E4068[xAC[4]].id);
-        HSD_ASSERT(0x286, bg_gobj);
+        bg_gobj = grIceMt_801F71E8(grIm_803E4068[grIm_804D69F4->xAC[4]].id);
+        (bg_gobj) ? ((void) 0)
+                  : __assert(grIm_803E46F8, 0x286, &grIm_804D4710);
         jobj = bg_gobj->hsd_obj;
-        HSD_ASSERT(0x287, jobj);
-        HSD_JObjSetTranslateY(jobj, y_pos4);
+        (jobj) ? ((void) 0) : __assert(grIm_803E46F8, 0x287, &grIm_804D4718);
+        Fake_HSD_JObjSetTranslateY(jobj, y_pos4);
```
jurrejelle: Don't replace HSD_ASSERT wiith __asserts if it can be avoided

src/melee/gr/gricemt.c
```diff
@@ -180,7 +180,32 @@ void grIceMt_801F6868(bool id) {}
 
 #define ICEMT_FIELD_MAX 6
 
-/// #grIceMt_801F686C
+extern char grIm_804D4710;
+extern char grIm_804D4718;
+extern char grIm_804D4720;
+
```
jurrejelle: dont use extern chars, just unline the actual strings (even if that means a slightly lower match)

jurrejelle (dismissed):

ribbanya (changes requested):
The data in this file is a total mess. I'm going to clean it up a little at least.

ribbanya (dismissed):

PR #2710, opened 2026-06-16
link ft co shouldered

jurrejelle:


src/melee/ft/chara/ftCommon/ftCo_Shouldered.c
```diff
@@ -26,15 +24,12 @@
...
-float const ftCo_804D8750 = 0;
-float const ftCo_804D8754 = 1;
-double const ftCo_804D8758 = S32_TO_F32;
-
 /* 09C744 */ static void ftCo_8009C744(Fighter_GObj* gobj);
 
 void ftCo_8009C5A4(Fighter_GObj* gobj, FtMotionId msid)
 {
-    u8 _[8] = { 0 };
+    u32 unused1;
```
ThePlayerRolo, line 31: shouldn't this use the PAD_STACK macro?

jellejurre: No, you van only do that at the end of the list of declarations

PR #2711, opened 2026-06-16
link ft co thrown kirby

jurrejelle:
Still has some improvements that could be done to make it neater but I couldnt get em to work, link is link


src/melee/ft/chara/ftCommon/ftCo_ThrownKirby.c
```diff
@@ -97,19 +123,27 @@ static inline float inlineB1(Fighter_GObj* gobj)
...
+    fp->mv.co.thrownkirby.x8 = 0.0f;
     fp->mv.co.thrownkirby.x4 =
         vel_func(thrower_gobj, &fp->self_vel, fp->facing_dir);
     Fighter_UpdateModelScale(gobj);
-    HSD_JObjGetScale(jobj, &fp->mv.co.thrownkirby.scale);
-    Fighter_ChangeMotionState(gobj, msid, Ft_MF_SkipThrowException, 0, 1, 0,
-                              thrower_gobj);
+    // TODO inline this to HSD_JObjGetScale someway
+    if (jobj == NULL) {
+        __assert("jobj.h", 0x337, "jobj");
```
ThePlayerRolo, line 139: isn't there a macro for this?

src/melee/ft/chara/ftCommon/ftCo_ThrownKirby.c
```diff
@@ -25,20 +25,46 @@
...
+    (void) .5; // 40 08 00 ..
+    (void) 3.0; // 4f e0 00..
+    (void) -1.0f;
+    (void) +1.0f;
+}
+
+// moved here for linking purposes
+extern double __frsqrte(double);
+
+extern inline float sqrtf(float x)
```
ThePlayerRolo, line 46: this is very much a fake match, might as well leave it unlinked

src/melee/ft/chara/ftCommon/ftCo_ThrownKirby.c
jellejurre: Yeah, but it requires re-setting __FILE__ if you want it in the wrong file (which it is here due to the inline). I think having the comment here is fine

src/melee/ft/chara/ftCommon/ftCo_ThrownKirby.c
jellejurre: We link fake matches all the time

PR #2713, opened 2026-06-16
link ftmaterial

jurrejelle:


src/melee/ft/ftmaterial.c
```diff
@@ -315,12 +321,17 @@ void ftMaterial_800BF6BC(Fighter* fp, HSD_MObj* mobj, HSD_TExp* texp)
                 sp_cnst1.next = NULL;
             }
             sp_cnst1.reg = (u8) reg2;
-            sp18.r = sp168.a;
-            sp18.g = sp168.a;
-            sp18.b = sp168.a;
-            sp_cnst1.val = &sp18;
+            {
+                GXColor* color = (GXColor*) ((u8*) &sp_tevdesc - 4);
```
r-burns, line 326: 🤨

ThePlayerRolo, line 326: there HAS to be a better way to match this than something like this.

PR #2715, opened 2026-06-16
nix: attempt to speed up builds

r-burns:
by reducing build-time closure size

Nix is a little slower than the other CI builds, and I'm guessing it's mostly because of having to download a bunch of individual dev tools instead of pulling a single docker image. That's pretty unavoidable, but I'm going to see how much of a difference reducing the download size makes. Specifically:

* use stdenvNoCC - we don't need a native compiler, we're only compiling using our special gamecube toolchain
* use python3Minimal - we don't need any fancy python bells and whistles, we just need to run configure.py
* declutter compilers - we don't need Wii or ProDG compilers, or any GC tools other than the specific versions we've selected
* disable objdiff-gui - not needed by the CLI-only nix sandbox build, and in a shell, I can use a separate objdiff-gui from my PATH

Altogether that should shave a couple hundred megs off the download.

### DIFF
--- a/.nix/melee.nix
+++ b/.nix/melee.nix
@@ -1,12 +1,12 @@
 { lib
-, stdenv
+, stdenvNoCC
 , decomp-toolkit
 , devkitppc
 , fetchurl
 , mwcc
 , objdiff
 , ninja
-, python3
+, python3Minimal
 , wibo
 }:
 let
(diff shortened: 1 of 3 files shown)


r-burns:
Ah, forgot that secrets aren't visible from forks, so the cache isn't getting warmed up. Will recreate from an in-repo branch.

PR #2723, opened 2026-06-17
6/14: GR Stage Pass C

fjooord:
## Summary

Exact-match decompilation slice for GR stage pass C. Rebased onto current `origin/master` after upstream moved under the original local branch.

## Files

- `src/melee/gr/grflatzone.c`
- `src/melee/gr/grfourside.c`
- `src/melee/gr/grhomerun.c`
- `src/melee/gr/grkongo.c`
- `src/melee/gr/groldpupupu.c`
- `src/melee/gr/groldyoshi.c`
- `src/melee/gr/gronett.c`
- `src/melee/gr/grpura.c`
- `src/melee/gr/gryorster.c`

## Verification

- Rebased onto current `origin/master` (`f59b5e08b5`).
- Dashboard Open Draft gate applied the rebased slice onto the production baseline and reached the upstream Issues lint stage, which only reported an existing out-of-slice issue in `src/melee/gm/gm_18A5.c`.


src/melee/gr/grpura.c
```diff
@@ -555,18 +563,27 @@ void grPura_80213128(HSD_DObj* dobj)
     }
 }
 
+#pragma push
+#pragma dont_inline on
```
fjooord, line 561: Pragmas here, may need to resolve

src/melee/gr/grpura.c
```diff
@@ -155,6 +150,7 @@ HSD_GObj* grPura_80211E08_noinline2(int gobj_id)
     return grPura_80211E08_noinline(gobj_id);
 };
 
+#pragma auto_inline off
```
fjooord, line 153: pragmas here, may need to resolve

src/melee/gr/grkongo.c
```diff
@@ -636,6 +636,8 @@ void grKongo_801D6378(Ground_GObj* arg)
     return;
 }
 
+#undef __FILE__
+#define __FILE__ (&grKg_803E1858[0])
```
fjooord: This smells funky

src/melee/gr/grpura.c
fjooord: Kept the dont_inline behavior, but normalized the pragma scope from push/pop to dont_inline on/reset. I validated the full PR after this cleanup with main.dol OK and dol_diff clean.

src/melee/gr/grpura.c
fjooord: Left this auto_inline guard in place to preserve codegen around grPura_80211E08. The full PR validates with main.dol OK and dol_diff clean.

src/melee/gr/grflatzone.c
```diff
@@ -690,68 +696,74 @@ void grFlatzone_8021805C(Ground_GObj* gobj)
...
-                    ud[0xC7] = HSD_Randi(8);
-                    v = ud[0xC7];
-                    if (v == ud[0xC8]) {
+                    gp->gv.flatzone.xC7 = HSD_Randi(8);
+                    v = gp->gv.flatzone.xC7;
+                    if (v == (u8) gp->gv.pad_0[4]) {
                         goto loop_4;
                     }
-                } while (tbl[v * 5 + 1] == -1);
+                } while (((s16(*)[5])(tbl + 0x94))[v][1] == -1);
```
PsiLupan: This is pretty gross. Seems like it's more indicative that a table starts at 0x803E79D0 that is separate from 0x803E7940

src/melee/gr/grhomerun.c
```diff
@@ -376,18 +372,25 @@ void grHomeRun_8021D680(Ground_GObj* gobj)
...
             }
         }
-        vars->xE8_flags.b0 = 0;
-        if (vars->xD0 != NULL && vars->xCC == NULL) {
-            vars->xCC = grHomeRun_8021EC58(0);
-            lb_8000B1CC(vars->xD0, NULL, &pos2);
+        ((grHomeRun_MainGroundVars*) &gp->gv.homerun)->xE8_flags.b0 = 0;
+        if ((u32) gp->gv.unk.xD0 != 0 && (u32) gp->gv.unk.xCC == 0) {
+            Vec3 pos2;
+            gp->gv.unk.xCC = (int) grHomeRun_8021EC58(0);
```
PsiLupan: This seems like it should be an `intptr_t` or the return type should be fixed. Otherwise, we're going to be dealing with pointer issues later, since as indicated below - it's an `HSD_Text*`

src/melee/gr/grhomerun.c
```diff
@@ -410,62 +413,80 @@ void grHomeRun_8021D680(Ground_GObj* gobj)
...
+            } else {
+                break;
+            }
         }
 
-        jobj = GET_JOBJ(vars->rear_gobj);
-        HSD_ASSERT(0x257, jobj);
+        jobj =
+            GET_JOBJ(((grHomeRun_MainGroundVars*) &gp->gv.homerun)->rear_gobj);
+        HSD_ASSERTMSG(0x257, jobj != NULL, "jobj");
```
PsiLupan: Not really sure why this assert was changed. `HSD_ASSERT(0x257, jobj)` is equivalent.

src/melee/gr/grflatzone.c
```diff
@@ -549,6 +550,12 @@ bool grFlatzone_80217DFC(Ground_GObj* gobj)
     return 0;
 }
 
+typedef struct grFz_AnimRow {
+    s16 entries[5];
+} grFz_AnimRow;
+
+#define GRFZ_ANIM_ROWS(base) ((grFz_AnimRow*) ((base) + 0x94))
```
PsiLupan: This is just obscuring my original review issue behind a define and doesn't actually solve the problem.

src/melee/gr/grhomerun.c
```diff
@@ -410,62 +413,80 @@ void grHomeRun_8021D680(Ground_GObj* gobj)
...
 
         for (i = 63; i >= 0; i--) {
-            gobj2 = vars->gobjs[i];
+            gobj2 = ((grHomeRun_MainGroundVars*) &gp->gv.homerun)->gobjs[i];
             if (gobj2 != NULL) {
                 break;
             }
         }
-        HSD_ASSERT(0x239, gobj2);
+        HSD_ASSERTMSG(0x239, gobj2 != NULL, "gobj2");
```
PsiLupan: Same issue as my prior comment that was marked as resolved regarding using `HSD_ASSERT` here, which is equivalent.

src/melee/gr/grhomerun.c
```diff
@@ -410,62 +413,80 @@ void grHomeRun_8021D680(Ground_GObj* gobj)
...
+        jobj =
+            GET_JOBJ(((grHomeRun_MainGroundVars*) &gp->gv.homerun)->rear_gobj);
         HSD_ASSERT(0x257, jobj);
         HSD_JObjSetTranslateX(jobj, x - 1.5F * (2150.99F * Ground_801C0498()));
 
-        jobj = GET_JOBJ(vars->rear2_gobj);
-        HSD_ASSERT(0x259, jobj);
+        jobj = GET_JOBJ(
+            ((grHomeRun_MainGroundVars*) &gp->gv.homerun)->rear2_gobj);
+        HSD_ASSERTMSG(0x259, jobj != NULL, "jobj");
```
PsiLupan: Same issue I previously brought up regarding `HSD_ASSERT` - wasn't addressed on this line.

src/melee/gr/grhomerun.c
```diff
@@ -410,62 +413,80 @@ void grHomeRun_8021D680(Ground_GObj* gobj)
...
 
         HSD_JObjSetTranslateX(jobj, x - 0.5F * (2150.99F * Ground_801C0498()));
 
-        jobj = GET_JOBJ(vars->rear3_gobj);
-        HSD_ASSERT(0x25B, jobj);
+        jobj = GET_JOBJ(
+            ((grHomeRun_MainGroundVars*) &gp->gv.homerun)->rear3_gobj);
+        HSD_ASSERTMSG(0x25B, jobj != NULL, "jobj");
```
PsiLupan: Same `HSD_ASSERT` issue here.

src/melee/gr/grhomerun.c
```diff
@@ -410,62 +413,80 @@ void grHomeRun_8021D680(Ground_GObj* gobj)
...
 
         HSD_JObjSetTranslateX(jobj, x + 0.5F * (2150.99F * Ground_801C0498()));
 
-        jobj = GET_JOBJ(vars->rear4_gobj);
-        HSD_ASSERT(0x25D, jobj);
+        jobj = GET_JOBJ(
+            ((grHomeRun_MainGroundVars*) &gp->gv.homerun)->rear4_gobj);
+        HSD_ASSERTMSG(0x25D, jobj != NULL, "jobj");
```
PsiLupan: Same `HSD_ASSERT` issue here.

src/melee/gr/groldpupupu.c
```diff
@@ -359,14 +362,16 @@ void grOldPupupu_80210D10(Ground_GObj* gobj)
...
+            y = direction *
+                (((f32) grOp_804D6A98->x4 * ((2.0F * rand) - 1.0F)) + 20.0F);
+        }
         for (i = 0; i < count; i++) {
+            HSD_GObj* spawn;
             spawn = grOldPupupu_802108B4_noinline(2);
             if (spawn != NULL) {
-                jobj = spawn->hsd_obj;
-                HSD_ASSERT(0x216, jobj);
+                HSD_ASSERTMSG(0x216, (jobj = spawn->hsd_obj), "jobj");
```
PsiLupan: Unless this is an actual improvement to the match, same `HSD_ASSERT` issue.

fjooord:
Added follow-up fixes for the review comments in `041715e` and the clang-format-only cleanup in `608b922`.

The Flat Zone table issue is now handled directly instead of hidden behind an offset macro. I removed `GRFZ_ANIM_ROWS` and stopped deriving the animation lookup table from `grFz_803E7940`; both `grFlatzone_802174EC` and `grFlatzone_80218060` now use the actual `grFz_803E7A68` table symbol directly. That keeps the stage callback table and animation table separate in the source instead of relying on pointer arithmetic across adjacent data.

For Home-Run Stadium, the text pointer issue is handled by treating `grHomeRun_8021EC58` as returning `HSD_Text*`. In `grHomeRun_8021D680`, the value still has to pass through the generic ground-var storage slot, so the cast uses `intptr_t` only at that storage boundary; the value is converted back to `HSD_Text*` before writing `pos_x`, `pos_y`, and `pos_z`. That avoids carrying the pointer around as a plain integer while still fitting the existing ground union layout.

The remaining reviewed spots were the equivalent `HSD_ASSERTMSG(..., ptr != NULL, "ptr")` cases, which are now back to plain `HSD_ASSERT(line, ptr)`. Local validation: `ninja` passed with `build/GALE01/main.dol: OK`, and `pre-commit run clang-format --all-files` now passes.

fjooord:
@PsiLupan 

All HSD_ASSERTS fixed.

I believe the flatzone is fixed as well, but a little above my paygrade.

The comment above details what was done

src/melee/gr/grkongo.c
```diff
@@ -642,8 +642,9 @@ void grKongo_801D828C(HSD_GObj* gobj)
     if (gp->gv.kongo3.xC6 != 1) {
         return;
     }
-    HSD_ASSERTMSG(1719, gp->gv.kongo.u.taru.keep,
-                  grKongo_801D828C_TaruKeepMsg());
+    if (gp->gv.kongo.u.taru.keep == NULL) {
+        __assert(grKg_803E1858, 1719, grKongo_801D828C_TaruKeepMsg());
+    }
```
ribbanya: Replace with proper `HSD_ASSERT`, get rid of `grKongo_801D828C_TaruKeepMsg`.

src/melee/gr/grpura.c
```diff
@@ -283,6 +280,11 @@ void grPura_802120E0(Ground_GObj* arg0)
     gp->gv.pura.xC8 = 0;
 }
 
+const f32 grPu_804DBA70 = 0.0;
+const f32 grPu_804DBA74 = 2.0;
+const f32 grPu_804DBA78 = 30.0;
+const f32 grPu_804DBA7C = -30.0;
+
```
ribbanya:
```suggestion
```

ribbanya (dismissed):

PR #2756, opened 2026-06-24
Improve several decomp matches

dberweger2017:
Improves several source matches across a few files.


dberweger2017:
oops, didnt mean to commit that last regression

PR #2757, opened 2026-06-24
texpdag matches

r-burns:


config/GALE01/symbols.txt
```diff
@@ -25420,8 +25420,13 @@ dst$740 = .data:0x80407A20; // type:object size:0x10 scope:local
...
 list_type = .data:0x80407A84; // type:object size:0x1B scope:local data:string
-HSD_TExpDag_80407AA0 = .data:0x80407AA0; // type:object size:0x78 scope:global
-jumptable_80407B18 = .data:0x80407B18; // type:object size:0x40 scope:local
+lbl_80407AA0 = .data:0x80407AA0; // type:object size:0xA data:string
+lbl_80407AAC = .data:0x80407AAC; // type:object size:0x24 data:string
+lbl_80407AD0 = .data:0x80407AD0; // type:object size:0x13 data:string
+lbl_80407AE4 = .data:0x80407AE4; // type:object size:0x10
+lbl_80407AF4 = .data:0x80407AF4; // type:object size:0x10
+lbl_80407B04 = .data:0x80407B04; // type:object size:0x10
+jumptable_80407B18 = .data:0x80407B18; // type:object size:0x40 scope:local align:8
```
r-burns, line 25429: Jump tables are finicky matches. Was helpful to flip back/forth between objdiff view (which shows only `????` for the not-yet-linked data jump labels) and the asm, which actually has useful labels. I verified the correctness of the data match using the most-aggressive `Name or address, data value` diffing type.

src/sysdolphin/baselib/texpdag.c
```diff
@@ -489,39 +400,34 @@ void HSD_TExpSchedule(int num, HSD_TExpDag* list, HSD_TExp** result,
...
 }
 
-static s32 HSD_TExpDag_804D5FF8 = 0x7FF00;
-static HSD_TExp* HSD_TExpDag_804D5FFC = NULL;
+struct Clear {
+    HSD_TExp* exp;
+    int arg;
+};
+
+static HSD_TEArg HSD_TExpDag_804D5FF8 = { 0x00, 0x07, 0xFF };
```
r-burns, line 408: An interesting part of this match - the weird instruction ordering is due to HSD_TExpDag_804D5FF8 being a struct rather than individual scalar types. That also explains how to prevent HSD_TExpDag_804D5FFC from getting moved to a BSS-like section, despite being zero'd out: it's a contiguous part of a larger struct with initialized data, just only in the lower addresses.

src/sysdolphin/baselib/texpdag.c
```diff
@@ -828,50 +719,113 @@ int SimplifyThis(HSD_TExp* arg0)
...
+                CLEAR_ARG(arg0->tev.a_in[2]);
+            }
+            break;
+        }
+
+#if 0
         if (op != 0xE) {
             if (op < 0xE) {
                 if (op < 2) {
                     if (op >= 0) {
```
r-burns: When you see deeply nested scalar comparisons, that's usually a hint that it's supposed to be a switch statement.

There are multiple types of codegen for switch statements (see also jumptables, usually for very large or very branchy switches). m2c isn't smart enough (yet?) to RE this back into its switch-statement form.

src/sysdolphin/baselib/texpdag.c
```diff
@@ -615,46 +514,44 @@ int SimplifySrc(HSD_TExp* arg0)
...
     return result;
 }
 
 #define CLEAR_ARG(arg)                                                        \
     do {                                                                      \
-        clear_arg = HSD_TExpDag_804D5FF8;                                     \
-        clear_exp = HSD_TExpDag_804D5FFC;                                     \
-        *(u32*) &(arg) = clear_arg;                                           \
-        (arg).exp = clear_exp;                                                \
+        (arg) = HSD_TExpDag_804D5FF8;                                         \
```
r-burns, line 554: Converting this to a struct copy made it pretty trivial, so we can probably just remove it in the future. Didn't want to make the diff _too_ messy for this PR.

src/sysdolphin/baselib/texpdag.c
```diff
@@ -168,13 +159,9 @@ void CalcDistance(HSD_TExp** tevs, int* dist, HSD_TExp* tev, int num,
 
 int HSD_TExpMakeDag(HSD_TExp* root, HSD_TExpDag* list)
```
r-burns, line 160: Although I "improved" the match on this function a lot, lacking a very close match, I don't really trust what I did. Too easy to use the wrong array index or drop a necessary increment during my refactoring. If I revisit this in the future, I'll probably start from scratch with decomp.py, and I'd probably advise others to do the same.

src/sysdolphin/baselib/texpdag.c
```diff
@@ -686,25 +580,23 @@ int SimplifyThis(HSD_TExp* arg0)
...
-            switch (cur->tev.a_in[0].type) {
+            switch (arg0->tev.a_in[i].type) {
             case HSD_TE_TEX:
                 alpha_tex = i;
                 break;
             case HSD_TE_RAS:
                 alpha_ras = i;
                 break;
             }
-            cur = (HSD_TExp*) ((u8*) cur + sizeof(HSD_TEArg));
```
r-burns, line 707: Another good compiler optimization to know - array indexing on an incrementing counter gets converted to a dereference of a forward-marching pointer.

config/GALE01/symbols.txt
ribbanya: If you mouse over the `????` in objdiff it shows the offsets from the function base, which is how I do it.
<table cellpadding="0" cellspacing="0">
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

r-burns: Good to point that out; personally I find the friction (especially if I want to copy/paste an address into a grep) to be just too much and I go back to the asm.

PR #2766, opened 2026-06-28
Improve itflipper names

lukechampine:
I tried a different approach here: I gave Claude an MCP server that let it talk to Dolphin running with a debugger, then ran various experiments in-game and had Claude use the evidence to give proper names to all of the functions and fields. My hope was that having better names would give us some insight into how to crack the remaining two flipper functions. But in the end, the fix was a simple inline; the names didn't help much at all.

I decided to keep them anyway, since they are accurate as far as I can tell. If we'd rather save this sort of thing for later, I can revert and keep just the changes required for matching. Otherwise I'll probably start doing this for the other unmatched items as well.

Remaining item functions: 26
Remaining item TUs: 8

PR #2778, opened 2026-07-01
improve itflipper.c names

lukechampine:
Added doxygen comments as per https://github.com/doldecomp/melee/pull/2766


ribbanya (changes requested):
Thanks for doing this. Can we try streamlining the attribution rather than just adding it to the summary of the type? I asked DeepSeek about it and it said [this](https://chat.deepseek.com/share/xpkbtbczkjt0o93mhh).

lukechampine:
ok, added \ref tags

ribbanya (approved):

PR #2779, opened 2026-07-01
Match `Toy_8030813C` and `Toy_OnEnter_80311AB0`

kolt-mcb:
Two function matches in `ty/toy`, plus a cascade into `gm_1601`.

> **`Toy_8030813C`** — change `arg0` from `s16` to `int`. The EABI passes
> `s16` parameters sign-extended, so `s32 id = arg0` compiles to a plain
> move rather than a redundant `extsh`. The local prototype inside
> `Toy_80308250` is kept `s16` so that caller still truncates its `s32`
> argument. This also removes the redundant `extsh` in the two `gm_1601`
> callers `gm_801604DC` and `gm_80160564`, which pass an `lha`-loaded
> (already sign-extended) value.
>
> **`Toy_OnEnter_80311AB0`** — the buffer allocations used `sizeof(*ptr)`,
> which is the wrong size for the array pointers (`s16*`, etc.). Replaced
> with the buffer sizes documented in the commented-out `STATIC_ASSERT`s
> in `toy.static.h`.

`main.dol` sha1 is unchanged and `diff_changes` reports no functions broken.

> Prepared with AI assistance (function matching only — no globals or struct
> fields named, no data sections).

### DIFF
--- a/src/melee/ty/toy.c
+++ b/src/melee/ty/toy.c
@@ -2133,7 +2133,7 @@ void _Toy_80307F64(s32 arg0, s32 arg1)
     }
 }
 
-char* Toy_8030813C(s16 arg0, enum_t unused)
+char* Toy_8030813C(int arg0, enum_t unused)
 {
     char* ptr;
     s32 i;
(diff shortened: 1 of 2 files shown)


dberweger2017:
Wourld it not be better to keep Toy_OnEnter_80311AB0 allocation/memzero size fixes and leave Toy_8030813C alone.
That way you get one new match with no regressions. the toy.h prototype change makes gm_1601.c rebuild differently.

ribbanya (changes requested):
> Wourld it not be better to keep Toy_OnEnter_80311AB0 allocation/memzero size fixes and leave Toy_8030813C alone. That way you get one new match with no regressions. the toy.h prototype change makes gm_1601.c rebuild differently.

I think the signature of `Toy_8030813C` should facilitate its own match. @dberweger2017's fix on `master` right now matches the `gm` functions but not the function itself. I have them both matching here but haven't submitted it yet. It's still not a clean solution because it involves a conditional re-declaration which is surely not what was originally written.
https://github.com/ribbanya/melee/commit/25c6cebfe8863256934df72184776f93050fcbf3

As to the allocation sizes, yes, reverting to hardcoded values is not the right fix.

ribbanya:
Superseded by #2846.

PR #2780, opened 2026-07-01
Name pause related functionality (Camera methods, pause detection, callbacks)

Jacoby6000:
These renames are largely mechanical.  They're centered around pause/unpause functionality

Most of what I've determined here is derived from existing naming that has already been done + some deeper investigation of the code.  I've done some cursory checks in dolphin and am fairly sure about all of this, but if you suspect any issues I'm happy to dig deeper and pull up evidence/revise my findings.

I followed the contributing guide standards to the best of my ability, but some of the names may leave something to be desired. Happy to change them however.

I'll be adding comments to the PR diffs for clarification on parts that seem strange.


src/melee/cm/camera.c
```diff
@@ -3863,24 +3863,24 @@ s32 fn_8002F488(Vec4* arg0)
...
-    if ((arg0 < 0 || arg0 >= 6) && (u8) (arg0 - 10) > 1) {
-        arg0 = 0;
+    if ((pauserSlot < 0 || pauserSlot >= 6) && (u8) (pauserSlot - 10) > 1) {
+        pauserSlot = 0;
     }
-    if ((arg1 < 0 || arg1 >= 4) && (u8) (arg1 - 4) > 1) {
-        arg1 = 4;
+    if ((pauserId < 0 || pauserId>= 4) && (u8) (pauserId - 4) > 1) {
+        pauserId = 4;
     }
```
Jacoby6000, line 3879: This case is hit from [here](https://github.com/doldecomp/melee/pull/2780/changes#diff-5e54cd75328be7494ca7cd489eb2f4382e888d3667df954ad9a17464849a4541R2637) and every time SetUpPauseCamera is invoked via `fn_80165190`

The edge case makes the variable names I set on the parameters a little funky, but I think they make the most sense.

src/melee/cm/camera.c
```diff
@@ -3863,24 +3863,24 @@ s32 fn_8002F488(Vec4* arg0)
     return 1;
 }
 
-void Camera_8002F4D4(s8 arg0, s8 arg1, s32 arg2)
+void Camera_SetUpPauseCamera(s8 pauserSlot, s8 pauserId, s32 arg2)
```
Jacoby6000, line 3866: This switches the camera to the pause camera. "SetUp" may be a little misleading here. Open to better names.

src/melee/gm/gm_16AE.c
```diff
@@ -1366,15 +1374,15 @@ void gm_8016D32C_OnFrame(void)
...
             tmp->pause_timer--;
         }
     } else {
-        fn_8016CA68(tmp, 2);
-        if (tmp->unk_4 != 0) {
-            tmp->unk_4--;
+        gm_DoPauseChecksAndRoutine(tmp, 2);
+        if (tmp->unpause_timer != 0) {
+            tmp->unpause_timer--;
         }
```
Jacoby6000, line 1404: The unpause_timer name comes from the similarity with the pause_timer semantics. They both get set to 0xA (10), and they both have this decrementing behavior, and pause/unpause both gate behavior behind their respective timer

src/melee/gm/gm_16AE.c
```diff
@@ -1159,11 +1167,11 @@ void fn_8016CBE8(lbl_8046B6A0_t* arg0, int arg1)
     if (arg0->hud_enabled == 0) {
         return;
     }
-    if (arg0->x24C8.x2_4) {
+    if (arg0->x24C8.disable_pausing) {
```
Jacoby6000, line 1185: the disable_pausing rename here comes from the fact that both unpause and pause routines are gated behind the flag, and then all assignments to disable_pausing come from default game mode setups or from the rule's pause flag.

src/melee/gm/gm_16AE.c
```diff
@@ -1105,50 +1111,52 @@ void fn_8016CA68(lbl_8046B6A0_t* arg0, int arg1)
...
-    int i;
-    for (i = 0; i < PAD_MAX_CONTROLLERS; i++) {
-        pad = &HSD_PadCopyStatus[(u8) i];
+    int playerId;
+    for (playerId = 0; playerId < PAD_MAX_CONTROLLERS; playerId++) {
+        pad = &HSD_PadCopyStatus[(u8) playerId];
         if (pad->err == 0) {
             if (DbLevel >= 3) {
-                if ((pad->trigger & 8) && (pad->button & HSD_PAD_X)) {
+                if ((pad->trigger & HSD_PAD_DPADUP) &&
```
Jacoby6000, line 1156: HSD_PAD_DPADUP is 8 (1 << 3)

src/melee/gm/gmscdata.c
```diff
@@ -465,7 +465,7 @@ GameMode gm_803DACA4[] = {
         NULL,
         NULL,
         gm_801B2B7C_OnInit,
-        gm_803DDBE0_Scenes,
+        gm_CameraModeScenes,
```
Jacoby6000, line 468: Tracked this down by looking through what places were setting the pause check override. The scenes associated with camera mode are the only place I found IIRC

src/melee/gr/ground.c
```diff
@@ -3193,12 +3193,12 @@ f32 Ground_801C57F0(void)
     return stage_info.x6E0;
 }
 
-void Ground_801C5800(void)
+void Ground_EnableMatchCamera(void)
```
Jacoby6000, line 3196: "EnableMatchCamera" is meant to represent setting the camera to the standard mode (or fixed) during a fight (unpaused, active gameplay).  Open to better names here. Both methods invoked by this one simply set camera mode without any other ceremony.

ribbanya (changes requested):
Seems good to me, though the regression in `gm_DoPauseChecksAndRoutine` needs to be fixed.

Could you add Doxygen comments for the commentary you gave here? It's better if people can reference them by looking at the source or documentation. Our documentation builds to [here](https://doldecomp.github.io/melee/) and we use standard [Doxygen commands](https://www.saoe.net/blog/doxygen-cheatsheet/).

Jacoby6000:
> ... the regression in `gm_DoPauseChecksAndRoutine` needs to be fixed

Done

> Could you add Doxygen comments for the commentary you gave here?

Done 

:smile:

src/melee/gm/gm_1BA8.h
```diff
@@ -108,7 +108,13 @@
 /* 1BF8D8 */ void gm_801BF8D8(GameScene*);
 /* 1BF8F8 */ void gm_801BF8F8(GameScene*);
 /* 1BF920 */ void gm_801BF920(GameScene*);
-/* 3DDBE0 */ extern GameScene gm_803DDBE0_Scenes[5];
+/**
+ * Scene table for #GM_CAMERA_MODE.
+ *
+ * The only game-mode scenes that install a @c x40_check_for_pauser_override
```
ribbanya: You can reference a field like this:

```suggestion
 * The only game-mode scenes that install a #StartMeleeRules::x40_check_for_pauser_override
```

src/melee/mn/types.h
```diff
@@ -179,11 +182,20 @@ struct StartMeleeRules {
...
+              // pauser playerId when a player unpauses the match. If not set,
+              // falls back to Ground_EnableMatchCamera x4_0 must also be true.
+    void (*x3C_on_pause_override)(
+        int); // on pause callback. When set, this method is called when a
+              // player presses Start during an unpaused match. Otherwise,
+              // gm_EnablePlayerPauseCamera is called x4_0 must also be true.
+    int (*x40_check_for_pauser_override)(
+        void); // When set, this method is used for checking if a player/which
+               // player has pressed pause while unpaused. Otherwise falls back
+               // to gm_DefaultVSGetPauser (x4_0 must also be true)
```
ribbanya: Make these Doxygen comments. Also, not your fault, but prefixing with the offset should be considered deprecated:
```suggestion
    float x30; // damage ratio
    float x34; // game speed
    void (*on_unpause_override)(
        int); ///< on unpause callback. When set, this method is called with the
              ///< pauser playerId when a player unpauses the match. If not set,
              ///< falls back to Ground_EnableMatchCamera x4_0 must also be true.
    void (*on_pause_override)(
        int); ///< on pause callback. When set, this method is called when a
              ///< player presses Start during an unpaused match. Otherwise,
              ///< gm_EnablePlayerPauseCamera is called x4_0 must also be true.
    int (*check_for_pauser_override)(
        void); ///< When set, this method is used for checking if a player/which
               ///< player has pressed pause while unpaused. Otherwise falls back
               ///< to gm_DefaultVSGetPauser (x4_0 must also be true)
```

ribbanya (changes requested):
It would be preferable to have a clean `.dox` file (which is just a `.h` file with a different extension) for most of this, especially the longer comments, but I'm okay with merging as-is and doing a cleanup pass later, since this wouldn't be the only example of using the headers for documentation.

src/melee/mn/types.h
NewtonMind: Real pre-existing quirk — agreed. Out of scope for this verbatim, behavior-preserving relocation; both dose-response lag issues are now explicitly folded into the ARCH P3c error-handling pass (typed exceptions + these two behavioral fixes, each with a test).

Jacoby6000:
> It would be preferable to have a clean `.dox` file (which is just a `.h` file with a different extension) for most of this, especially the longer comments, but I'm okay with merging as-is and doing a cleanup pass later, since this wouldn't be the only example of using the headers for documentation.

Noted for the future.. sorry about that

Jacoby6000:
Started seeing this in the builds

sjiswrap: File Z:\build\source\src\melee\cm\camera.h contains Shift JIS encoding errors

gonna try to get that fixed.  Build will pass, but probably should not merge

src/melee/gm/gm_1BA8.h
Jacoby6000: Updated all the references accordingly

Jacoby6000:
Should be good to go now

ribbanya (approved):

PR #2781, opened 2026-07-01
Improve decomp declaration matches

dberweger2017:
Small pile of decomp match improvements.

Mostly declaration ordering/lifetime tweaks, plus a few tiny expression-shape changes. The bot can give the exact score diff, but locally this built cleanly and the branch is source-only.


src/melee/ty/toy.c
```diff
@@ -2181,7 +2183,11 @@ char* Toy_8030813C(s16 arg0, enum_t unused)
 
 void Toy_80308250(u8* arg0, s32 arg1, s32 arg2)
 {
+#ifdef MWERKS_GEKKO
```
ribbanya: The PR seems good overall but I wish we could find a better solution for this kind of thing. Let's add a `/// @todo` comment for now.

ribbanya (changes requested):

src/melee/ty/toy.c
r-burns: Agreed, and I'd probably say we should drop these local overrides of function declarations if they're only improving fuzzy percentage. If I'm fiddling with a function declaration, I want to see ALL of its call sites changed, and not silently ignore my changes because of this stuff. I haven't seen anything like this yet in the codebase, so I think the final full match will be with a declaration that is visible globally and works for every caller.

ribbanya (approved):

PR #2782, opened 2026-07-01
[WIP] Big Session Run

fjooord:
This is an active decomp orchestrator session. 

Changes here have not been final-QA'd or broken down for review, it still needs to be split into reviewable pieces.

Please take from this what you want

---

### Update 7/5: 
- Will be breaking this down this week

### Update 7/13:
- Still have not broken down as changing jobs
- With release of 5.6 and removal of 5 hour rate limits, I have been just letting it run and it is consistently getting ~1% / 12hr, so just going to let it run until it plautues

PR #2784, opened 2026-07-02
Switch progress report to data-value diffing

r-burns:
This is a stricter diffing type which additionally looks at the data the relocs are pointing to, rather than just the names of the relocs. Now that we're far along in progress, this should be the default as we focus more on matching the data. I usually set this locally, and have been encouraging others to do so as well, but setting it here will also benefit our CI and AI assistants.

As suggested in https://github.com/doldecomp/melee/pull/2782#discussion_r3509722175


r-burns:
The clang-format / style-check workflows appear stuck for some reason, creating a new PR

PR #2787, opened 2026-07-03
grcastle data improvements

MarioLover63:


ribbanya (approved):
~~Looks like this file needs to be split, based on repeated literals in `.sdata2`.~~

Actually nevermind, looks like `double` 0 and `float` 0.

PR #2791, opened 2026-07-05
match grzebes .data

MarioLover63:


MarioLover63:
i disagree on the code style action, my structs are layed out fine

r-burns:
> i disagree on the code style action, my structs are layed out fine

Yeah I agree, this is an annoying behavior of clang-format. Want to wrap it in some `clang-format off`/`on` directives like this? Then you'll be green and we can merge.

https://github.com/doldecomp/melee/blob/6cbf21c97513ccf34fa6ba8ee4e2fef950f8bb28/src/melee/gm/gmmovieend.c#L71-L98

PR #2794, opened 2026-07-06
Disable auto-inlining for if/textdraw.c and if/textlib.c

MarkMcCaskey:
This came up while investigating something else and based on local testing I believe this keeps the match rate the same while removing manually opting out of inlining which seems like a code smell. That pattern is not uncommon elsewhere in the repo but I don't think any original source code would bother with opting out code from inlining, it was probably compiled with different flags which is what this change represents.

If this is intentional / a stylistic choice of the project we can close this PR.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1395,8 +1395,16 @@ config.libs = [
             Object(NonMatching, "melee/if/ifprize.c"),
             Object(NonMatching, "melee/if/ifcoget.c"),
             Object(NonMatching, "melee/if/soundtest.c"),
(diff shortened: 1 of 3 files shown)


MarkMcCaskey:
Based on discussion on Discord with @ribbanya, it sounds like inlining behavior may be determined by call depth. Rather than dealing with invisible layers of inlined functions or touching the flags (this code is not an independent library so it's unlikely to have special flags), it sounds like opting out per function is the best way to deal with this. Closing this PR as it's not helpful. Thanks for reviewing and sharing context!

PR #2786, opened 2026-07-02
Fix some data issues revealed by data-value diffing

r-burns:

PR #2800, opened 2026-07-06
Fix regressions from data-value fixes (#2786)

jurrejelle:
Builds on #2786 and fixes the 5 broken matches its bot report flagged, without giving up any of that PR's data/code gains.

## Root causes & fixes

**`grcorneria.c` — 4 regressions from one bug** (`grCorneria_801DCCFC` 100 → 94.65, plus `801DDDA8`/`801DDE88`/`801E2738`)
#2786 dropped `/* 0x000 */ u8 pad0[0x30];` from `struct grCn_Data` while keeping the field-offset comments at `0x030+`, shifting every field down by 0x30 so struct accesses compiled to the wrong offsets. Restored the `pad0` line, making all four functions byte-identical to their pre-PR form.

**`grbigblueroute.c` — `grBigBlueRoute_8020BC68`** (100 → 99.97)
The `jobj` rescoping added a stack slot, moving the `origin` Vec3 from `0x10` to `0x14`. Reverted to the original single function-scoped `jobj` and removed the now-stale `@todo`.

## Verification
- Full build passes, `main.dol` matches SHA-1.
- In the report (virtual-address diffing) all five functions are back to **100.00%**.
- #2786's gains are intact: new matches (`grCorneria_801DFC98`, `801E1060`, `grBigBlueRoute_8020C238`, `8020DED4`, …) still 100%, `.sdata` still 100%, `.data` gains preserved.


ribbanya (approved):

src/melee/gr/grbigblueroute.c
```diff
@@ -341,12 +339,10 @@ void grBigBlueRoute_8020BC68(Ground_GObj* gobj)
...
     }
 
-    {
-        HSD_JObj* reb0_jobj = Ground_801C2CF4(4);
-        HSD_ASSERT(452, reb0_jobj);
-        HSD_JObjGetTranslation(reb0_jobj, &gp->gv.bigblueroute2.xCC);
-        Ground_801C10B8(gobj, grBigBlueRoute_8020BC34);
-    }
+    jobj = Ground_801C2CF4(4);
+    HSD_ASSERT(452, jobj != NULL);
```
r-burns: Weird, this isn't reported as a non-match? Because I recall the data saying the assert message here should be "reb0_jobj"

r-burns: Anyway, won't block on this - lots of good fixes here and we can revisit this later.

PR #2804, opened 2026-07-06
Match fn_8019A86C to 99.86%

MarkMcCaskey:
Almost at 100% but not quite. I'll follow up with a decomp.me link later to sort out the final scratch register permutations but I tried a bunch of tricks to fix it and none worked so this PR is good to ship as-is, getting to 100% might be a big task.

From Claude:
> Take TmData* directly (the target's mr param-home shape), use a switch for the 0x1B dispatch, convert the hmn/active count blocks to their original 4-iteration loop, single-store ternary-style xC ordering, and index x44 as ((u8*) d8)[i + 0x44] to trigger the stride-1 strength reduction walker. Residual is two coalescing artifacts and a callee-saved rotation.

> The d8 pointer was the root cause of all three documented residuals. Its function-length web forced the materialization temp+copy, kept the base register alive so the fix loop couldn't destructively reuse it as a walker, and swapped the ready_count/i vs d8 register colors. With it gone, MWCC's own CSE of &lbl_804799D8 reproduces the target's exact base+copy structure — all three fixed in one edit. The old "cast-index form required" finding was an artifact of measuring with d8 still present; the stride-1 walker reduces fine from plain member syntax once nothing competes.

### DIFF
--- a/src/melee/gm/gmtou_1.c
+++ b/src/melee/gm/gmtou_1.c
@@ -2021,7 +2021,9 @@ void gm_8019A828(void)
 /// loop's x2A walker copies its base instead of reusing it) plus a
 /// callee-saved rotation (tm/arg2 swapped one slot down the r25-r29 order).
 /// The ((u8*) d8)[i + 0x44] form is required for the x44 strength-reduction
-/// walker; d8->x44[i] compiles to per-site address computation instead.
+/// walker (base must be the register variable d8 itself, displacement 0x44):
+/// d8->x44[i] scores 95.61%, a u8* alias of d8->x44 scores 96.66%, and
+/// lbl_804799D8.x44[i] scores 98.51%, vs 98.90% for this form.
 void fn_8019A86C(TmData* tm, u32 arg1, u32 arg2)
 {
     s32 ready_count = 0;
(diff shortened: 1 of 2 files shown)


src/melee/gm/gmtou_1.c
```diff
@@ -2016,17 +2016,26 @@ void gm_8019A828(void)
...
 
-void fn_8019A86C(s32* arg0, u32 arg1, u32 arg2)
+/// @todo 98.90%: all instruction shapes match except two coalescing artifacts
+/// (the &lbl_804799D8 materialization takes a temp+copy, and the ready-check
+/// loop's x2A walker copies its base instead of reusing it) plus a
+/// callee-saved rotation (tm/arg2 swapped one slot down the r25-r29 order).
+/// The ((u8*) d8)[i + 0x44] form is required for the x44 strength-reduction
+/// walker (base must be the register variable d8 itself, displacement 0x44):
+/// d8->x44[i] scores 95.61%, a u8* alias of d8->x44 scores 96.66%, and
+/// lbl_804799D8.x44[i] scores 98.51%, vs 98.90% for this form.
```
r-burns: Sounds like `lbl_804799D8.x44[i]` scores almost as well, so I'd prefer that to the pointer arithmetic. The other changes here LGTM

MarkMcCaskey: Good call, I was able to get a higher match % and remove that pointer math

PR #2805, opened 2026-07-06
Improve fn_8019A158 match to 97.75%

MarkMcCaskey:
Cleaning up this code entirely will require bigger structural changes, some of this code is clearly unrolled loops that will need data structures with arrays inside them to be written idiomatically.

Marking this PR as a non-draft as it's big but it's a real clean up that we can follow up on more.

From Claude:
> Restructure the slot search as the unrolled goto-loop, convert the four per-slot if-chains to their original 4-iteration walker loops, merge the zero init (x0 = mode = 0), rematerialize lbl_80473AB8 accesses in the case-2/propagate/tail blocks, and fix the stack frame. Residual is a callee-saved rotation and two scheduling artifacts.


src/melee/gm/gmtou_1.c
```diff
@@ -1684,25 +1717,35 @@ void fn_80199AF0(void)
...
 
+/// @todo 96.28%: all loops and shapes match; the residual is a callee-saved
+/// rotation (the lbl_804799D8 bases and the tail's entry/offset temps each
+/// colored one register off), a zero/constant register collision in the
+/// x4E-clear remainder loop, and the model-index load staying indexed
+/// (lbzx) instead of rematerializing the entry base. The four per-slot
+/// bracket loops keep byte-offset walkers: GET_BRACKET_SLOT only matches in
+/// straight-line code; in loops the cast base keeps the +0x2C out of the
+/// displacement, unlike a true member array (typing them needs a slots[4]
+/// array inside BracketEntry itself).
```
ribbanya: Simplify language; also don't talk about colored registers, it means nothing. If it wants to refer to objdiff output, speak in semantics like "regswap" or named registers like `r31`.

ribbanya (changes requested):

ribbanya (approved):

PR #2806, opened 2026-07-06
Refactor BracketEntry per-slot fields into a slots[4] array

MarkMcCaskey:
Stacked on #2805. Refactors the repeated per-slot fields of BracketEntry into a `slots[4]` array and updates uses.

From Claude:
> BracketEntry's fields from 0x2C to 0xDB repeat at a 0x2C stride, one block per bracket slot; gmtoulib.c already described this unit with a local `BracketEntrySlot` typedef. This moves that struct into gm/types.h, replaces the flat fields (x2C..xD8) with `BracketEntrySlot slots[4]`, and rewrites all uses in gmtou_1.c and gmtoulib.c (e.g. `ent->x7D` becomes `ent->slots[1].x51`). Verified with objdiff that every function in gmtoulib.o and gmtou_1.o is byte-identical to its previous state, except fn_8019A158 which improves 96.28% -> 96.88%.


MarkMcCaskey:
Comments about the merge from Claude:

> - Upstream had independently duplicated part of this PR. Master picked up a local BracketEntrySlot typedef + cast-based accessor in gmtoulib.c (#2877) and a GET_BRACKET_SLOT macro + second typedef copy in gmtou_1.c (#2805). The merge consolidates everything onto the PR's real slots[4] member in types.h and deletes both local duplicates; BracketEntry_GetSlot now reads &entry->slots[slot_idx].
> - fn_8019A158 keeps the branch's better version — master merged the older 97.75% improvement, the branch carries 98.32%. Verified against the target: master's build scores 97.69%, the merged build 98.29% (the small delta from the branch's recorded number is stale-target reloc naming, not code).
> - One resolution I deliberately left un-idiomatic: two draw functions master added in #2877 (fn_8018DC18, fn_8018DF68) use raw M2C_FIELD offset macros. I first resolved them with typed slots[]/x20 access, but that changed codegen (the cast form produces an lwzu walker that member syntax doesn't), dropping them to ~96-98% against master's build. So those keep master's exact M2C_FIELD spellings; converting them to typed access would be a separate matching effort if you want it.

src/melee/gm/gmtoulib.c
```diff
@@ -1573,11 +1580,14 @@ void fn_8018DF68(BracketEntry* arg0, s32 arg1, s32 arg2, s32 arg3, s32 arg4,
...
-        slot3_vertical_color = M2C_FIELD(arg0, GXColor*, 0x20);
-        DrawRectangle((f32) right, (f32) arg2, thickness, (f32) arg4,
-                      &slot3_vertical_color);
-        half = arg3 / 2;
-        slot3_horizontal_color = M2C_FIELD(arg0, GXColor*, 0x20);
+        {
+            GXColor* entry_color = &arg0->x20;
+            slot3_vertical_color = *entry_color;
+            DrawRectangle((f32) right, (f32) arg2, thickness, (f32) arg4,
+                          &slot3_vertical_color);
```
ribbanya: These casts are superfluous.

ribbanya (changes requested):

ribbanya (approved):

PR #2810, opened 2026-07-07
Link ftCo_WarpStar

vabold:


src/melee/ft/chara/ftCommon/ftCo_WarpStar.c
```diff
@@ -35,17 +35,21 @@
 /* 0C4A38 */ static void ftCo_800C4A38(Fighter_GObj* gobj);
 /* 0C4C60 */ static void ftCo_800C4C60(Fighter_GObj* gobj);
 
-/* literal */ float const ftCo_804D8D48 = 0;
-/* literal */ float const ftCo_804D8D4C = 1;
-/* literal */ float const ftCo_804D8D50 = M_PI_2;
-/* literal */ SDATA char ftCo_804D3C58[] = "jobj.h";
-/* literal */ SDATA char ftCo_804D3C60[] = "jobj";
+static inline const ftCo_DatAttrs* fpGetAttrs(const Fighter* fp)
```
ribbanya: I haven't seen the need for this kind of `const` declaration in our codebase so far:
https://decomp.me/scratch/ysu00

Not at my PC to edit it at the moment.

ribbanya (dismissed):

src/melee/ft/chara/ftCommon/ftCo_WarpStar.c
r-burns: I think there are situations where it can be nice to mark things like this as const, to make it clear that it's an input-only struct pointer and isn't going to be modified by the function. But yeah, that's not really a pattern we've been using yet so we can have a separate discussion about that.

jurrejelle (approved):

PR #2811, opened 2026-07-07
Match ft_800C85B8

MarkMcCaskey:
This is the last non-fully matched function in this TU! Is the linked metric a measure of the TUs with 100% matches weighted by size? Nice to be able to finish it off regardless

From Claude:
> Chain the zero initializations so both counters' reaching definitions are assignment expressions rather than literals. This stops constant propagation from pre-folding the strength-reduced dobj index (slwi r29, r26, 2 at the second loop's preheader), and lets the sp1C store recycle part_idx's zero register. All 165 instructions now match.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -668,7 +668,7 @@ config.libs = [
             Object(Matching, "melee/ft/ftCo_800C7590.c"),
             Object(Matching, "melee/ft/ftCo_800C78B0.c"),
             Object(Matching, "melee/ft/ftCo_800C7CA0.c"),
-            Object(NonMatching, "melee/ft/ftmetal.c"),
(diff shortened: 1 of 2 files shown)


r-burns:
> Is the linked metric a measure of the TUs with 100% matches weighted by size?

Not quite, your match here contributes to "complete" percent, but it isn't fully considered "linked" until we actually set it to Matching in the configure.py script so it gets linked in by the build process for real. That might require some more niche data-matching techniques, so you're welcome to take a crack at it yourself, or leave it to someone else with more linking experience. That's just the last step over the finish line and I'd consider the code matching here the more difficult part, so nice work!

PR #2815, opened 2026-07-07
Improve lbShadow_8000E9F0 match to 98.16%

MarkMcCaskey:
I investigated finishing up this TU, got some code simplification and an improved match on one and no progress on the other. Happy to remove or reduce the verbosity of the Claude comments if desired.

From Claude:
> Cases 0 and 3 now match exactly; cases 1 and 2 are down to one scheduler pair-swap and one fp-home chain (documented in @todo).
> 
> - case 3: natural car0..car3 statement order, u2 declared before tension, and the car0_y block hack removed (it was compensating for the old statement order; its 8 stack bytes are now PAD_STACK(8))
> - case 2: b2's block temp unwound to a plain assignment; b0's block temp kept (load-bearing); b0 computed before b1
> - case 1: bez0 before the compound bez1 statement fixes that case's register web
>
> Also documents lbShadow_8000F38C's single remaining instruction (li vs addi zero-recycle) as compiler-state-dependent after 18 variants and a codebase-wide twin classification.


ribbanya (changes requested):
> Happy to remove or reduce the verbosity of the Claude comments if desired.

Yeah these are slop, mostly because it frequently refers to numbers in its own memory (like "130 earlier variants"). Can you either have it spit out a tl;dr, or put in your own words what needs to be done on these functions?

MarkMcCaskey:
> > Happy to remove or reduce the verbosity of the Claude comments if desired.
> 
> Yeah these are slop, mostly because it frequently refers to numbers in its own memory (like "130 earlier variants"). Can you either have it spit out a tl;dr, or put in your own words what needs to be done on these functions?

Makes sense, yep I agree more spam than helpful. Will clean them up 🙏

ribbanya (approved):

PR #2819, opened 2026-07-09
gxset cpu fifo

stephenjayakar:
Testing out GPT 5.6 Sol High


.github/workflows/editorconfig.yml
ribbanya, line 1: Oh my god, did you finally fix this bug?

Anyway, can this be simplified into one `jq` invocation instead of jq-grep-jq?

ribbanya (changes requested):

extern/dolphin/src/dolphin/gx/GXFifo.c
```diff
@@ -159,12 +159,46 @@ do { \
     ASSERTMSGLINE(line, ((u32)(val) & ~((1 << (size)) - 1)) == 0, str_reg_field_out_of_range); \
     (reg) = ((u32)(reg) & ~(((1 << (size)) - 1) << (shift))) | ((u32)(val) << (shift)); \
 } while (0)
-asm void GXSetCPUFifo(GXFifoObj *fifo)
```
ribbanya: Remove this

extern/dolphin/src/dolphin/gx/GXFifo.c
ribbanya: And this

ribbanya (changes requested):

extern/dolphin/src/dolphin/gx/GXFifo.c
stephenjayakar: Wait, but it's no longer asm, should we remove that?

extern/dolphin/src/dolphin/gx/GXFifo.c
stephenjayakar: Added the `#pragma peephole on` back

ribbanya, line 1: This file is already linked, so it won't show as improving anything, but getting rid of these hacks is necessary.

Seems like `DEBUG` is being used elsewhere in this file for both real debug code, and for non-matching hackery, which is still a problem.

ribbanya (dismissed):

PR #2822, opened 2026-07-10
Some controller related renames, replace some button hex codes with readable defines

Jacoby6000:
This is the first in a series of PRs from the work I've been doing renaming stuff in Melee.

These changes stem from just a couple of controller related renames.   The "gm_GetButtonsPressedSinceLastFrame" method could probbaly use a more concise name.  On the controller struct, we use "trigger" to represent the bitmask of buttons pressed last frame, but that struct has its own comment attached about finding a better name, so I opted not to use "GetButtonsTriggered".


src/melee/gm/gm_1A36.dox
```diff
@@ -0,0 +1,29 @@
+#ifndef MELEE_GM_1A36_H
+#define MELEE_GM_1A36_H
+
+#include <placeholder.h>
+#include <platform.h>
```
ribbanya: You don't need preprocessor directives in `.dox` files unless you're annotating them, like for `PAD_ALL_CONTROLLERS`.

src/melee/gm/gm_1A36.h
```diff
@@ -4,12 +4,14 @@
 #include <placeholder.h>
 #include <platform.h>
 
-/* 1A3680 */ u64 gm_801A3680(u8);
```
ribbanya, line 7: Should be above the functions.

src/melee/gm/gm_1A36.h
```diff
@@ -4,12 +4,14 @@
...
+
+/* 1A3680 */ u64 gm_GetButtonsPressed(u8);
+/* 1A36A0 */ u64 gm_GetButtonsPressedSinceLastFrame(u8);
 /* 1A36C0 */ u64 gm_801A36C0(u8);
 /* 1A36E0 */ void gm_801A36E0(s32, s32);
 /* 1A3714 */ void gm_801A3714(s32, u64, u64);
-/* 1A3A74 */ UNK_RET gm_801A3A74(UNK_PARAMS);
+/* 1A3A74 */ UNK_RET gm_EvaluateAllControllerInputs(UNK_PARAMS);
 /* 1A3E88 */ UNK_RET gm_801A3E88(UNK_PARAMS);
 void gm_801A3EF4(void);
```
ribbanya: Should have its 6-digit virtual address like the other functions. Then in vim you can do the following to sort them by address:
```
:'<,'>!tools/sort-declarations.sh
```

ribbanya (changes requested):

src/melee/gm/gm_1A36.dox
Jacoby6000: Ah gotcha. There are a bunch of other .dox files that include all of this (and some that don't).  Would y'all like a followup PR that removes all the unnecessary preprocessor stuff from .dox files?

src/melee/gm/gm_1A36.h
Jacoby6000: It is, unless I'm misunderstanding

config/GALE01/symbols.txt
```diff
@@ -9274,14 +9274,14 @@ gmCamera_801A33BC = .text:0x801A33BC; // type:function size:0x140 scope:global
 gmCamera_801A34FC_OnFrame = .text:0x801A34FC; // type:function size:0x138 scope:global
 gmCamera_801A3634_OnEnter = .text:0x801A3634; // type:function size:0x48 scope:global
 gmCamera_801A367C_OnLeave = .text:0x801A367C; // type:function size:0x4 scope:global
-gm_801A3680 = .text:0x801A3680; // type:function size:0x20 scope:global
-gm_801A36A0 = .text:0x801A36A0; // type:function size:0x20 scope:global
+gm_GetButtonsPressed = .text:0x801A3680; // type:function size:0x20 scope:global
+gm_GetButtonsPressedSinceLastFrame = .text:0x801A36A0; // type:function size:0x20 scope:global
```
Jacoby6000: The more I think about this, the more I think I should make it `gm_GetButtonsTriggered` to be more consistent with the other controller api, even if that struct does currently mention looking for a better name.

src/melee/gm/gm_1A36.h
ribbanya: My bad, misread the diff.

src/melee/gm/gm_1A36.dox
ribbanya: Yeah, that'd be helpful.

ribbanya (dismissed):

PR #2825, opened 2026-07-11
Match five Melee functions

stephenjayakar:
## Summary

- match `itLinkbomb_UnkMotion3_Anim`
- match `grGreens_80214FA8`
- match `gm_801BF4DC`
- match `fn_8016CFE0`
- match `grKongo_801D828C`

These replace remaining assembly mismatches with typed C and small code-generation corrections, increasing matched code without changing runtime behavior.

## Validation

- all five functions report exactly `100.0` in objdiff
- `ninja progress`
- `ninja changes_all`
- strict `no_regressions.py` gate passed with no unit or function fuzzy-match decreases
- `git diff --check`


src/melee/gr/grgreens.c
```diff
@@ -914,7 +914,7 @@ void grGreens_80214FA8(Ground_GObj* gobj)
         if (jobj != NULL) {
             HSD_JObjSetFlags(jobj, JOBJ_HIDDEN);
         } else {
-            HSD_ASSERT(1203, 0);
```
ribbanya, line 917: Revert.

src/melee/gr/grkongo.c
```diff
@@ -634,11 +634,13 @@ void grKongo_801D6378(Ground_GObj* arg)
 
 void grKongo_801D828C(HSD_GObj* gobj)
 {
-    Ground* gp = gobj->user_data;
+    Ground* gp = GET_GROUND(gobj);
     if (gp->gv.kongo3.xC6 != 1) {
         return;
     }
-    HSD_ASSERT(1719, gp->gv.kongo.u.taru.keep);
```
ribbanya, line 641: Use `HSD_ASSERTREPORT`.

ribbanya (changes requested):

src/melee/gr/grkongo.c
stephenjayakar: > The exact reviewer-requested HSD_ASSERTREPORT macro emits an extra OSReport call and drops the function to 88.24%, so it cannot represent this assembly. The matching project idiom is HSD_ASSERTMSG, which removes the raw __assert while preserving the reviewer’s intended assertion message and restores grKongo_801D828C to 100%.

This is what I got. WDYT? Should we make another macro?

ribbanya: My bad, I meant `HSD_ASSERTMSG` in all of my comments.

ribbanya (approved):

PR #2826, opened 2026-07-12
Match ftCo_800CE650 / ft_0CE3

MarkMcCaskey:
This is an exact match but removes inline helpers making the code less clean. Will leave it up to the maintainers if this is a net win. Perhaps there's another way to fix the stack ordering.

From Claude:

> The former inline2/inline3 helpers reverse each expansion's stack-slot order; writing the expansions out flat with the temps declared in frame order, plus two 8-byte pads, produces the exact target layout. All sections of ft_0CE3.c are now byte-identical, so it moves to Matching.


ribbanya (approved):
> This is an exact match but removes inline helpers making the code less clean. Will leave it up to the maintainers if this is a net win.

This is _always_ preferred.

PR #2836, opened 2026-07-13
Rename and document game modes and scene transitions

Jacoby6000:
The second in a series of PRs going through the code and figuring out what is going on.

I've gone through and updated all game mode comparisons/references to use actual game mode enum values to make things a little easier to reason about. I've done the same with scenes where reasonable. Unfortunately most scene comparisons seem to use scene index.

A few key findings in this:
* curr_scene/pending_scene/previous_scene values are references to the index of a scene in a given scene graph for a given game mode. Not the scene kind.
* class_id in scene data is the actual GameSceneKind
* The SetPendingScene method is actually setting index + 1. So SetPendingScene(0) sets the pending scene to 1. 
* Scene graphs seem to be broken up in to contiguous groups. For example, some scenes will go 0x0,0x1, 0x2 (big gap) 0x80, 0x81, ...

This one wound up being quite a bit larger than I originally intended. Happy to split it up if desired.


fjooord:
Not all hero's wear capes

PR #2842, opened 2026-07-15
match yoshi specialS

r-burns:


src/melee/ft/chara/ftYoshi/ftYs_SpecialS.c
```diff
@@ -1237,23 +1206,36 @@ void ftYs_SpecialAirSStart_1_Coll(Fighter_GObj* gobj)
...
+static inline void stack_pad_hack(void)
+{
+    /// @todo egregious stack padding hack
+    GET_FIGHTER(NULL);
+    GET_FIGHTER(NULL);
+    GET_FIGHTER(NULL);
+    GET_FIGHTER(NULL);
+    GET_FIGHTER(NULL);
+    GET_FIGHTER(NULL);
+    GET_FIGHTER(NULL);
```
ribbanya, line 1218: Does this have the same effect as `FORCE_PAD_STACK_8`?

r-burns: Unfortunately no lol, the force padding enlarges the stack frame but I couldn't get it to bump up the stack offsets of local variables. Kinda hard to explain but try it out locally and you'll see what I mean.

PR #2843, opened 2026-07-15
Match ftKb_SpecialN_800F10D4, ftColl_80079AB0, and it_802A3630

MarkMcCaskey:
Found some similar already matched functions with the semantic search tool. Quite neat to see how it can identify that the assembly for Link's hookshot is very similar to Samus's grapple.

From Claude:
> - ftKb_SpecialN_800F10D4 → 100% (was 97.7). Token-identical to the Jigglypuff hat loader 800F11F0; transplanted its #pragma dont_inline + fp2-check shape and deleted the one-user _noinline wrapper. It's the Yoshi hat loader.
> - ftColl_80079AB0 → 99.82% calibrated-full (was 95.9). The donor ftColl_80079EA8 writes the knockback formula as one giant nested expression with one = 1.0F where ours has real ratio params — the nesting itself is the codegen lever (keeps more float-conversion temps live).
> - it_802A3630 → 99.96% calibrated-full (was 99.4). Cross-file donor fn_802B895C (samusgrapple): the fixes were passing fp2 = fp into the checking inline, passing the item_link variable instead of re-spelling the field expression, and a 4-byte pad above pos.

### DIFF
--- a/src/melee/ft/chara/ftKirby/ftkirby.c
+++ b/src/melee/ft/chara/ftKirby/ftkirby.c
@@ -4232,28 +4232,26 @@ void ftKb_SpecialN_800F10A4(Fighter_GObj* gobj)
     ftKb_SpecialN_800EF69C(gobj, 3, ft_80459B88.hats[FTKIND_CAPTAIN]);
 }
 
-static void ftKb_SpecialN_800EF040_noinline(Fighter_GObj* gobj, int arg1,
-                                            KirbyHatStruct* hat)
-{
-    ftKb_SpecialN_800EF040(gobj, arg1, hat);
-}
-
+/// Load Yoshi's hat for Kirby copy ability.
+/// @note The split Fighter* locals are required for register allocation.
+#pragma push
+#pragma dont_inline on
 void ftKb_SpecialN_800F10D4(Fighter_GObj* gobj)
 {
-    u8 sp14[0x90];
+    u8 part_dobj_indices[0x88];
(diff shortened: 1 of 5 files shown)


ribbanya:
It's nice that it can identify similar code, but can it extract shared logic to helper functions that still match?

MarkMcCaskey:
> It's nice that it can identify similar code, but can it extract shared logic to helper functions that still match?

Good question, I can give it a try! Any general guidelines on code structure for code sharing boundaries? Or anything reasonably complex that can be shared is good to share?

ribbanya:
I'm not sure of a good exact threshold, just anything that can reasonably be extracted to a header function would make sense. That's probably how the code was originally structured.

MarkMcCaskey:
Interesting, so the Kirby hat case got close with an inline function but couldn't figure out how to 100% match it. I can maintain 100% matches with a macro which is still a win. But with the inline function we can clean up all the pragmas disabling inlining too, just couldn't get register allocation to match that way. The function that got close had a switch at the bottom for the characters with special hat logic and all these functions became nearly empty.

ribbanya:
I would generally prefer a todo comment linking the implementations to a macro, like:
```c
/// @todo Shared code with #ftKb_SpecialN_800F11F0
```

where the `@todo` will send it to the [todo list](https://doldecomp.github.io/melee/todo.html) and the `#` will create a hyperlink to the other function.

PR #2849, opened 2026-07-15
Melee decomp 1/12: CM camera + MP

fjooord:
## Summary
Decompilation match + standards work for `CM + MP` from the melee decomp orchestrator session (July 2026).

All changes passed a standards review/repair pipeline (deterministic review_lint + adversarial LLM review vs curated maintainer-rejection exhibits, with per-file repair validated by object build + objdiff comparison) and a full-build regression gate: no previously-matched code or data regressed.

## Files (3)
- `src/melee/cm/camera.c`
- `src/melee/mp/mpisland.c`
- `src/melee/mp/mplib.c`

## Review notes
30 unresolved review finding(s) are posted as line comments below — these are the spots that need human judgment; everything else passed the automated standards gate.


src/melee/cm/camera.c
```diff
@@ -1393,18 +1397,82 @@ void Camera_8002B1F8(CameraTransformState* transform)
...
+inline void reset_bounds_unless_fighter_near(CameraBounds* bounds,
+                                             CameraTransformState* transform)
+{
+    struct {
+        u8 pad[4];
+        Vec3 pos;
+    } fighter_pos;
+    BOOL fighter_near;
+    HSD_GObj* p1_fgp;
+    f32 fighter_z;
```
fjooord: **`preship_reject_global_standard_canonical_control_flow_and_macros`** — The newly introduced single-iteration `do { ... break; } while (0)` plus comma-expression assignment is generated control-flow residue rather than normal source control flow. No local evidence is recorded to justify this noncanonical form.

**Match context:** in `reset_bounds_unless_fighter_near` (not individually scored — likely data/static) | an automated fix attempt was **reverted**: qa-repair output failed schema validation: validation must be an array of command/status rows

src/melee/mp/mplib.c
```diff
@@ -4294,8 +4321,8 @@ void mpLeftWallGetTop(int line_id, Vec3* pos_out)
 
     {
         CollVtx* vtx =
-            &groundCollVtx[((CollLine*) ((u8*) groundCollLine + line_offset))
-                               ->x0->v1_idx];
+            &groundCollVtx[groundCollLine[line_offset / sizeof(CollLine)]
```
fjooord, line 4263: **`preship_reject_global_standard_canonical_control_flow_and_macros`** — Several loop conditions replace the canonical equality test with the generated double-negative form `!(kind != (...))`. This is equivalent to equality but is not normal authored control flow and has no recorded evidence proving the canonical expression fails. The same regression appears in the left-wall, right-wall, and ceiling traversal changes.

**Match context:** in `mpLeftWallGetTop` — improvement-lane (fuzzy 98.3%); safe to change at some score cost | an automated fix attempt was **reverted**: exact function regressed: mpLib_80056C54 (100 -> 99.992); post-repair QA scan is clean for the item

src/melee/cm/camera.c
```diff
@@ -1258,9 +1262,9 @@ void Camera_8002A768(CameraTransformState* transform, s32 arg1)
         } else if (var_r31 & 2) {
             cam_correction.y = bottom_overlap;
         }
-        cam_correction.x *= -1.0f;
-        cam_correction.y *= -1.0f;
-        cam_correction.z *= -1.0f;
+        cam_correction.x *= scale = *(volatile const f32*) &cm_804D7E04;
+        cam_correction.y *= scale;
```
fjooord, line 1266: This is removal of an inline, but the value seems more correct than just -1.0f

src/melee/cm/camera.c
```diff
@@ -1393,18 +1397,82 @@ void Camera_8002B1F8(CameraTransformState* transform)
...
+        fighter_near = false;
+    } while (0);
+    if (!fighter_near) {
+        Camera_80029CF8(bounds, transform);
+        Camera_8002A768(transform, 0);
+    }
+}
+
 /// update gameplay camera
 void Camera_8002B3D4(void* arg0)
```
sadkellz, line 1422: I matched this one [here](https://decomp.me/scratch/d5xMI) without the do while inlines.

src/melee/cm/camera.c
sadkellz: `*(volatile const f32*) &cm_804D7E04`
I wouldn't accept this personally

jurrejelle: yeah this  shoudl either just be = scale = -1.0f, or replace it by the correct value if -1.0f is wrong

sadkellz: I've fixed this in my working branch [here](https://github.com/doldecomp/melee/compare/master...sadkellz:melee:camera#diff-e67a2e758f279a491137fdc7e5e605de941b43340e1f20f91d77884919d8e3d4R1291-R1294)

ribbanya (changes requested):
See comments:
- [x] [Dereferenced literal](https://github.com/doldecomp/melee/pull/2849#discussion_r3593590639)
- [x] [Cleaner inlines](https://github.com/doldecomp/melee/pull/2849#discussion_r3591086597)

src/melee/cm/camera.c
```diff
@@ -1918,52 +1974,72 @@ void Camera_8002C1A8(void)
...
+        double half = *(volatile const double*) &cm_804D7E78;
+        double three = *(volatile const double*) &cm_804D7E80;
+        guess = half * guess * (three - guess * guess * x);
+        guess = half * guess * (three - guess * guess * x);
+        guess = half * guess * (three - guess * guess * x);
+        *y = (float) (x * guess);
+        return *(volatile float*) y;
+    }
+    return x;
+}
```
ribbanya, line 1990: This is slop, why can't `sqrtf` from `math_ppc` be used?

src/melee/mp/mplib.c
```diff
@@ -5130,6 +5163,19 @@ int mpLib_80056B34(int arg0, int* arg1)
...
+    if (x > 0.0F) {
+        double guess = __frsqrte((double) x);
+        guess = 0.5 * guess * (3.0 - guess * guess * x);
+        guess = 0.5 * guess * (3.0 - guess * guess * x);
+        guess = 0.5 * guess * (3.0 - guess * guess * x);
+        *y = (float) (x * guess);
+        return *y;
+    }
+    return x;
+}
```
ribbanya, line 5116: Same here.

ribbanya (changes requested):

src/melee/cm/camera.c
fjooord: Fixed in 208a7b87f6 — now `cam_correction.x *= scale = 1.0f;`. Checked the value against the original DOL first: `cm_804D7E04` is +1.0f, not -1.0f (master's `-1.0f` actually pools into the slot at 0x804D7E10), so this also corrects the constant. A standalone `scale = 1.0f;` statement breaks the match (stack layout shifts); the comma-assign form matches at 100%.

src/melee/cm/camera.c
fjooord: Adopted your structure in 208a7b87f6 — both `do { ... } while (0)` inlines are gone, replaced with a single `static inline bool fighter_z_out_of_range(Vec3*)`, and the pad-struct hack and dead assignment are removed. Still 100%. Thanks for the scratch.

src/melee/cm/camera.c
fjooord: Done in 208a7b87f6 — plain `sqrtf` from math_ppc.h now, helper and store-slot hack deleted. Function goes from 96.80% (master) to 99.99%; the last two instructions are the second sqrtf's volatile temp landing at sp+0xC instead of sp+0x18, which I couldn't reproduce with any clean form. There's a `PAD_STACK(4)` holding that — happy to drop it if you'd rather (99.84% without).

src/melee/mp/mplib.c
fjooord: Done in 208a7b87f6 — `dist_f28 = sqrtf(x_f2 + y_f0);`, same form as master. Lands at exactly master's 99.992% for this function; the previous 100% was only reachable via the store-slot hack.

ribbanya (approved):

src/melee/mp/mplib.c
ribbanya: Revert to the 100% version, then.

ribbanya (changes requested):
https://github.com/doldecomp/melee/pull/2849#discussion_r3597011268

src/melee/cm/camera.c
ribbanya: If the hacked match is <100%, it's probably better to keep the header function. Stack padding is fine.

ribbanya (approved):

PR #2856, opened 2026-07-15
Deduplicate CPU update trio in ftCo_0A01

MarkMcCaskey:
From Claude:
> ftCo_800AEFB8, ftCo_800B1EF0, and ftCo_800B24B8 were three copies of the same 100-line CPU update body; the middle one only clears xF9_b3 and xF9_b6 where the others set them. Express all three through one shared body macro with the flag value as a parameter (a macro rather than an inline function so the compiled bytes stay identical). The compiled object is byte-identical.

### DIFF
--- a/src/melee/ft/chara/ftCommon/ftCo_0A01.c
+++ b/src/melee/ft/chara/ftCommon/ftCo_0A01.c
@@ -6772,116 +6772,119 @@ void ftCo_800AECF0(Fighter* fp)
     ftCo_800ADE48(fp);
 }
 
-/// Shared body of the three CPU update functions below (a macro rather
-/// than an inline function so the compiled bytes stay identical).
-#define FTCO_CPU_UPDATE(fp, fear_flag)                                        \
-    {                                                                         \
-        struct Fighter_x1A88_t* data2 = &fp->x1A88;                           \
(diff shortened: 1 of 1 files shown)


src/melee/ft/chara/ftCommon/ftCo_0A01.c
```diff
@@ -6772,109 +6772,116 @@ void ftCo_800AECF0(Fighter* fp)
     ftCo_800ADE48(fp);
 }
 
+/// Shared body of the three CPU update functions below (a macro rather
+/// than an inline function so the compiled bytes stay identical).
+#define FTCO_CPU_UPDATE(fp, fear_flag)                                        \
```
r-burns: Nice to deduplicate code, but the ergonomics of exposing this as a massive macro are pretty awful. Maybe ask claude to take a crack at using inlines instead?

MarkMcCaskey: Yeah completely agree. No pressure to merge any of these by the way. I'll see what I can do!

MarkMcCaskey: Nice, I was able to get this to be a function. Codex 5.6 Sol is way better at this than Claude Fable.

ribbanya (approved):

PR #2858, opened 2026-07-16
Introduce SelectableCharacterKind enum and rename selkind accessors, name some character unlock methods.

Jacoby6000:
This PR names methods related to unlocking characters, and introduces a SelectableCharacterKind enum.  I use `selkind` throughout as a counterpoint to `ckind`.

The new enum was introduced because there is a mapping that exists from CharacterKind -> SelectableCharacterKind, where SelectableCharacterKind has only 25 elements.  This makes an awkward second set of "CharacterKind" with values all offset by 1 after 0x13 (Shiek), and without as many elements as the CharacterKind enum. SelectableCharacerKind does not have all unplayable characters like its counterpart.  

Both CharacterKind and SelectableCharacterKind have arrays that can be used to map between them, and you'll see them in this PR as selkind_to_ckind_map and ckind_to_selkind_map.

The character unlock list and fighter data diagrams both rely on the selectable character kind; not distinguishing between Shiek and Zelda.  There may be others as well, but all I have found so far is in this PR.


src/melee/ft/forward.h
```diff
@@ -160,7 +160,7 @@ typedef enum CharacterKind {
     /* 1F */ CHKIND_SANDBAG,                       // Sandbag
     /* 20 */ CHKIND_POPO,                          // Popo
     /* 21 */ CHKIND_NONE,                          // None
-    /* 22 */ CHKIND_MAX = CHKIND_NONE
+    CHKIND_MAX = CHKIND_NONE
```
ribbanya:
```suggestion
    /* 21 */ CHKIND_MAX = CHKIND_NONE
```

src/melee/gm/gm_1601.c
```diff
@@ -2372,45 +2378,47 @@ int gm_801647F8(u8 arg0)
...
+    if (unlock_bit == NUM_UNLOCKABLE_CHARACTERS ||
+        (*unlocked_chars_bitmask & (1LL << unlock_bit)))
+    {
         return true;
     }
     return false;
 }
 
-static inline bool gm_80164840_inner(u8 ckind)
+static inline bool is_ckind_unlocked_inner(u8 ckind)
```
ribbanya: Camel case is preferred for static functions. You also shouldn't specify `inline`. If the "inner" is just placeholder for figuring out a match, you can treat it as a suffix, but up to you.
```suggestion
static bool isCKindUnlocked_inner(u8 ckind)
```

src/melee/gm/gm_1601.c
```diff
@@ -2422,11 +2430,12 @@ void gm_80164910(int arg0)
     }
 }
 
-void gm_80164A0C(u8 arg0)
+void gm_80164A0C(u8 ckind) ///< Lock character?
```
ribbanya: If you must put a doc comment at the definition, please use the independent line syntax.
```suggestion
/// Lock character?
void gm_80164A0C(u8 ckind)
```

src/melee/gm/gm_1601.static.h
```diff
@@ -95,87 +95,121 @@ static const u8 lbl_803B7844[] = {
...
+    /* 1A */ SELKIND_MARIO,
+    /* 1B */ SELKIND_MARIO,
+    /* 1C */ SELKIND_MARIO,
+    /* 1D */ SELKIND_COUNT,
+    /* 1E */ SELKIND_MARIO,
+    /* 1F */ SELKIND_MARIO,
+    /* 20 */ SELKIND_CAPTAIN,
+    /* 21 */ 0x00, ///< Likely extra unncessary bytes
+    0x00,
+    0x00
```
ribbanya: Padding to the nearest 4-byte alignment will be filled by the compiler automatically.
```suggestion
```

src/melee/vi/vi1101.c
```diff
@@ -6,6 +6,9 @@
 #include "dolphin/gx/GXStruct.h"
 #include "ef/efasync.h"
 #include "ef/eflib.h"
+
+#include "ft/forward.h"
```
ribbanya: You don't actually need to include `forward.h` files in `.c` files. They are the equivalent of `types.h` but intended to be used by `.h` files, to avoid circular inclusions.

ribbanya (changes requested):
Very useful changes, thank you.

src/melee/gm/gm_1601.c
Jacoby6000: I went ahead and added the comment in the style you suggested. In the doxygen cleanup PR this will be further refined

ribbanya (dismissed):

PR #2861, opened 2026-07-16
Deduplicate Game & Watch item spawn setup

MarkMcCaskey:


src/melee/it/items/inlines.h
ribbanya, line 1: Move to `src/melee/it/items/inlines.h`. The reason we do this is because the original header (or source file) of an inline can't be verified except through recognizing duplicate code, so we place them in their own headers for now. For all we currently know, this could be a real function somewhere in `melee/ft/ftGameWatch` for example.

ribbanya (changes requested):

ribbanya (approved):

PR #2863, opened 2026-07-16
Use shared stage JObj setup helper

MarkMcCaskey:
From Codex:
> Replace the remaining 35 exact setup bodies with Ground_JObjInline1, completing the 55-function cluster. This also removes the manual PAD_STACK(8) matching shapes where the inline now provides the original stack behavior.
>
> All 35 changed functions remain 100% matching.


ribbanya (approved):
Awesome.

PR #2864, opened 2026-07-16
Deduplicate target stage initialization

MarkMcCaskey:
From Codex:
> Share the common 21-function target-stage initialization sequence through Ground_InitTargetStage. Normalize eight private constructor index parameters from s32 to int so their function-pointer types match the majority convention without casts.
>
> All 21 initializers and all eight retagged constructors remain 100% matching.


ribbanya (approved):
Really cool that it identified an inlined callback like that.

PR #2865, opened 2026-07-16
Share fighter & item hitlag callback setup

MarkMcCaskey:
Small win but I can see this potentially being useful as a split out helper.

From Codex:
> Replace 24 repeated effect hitlag callback setup blocks across 15 fighter translation units with Fighter_SetEffectHitlagCallbacks. Effect spawning and move-specific control flow remain in each caller.
> 
> Keep ftLuigi_SpecialS_Setup explicit because it is itself inlined into launch and misfire entry functions; nesting this helper grows ftLg_SpecialSLaunch_Enter's stack frame by eight bytes.
> 
> All 16 audited objects retain identical code and data to master, and no translation unit emits an out-of-line helper copy.


ribbanya:
> Keep ftLuigi_SpecialS_Setup explicit because it is itself inlined into launch and misfire entry functions; nesting this helper grows ftLg_SpecialSLaunch_Enter's stack frame by eight bytes.

This could be a function defined in a header file, but because Luigi's is the first call site in link order, the header function is emitted to that TU. We see similar phenomena with, for example, common walk functions being emitted to some of Donkey Kong's TUs.

src/melee/it/inlines.h
```diff
@@ -19,6 +22,21 @@ static inline void itResetVelocity(Item* ip)
...
+    ip->exited_hitlag = efLib_ResumeAll;
+}
+
+static inline void Item_EnterStateWithEffectHitlag(Item_GObj* gobj, s32 msid)
+{
+    Item* ip = GET_ITEM(gobj);
+    it_802762BC(ip);
+    Item_80268E5C(gobj, msid, ITEM_ANIM_UPDATE);
+    ip->entered_hitlag = efLib_PauseAll;
+    ip->exited_hitlag = efLib_ResumeAll;
```
ribbanya, line 39: Does calling `Item_SetEffectHitlagCallbacks` disrupt matches due to call depth?

ribbanya (changes requested):

src/melee/it/inlines.h
MarkMcCaskey: Yeah it increases the stack frame size in many / most of the callers breaking the exact match

ribbanya (dismissed):

PR #2866, opened 2026-07-16
Share result stat clamp logic

MarkMcCaskey:
From Codex:
> Replace all 24 duplicated +/-999999 result-stat clamps with a small inline helper. Direct field calls in gm_17BA preserve its original stack frame, while computed and converted values use the same helper without changing generated code.


src/melee/gm/gm_17AD.h
```diff
@@ -3,6 +3,16 @@
...
+static inline s32 gmClampResultStat(s32 value)
+{
+    if (value > 999999) {
+        value = 999999;
+    } else if (value < -999999) {
+        value = -999999;
+    }
+    return value;
+}
+
```
ribbanya: Move to `src/melee/gm/inlines.h`.

ribbanya (changes requested):

ribbanya (approved):

PR #2869, opened 2026-07-16
Reuse emitted ftlib field accessors

MarkMcCaskey:
From Codex:
> Reuse load-only accessors for held items, camera subjects, motion IDs, fighter flags, and fighter kinds within ftlib.c. MWCC inlines these same-TU calls while preserving each emitted accessor body and all loadable sections exactly.
>
> Keep the collision-data access in ftLib_8008777C raw because the accessor adds an inline stack band. Also leave larger predicates and callers that retain their Fighter pointer raw because those forms change register allocation


src/melee/ft/ftlib_inline.h
ribbanya, line 1: Delete and merge contents into `src/melee/ft/inlines.h`.

src/melee/ft/ftlib.c
```diff
@@ -832,16 +826,13 @@ s32 ftLib_8008730C(HSD_GObj* gobj)
...
+    return ftLib_8008701C(gobj);
 }
 
 bool ftLib_8008732C(HSD_GObj* gobj)
 {
-    Fighter* fp = GET_FIGHTER(gobj);
-    if (fp->motion_id >= ftCo_MS_DeadDown &&
-        fp->motion_id <= ftCo_MS_DeadUpFallHitCameraIce)
-    {
+    FtMotionId msid = ftLib_80086C0C(gobj);
```
ribbanya: Let's rename `ftLib_80086C0C` to `ftLib_GetMotionId`.

ribbanya (changes requested):

src/melee/ft/ftlib_inline.h
MarkMcCaskey: Looks like I need a guard around just that function because some of the things in that header need to be included multiple times to maintain matching due to transitive imports 🤔 . But looks like there's some precedent for that

ribbanya (approved):

PR #2870, opened 2026-07-16
Reuse emitted gmmain accessors

MarkMcCaskey:
From Codex:
> Use gmMainLib_8015CC40 and GetPersistentFighterData throughout the compact save-data accessor families. All 76 changed functions remain exact matches.
>
> Keep the direct field access in gmMainLib_8015D00C, gmMainLib_8015D134, gmMainLib_8015D25C, and gmMainLib_8015D384 because routing their PAD_STACK-shaped setters through GetPersistentFighterData changes codegen. gmMainLib_8015CEFC likewise retains its direct 64-bit flag access because the base accessor changes its instruction shape.

PR #2871, opened 2026-07-16
Reuse emitted gm_1601 state accessors

MarkMcCaskey:
From Codex:
> Route 18 matched state readers and writers through fn_80169364 and gm_8016A98C. Every changed function remains an exact match.
>
> Keep fn_8016A09C on direct field access because caching fn_80169364 changes register allocation. Keep gm_8016A92C and gm_8016A97C direct because they precede gm_8016A98C's definition and therefore emit real calls. The earlier large fn_80169364 caller also remains unchanged because its call exists in the target.


MarkMcCaskey:
Moving this to one big PR for this GM stuff on #2870

PR #2874, opened 2026-07-16
Name and reuse stage cache accessors

MarkMcCaskey:
From Codex:
> Give the emitted stage archive and preload cache scene accessors names that describe the data they expose, and update their callers.
>
> Route all later preloadCache.scene accesses through lbDvd_GetPreloadCacheScene. The resulting lbdvd object remains byte-identical in every loadable section.
>
> Keep grDatFiles_801C6330 on direct archive-array access: routing it through grDatFiles_GetArchive changes register allocation and lowers the function match to 88.17073%.


src/melee/lb/lbdvd.c
```diff
@@ -800,7 +800,7 @@ void lbDvd_80018F68(void)
 {
     int i;
     preloadCache.persistent_heaps = 0;
-    preloadCache.scene = lbDvd_803BA638;
+    *lbDvd_GetPreloadCacheScene() = lbDvd_803BA638;
```
ribbanya: Let's rename `lbDvd_803BA638` to `preload_cache_scene`, since it's `static`.

ribbanya (changes requested):

ribbanya (approved):

PR #2875, opened 2026-07-16
Name and reuse core item accessors

MarkMcCaskey:
From Codex:
> Give nine emitted item accessors names matching the fields they expose: heavy status, hold kind, damage multiplier, grab ranges, motion ID, team ID, owner, and attack ID.
>
> Provide opt-in inline definitions for the motion ID and owner getters. Use them at 16 small motion-state sites and two duplicate owner wrappers; the affected item text sections remain byte-identical.
>
> Keep itMSBomb_Logic19_DmgReceived on direct motion-ID access. Its explicit stack pad makes the inline accessor change a handful of instructions.


src/melee/it/items/itzeldadinfire.c
```diff
@@ -46,7 +47,7 @@ typedef struct {
 
 Fighter_GObj* it_802C3AF0(Item_GObj* gobj)
```
ribbanya: Rename `it_802C3AF0` to `itZeldaDinFire_GetOwner`. It's used by `ftZd_SpecialS` so it needs a qualified name.

src/melee/it/it_26B1_inline.h
ribbanya, line 1: Delete and move contents to `src/melee/it/inlines.h`.

ribbanya (changes requested):

src/melee/it/inlines.h
```diff
@@ -14,6 +14,20 @@ static inline Item* GetItemData(HSD_GObj* gobj)
     return item_data;
 }
 
+#ifndef IT_26B1_SOURCE
```
ribbanya: This is not correct. Just use the declaration from `it_26B1` in other files. It doesn't need the inline keyword to be inlined.

ribbanya (changes requested):
If the original location of a function is known, we don't need to put it in an `inlines.h`, we just use it as-is.

src/melee/it/inlines.h
MarkMcCaskey: Oh yeah, good point

MarkMcCaskey: It breaks some uses by emitting a call but that's likely an indication that those are just direct accesses or something else not using the function directly, so I'll remove those

ribbanya: Could you put a `/// @todo` comment on those instances? It might be a call depth issue that can be worked out.

ribbanya (approved):

PR #2878, opened 2026-07-16
Melee decomp 6/11: LB support libraries

fjooord:
## Summary

Exact-match decompilation of 18 file(s) (30 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Verification

- Published from the local branch commit `2d2a6acebb`.


src/melee/lb/lbspdisplay.c
```diff
@@ -1321,6 +1310,26 @@ static void* setImageFromPreloadedArchive(HSD_ImageDesc* image_desc,
     return image_ptr;
 }
 
+struct BlurChannels {
```
fjooord: **`preship_reject_global_standard_literals_and_data_ownership`**

> The diff promotes the ordinary strings `"lbspdisplay.c"` and `"!image_desc->image_ptr"` into named static storage and places channel objects beside them, without section or symbol ownership evidence. Their later reinterpretation as one aggregate shows that these declarations are being used as a data-layout anchor rather than truthful source ownership.

**Match context:**

- match status unavailable

src/melee/lb/lbspdisplay.h
```diff
@@ -75,7 +75,7 @@ lb_800122C8(HSD_ImageDesc* image_desc, u16 origx, u16 origy,
                               struct lb_80014638_arg1_t*);
 /* 014770 */ bool lb_80014770(Vec3*, int);
 /* 0149E0 */ bool lb_800149E0(Mtx, u32);
-/* 4D3760 */ extern GXColor lb_804D3760;
-/* 4D3764 */ extern GXColor lb_804D3764;
+/* 4D3760 */ extern GXColor lb_ColorWhite;
+/* 4D3764 */ extern GXColor lb_ColorBlack;
```
fjooord, line 79: **`preship_reject_global_standard_no_define_alias_global_renames`**

> This changes canonical global symbol names without recorded source, symbol, map, or review evidence. Although the rename is direct rather than hidden behind a define alias, this standard permits direct global renames only when they are evidence-backed; no such evidence appears in the slice.

**Match context:**

- match status unavailable

src/sysdolphin/baselib/debug.h
```diff
@@ -23,6 +23,16 @@ ATTRIBUTE_NORETURN void HSD_Panic(char*, u32, char*);
 #define HSD_ASSERTREPORT(line, cond, ...)                                     \
     ((cond) ? (void) 0                                                        \
             : (OSReport(__VA_ARGS__), __assert(__FILE__, line, #cond)))
+/// HSD_ASSERTREPORT for code split out of a larger retail TU: the retail
+/// __FILE__ string and stringized condition belong to the original TU's
```
fjooord: **`preship_reject_global_standard_assert_report_macros`**

> The new HSD_ASSERTREPORTFILE macro is a fake assert/report variant that exposes the filename and stringized condition as caller-supplied data solely to preserve strings from a former merged TU. Its body manually expands OSReport plus __assert instead of retaining the established HSD_ASSERTREPORT idiom. The adjacent comment explicitly identifies split-TU data placement as the motivation, making this a data-ordering workaround rather than an authored project macro. The standard has no exception for replacing an existing macro form with a custom raw expansion.

Also at lines: 30, 33, 37

**Match context:**

- match status unavailable

src/melee/lb/lbaudio_ax.c
```diff
@@ -2537,16 +2529,36 @@ s32 lbAudioAx_8002785C(void)
...
+} PoolAllocFull;
+
+typedef struct AudioDataFull {
+    u8 x0[0xA00];
+    char* xA00;
+    u8 xA04[0xC4];
+    char* xAC8;
+    u8 xACC[8];
+    char* xAD4;
+} AudioDataFull;
```
fjooord: **`preship_reject_global_standard_header_inlines`**

> Calls to the existing `lbAudioAx_SetAudioPath` helper are replaced with open-coded `strcpy` operations using offsets from `lbl_803BB300`. This discards an established source-level helper boundary and exposes data-layout arithmetic in its place. The same regression is repeated in `lbAudioAx_80028690`.

**Match context:**

- match status unavailable

src/melee/lb/lbaudio_ax.c
```diff
@@ -2556,18 +2568,19 @@ void lbAudioAx_80027AB0(s32 arg0)
     if (lbl_804D3878 != lang) {
         lbl_804D3878 = lang;
         AXDriver_8038DCFC();
-        strcpy(&lbl_803BB340[lbl_804D38D0], "smash2.sem");
-        AXDriver_8038DA70(lbl_803BB340, lb_800195D0);
+        path = base + lbl_804D38D0;
```
fjooord: **`preship_reject_global_standard_no_string_literal_symbol_regression`**

> The inline asset name `"smash2.sem"` is replaced by `base + 0x17A8`, an address-derived reference into merged data. This is precisely a string-literal-to-data-symbol regression for code matching, with no maintainer-approved data-work exception. The same replacement is repeated near line 2958.

Also at line: 2820

**Match context:**

- match status unavailable

src/melee/lb/lbaudio_ax.c
```diff
@@ -2662,12 +2701,8 @@ void lbAudioAx_80027DF8(void)
             carry = 0;
             lbl_804D6428 = carry;
             lbl_804D38F0 = -1;
-            if (lbl_804D38F4 == -1) {
-                lbl_804D642C = carry;
-            } else {
-                carry = 0x7F;
-                lbl_804D642C = carry;
-            }
+            carry = lbl_804D38F4 == -1 ? carry : 0x7F;
```
fjooord, line 2645: **`numeric_literal_to_symbol`**

> Numeric literal replaced by address-style data symbol `lbl_804D38F4`. Keep float/constants inline unless the PR is explicitly scoped to evidenced data ownership. Keep literals inline unless data ownership evidence says otherwise.

Also at line: 2735

**Match context:**

- match status unavailable

fjooord, line 2645: **`preship_reject_global_standard_literals_and_data_ownership`**

> Lint deterministically identified `carry = lbl_804D38F4 == -1 ? carry : 0x7F;` as a numeric-literal-to-address-symbol regression. Ordinary logic now depends on the address-style data symbol without recorded section-ownership evidence explicitly scoping this as data work.

Also at line: 2735

**Match context:**

- match status unavailable

fjooord:
<details>

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/ty/tylist.c:527`

> The diff preserves newly edited identical `if`/`else` assignments to `state_tail->x0C` and `state_tail->x0D`. This is a generated branch ladder with no differing result and no recorded local evidence justifying the duplicated control flow.

Evidence:

> Collapsing only the duplicated x278/x270/x274 assignments into one unconditional block changed exact _tyList_80313774 from 100% to 94.83274%. The exact-match source was restored as required.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/ty/tylist.c:561`

> A second identical `Toy_GetTrophyTotal() <= 0xA` branch assigns the same three entry pointers in both arms. This is generated duplicated control flow, not a meaningful authored conditional.

Evidence:

> Collapsing only the duplicated x278/x270/x274 assignments into one unconditional block changed exact _tyList_80313774 from 100% to 94.83274%. The exact-match source was restored as required.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/ty/tylist.c:618`

> The cursor creation conditional has byte-for-byte equivalent calls in both branches. This is another generated branch with no semantic selection and no permitted evidence-backed exception.

Evidence:

> Collapsing only the duplicated x278/x270/x274 assignments into one unconditional block changed exact _tyList_80313774 from 100% to 94.83274%. The exact-match source was restored as required.



</details>

src/melee/lb/lbspdisplay.h
fjooord: will fix this

src/melee/lb/lbaudio_ax.c
ribbanya: Not true because `lbl_804D38F4` is in `sdata` (mutable).

ribbanya: Not true because `lbl_804D38F4` is in `sdata` (mutable).

src/melee/lb/lbaudio_ax.c
```diff
@@ -2730,97 +2761,99 @@ void lbAudioAx_80027DF8(void)
...
-    AXDriver_8038E30C(0, 2, &rvbStd, (u8*) ((char*) lbl_80433B44 + 0x120),
-                      0xD400);
+    AXDriver_8038E30C(0, 2, &rvbStd, st->x554, 0xD400);
 
     AXDriver_8038E37C(AXDRIVER_AUX_DELAY, &delay);
-    HSD_ASSERTMSG(0xF72, HSD_AudioGetAuxHeapSize(2, &delay) < 71 * 1024,
-                  "HSD_AudioGetAuxHeapSize(2, &delay) < 71*1024");
+    HSD_ASSERTMSGFILE(base + 0x1780, 0xF72,
+                      HSD_AudioGetAuxHeapSize(2, &delay) < 71 * 1024,
+                      base + 0x17E4);
```
ribbanya: Revert and fix the data section instead.

src/melee/lb/lbaudio_ax.c
```diff
@@ -2904,68 +2951,65 @@ s32 lbAudioAx_80028690(void)
...
             lbl_80433984[0x36] = 2;
         }
 
         if (lbl_80433984[2] < 1) {
-            strcpy(&lbl_803BB340[lbl_804D38D0],
-                   *(char**) (lbl_803BB300 + 0xA04));
-            lbl_80433A64[2] = HSD_SynthSFXLoad(lbl_803BB340, 2, 0, 0);
+            path = base + lbl_804D38D0;
+            strcpy(path + 0x40, data->xA04);
+            lbl_80433A64[2] = HSD_SynthSFXLoad(base + 0x40, 2, 0, 0);
```
ribbanya: Revert and fix the data section.

src/melee/lb/lbspdisplay.h
ribbanya: These renames are fine. They're shared by multiple TUs and the names are based on value, which is clearly demonstrated by the data (`0xFFFFFFFF` and `0x000000FF`).

src/sysdolphin/baselib/debug.h
ribbanya: True. Revert all of this garbage.

src/melee/lb/lbspdisplay.c
```diff
@@ -881,15 +867,15 @@ void lb_8001044C(DynamicsDesc* desc, void* colliders_raw, int num_colliders,
...
                 HSD_JObjSetRotationZ(jobj, rot_z);
             } else if (axis == 0) {
                 f32 rot_x = HSD_JObjGetRotationX(jobj);
-                rot_x *= 0.9f;
+                rot_x = rot_x * 0.9f;
                 HSD_JObjSetRotationX(jobj, rot_x);
             } else {
                 f32 rot_y = HSD_JObjGetRotationY(jobj);
-                rot_y *= 0.9f;
+                rot_y = rot_y * 0.9f;
```
ribbanya, line 878: Do these changes affect codegen...?

src/melee/lb/lbspdisplay.c
ribbanya: True.

src/melee/lb/lbspdisplay.c
```diff
@@ -1487,11 +1496,17 @@ void lb_8001285C(HSD_ImageDesc* image_desc, GXTexObj* tex_obj)
                    GX_LO_CLEAR);
 }
 
+static inline void consume_color(GXColor color) {}
+
+static inline void consume_blur_colors(GXColor color0, GXColor color1,
+                                       GXColor color2, GXColor color3)
+{
+}
```
ribbanya, line 1486: Add todo comments to these obviously-fake functions.

src/melee/lb/lbspdisplay.c
```diff
@@ -1500,200 +1515,220 @@ void lb_80012994(HSD_ImageDesc* img, u8 alpha, u8 blur_size, f32 x, f32 y,
     f32 off2 = 2.0f * off1;
 
     lb_800122F0(img, &tex, color_factor);
-    PAD_STACK(88);
+    PAD_STACK(8);
 
-    color.a = alpha;
-    GXSetTevColor(GX_TEVREG0, color);
+#define color_slot(i) (((GXColor*) &tex)[i])
```
ribbanya: Expand and delete this.

src/melee/lb/lbspdisplay.c
```diff
@@ -2162,6 +2199,14 @@ bool lb_80014638(struct lb_80014638_arg0_t* arg0,
     return true;
 }
 
+/// Initializers evidenced by retail .sdata bytes at 804D3760..804D3773
+/// (FFFFFFFF 000000FF FF0000FF FF000040 FFFF00FF).
```
ribbanya: Remove unnecessary comment.

ribbanya (changes requested):

src/melee/lb/lbaudio_ax.c
fjooord: Agreed — the lint's premise was wrong here: `lbl_804D38F4` (and `lbl_804D38F0` at 2735) are real mutable `.sdata` variables initialized to `-1`, not constants promoted to address symbols, so the reads are truthful. No change made; we'll teach the `numeric_literal_to_symbol` rule to skip initialized mutable data.

src/melee/lb/lbaudio_ax.c
fjooord: Reverted to master's form (plain `HSD_ASSERT`/`HSD_ASSERTMSG`, named arrays, no overlay structs); the `…MSGFILE` macros are gone entirely (see the debug.h thread). The data section didn't need a new split — this TU's `.data` is already fully named and C-defined, and the retail string pool (`__FILE__` at +0x1780, the assert messages at +0x17B4/+0x17E4) is emitted in retail order by the existing `order_data` anchor, so `.data` stays 100% matched. Cost of dropping the tactic: this function goes from a true 100% back to master's 97.58% (retail addresses everything off one `.data` base register, which the clean form doesn't reproduce).

src/melee/lb/lbaudio_ax.c
fjooord: Reverted to master's form. One caveat on 'fix the data section': the section is already fully split/matched — the `.ssm` filenames live in the named `lbl_803BBCFC` array and the path prefixes in the pool blob emitted by `order_data`. An inline `"/audio/us/"` literal can't work (retail overlays it with `"/audio/"` inside one pooled blob), so master's `lbAudioAx_SetAudioPath` char-store form is restored. Cost: 99.59% → master's 81.37% on this function; `.data` stays 100%.

src/melee/lb/lbspdisplay.h
fjooord: Thanks — keeping `lb_ColorWhite`/`lb_ColorBlack` (and the local `red`/`translucent_red`/`yellow`), matching the 0xFFFFFFFF/0x000000FF sdata values.

src/sysdolphin/baselib/debug.h
fjooord: Done — debug.h is back to master byte-for-byte; the two lbaudio_ax call sites are plain `HSD_ASSERT`/`HSD_ASSERTMSG` again, with the retail strings supplied by the TU's `order_data` pool anchor.

src/melee/lb/lbspdisplay.c
fjooord: Yes — measured: with `rot_x *= 0.9f` lb_8001044C drops 87.349% → 87.322% (MWCC schedules the temporary differently), so the expanded form is load-bearing and kept.

src/melee/lb/lbspdisplay.c
fjooord: Fixed — the promoted strings and the aggregate reinterpretation are gone; master's `lb_803BA1C0` channel struct is back, referenced through an ordinary pointer local. fn_80013614 is 99.94% (was a tactic-100%); the file/assert strings come from the `HSD_ASSERT` literals again.

src/melee/lb/lbspdisplay.c
fjooord: Done.

src/melee/lb/lbspdisplay.c
fjooord: Expanded and deleted; byte-identical (lb_80012994 still 100%).

src/melee/lb/lbspdisplay.c
fjooord: Removed.

src/melee/lb/lbaudio_ax.c
fjooord: Same as the sibling thread — the flagged reads are of a real mutable `.sdata` variable, so no change was needed.

src/melee/lb/lbaudio_ax.c
fjooord: Done, and applied class-wide rather than just at the two commented sites. Every branch-introduced instance of this tactic family in the lb slice — base+literal-offset reads into pooled strings/data, overlay typedefs cast over merged regions, and promoted pool strings — is now reverted to master's exact forms (6e27f1f9fe):

- `lbAudioAx_8002838C` and `lbAudioAx_80028690` (the commented sites): canonical `HSD_ASSERT`/`HSD_ASSERTMSG`, inline `"smash2.sem"`, named arrays; the `HSD_ASSERTMSGFILE` macros are deleted from debug.h entirely.
- `lbAudioAx_80027AB0`: dropped the `PoolAllocFull`/`AudioDataFull` overlays and the `base+0x1790/0x179C/0x17A8` pool-string reads (back to `lbAudioAx_SetAudioPath` + literal).
- `lbAudioAx_80023B24`: dropped the `lbAudioAx_BankName` overlay and `base+0x40` path math.
- `fn_80026C04`: dropped the `fn_80026C04_Data` overlay and accessor.

Base-pointer arithmetic that master itself already uses (`fn_80023254`, `lbAudioAx_800233EC`, `lbAudioAx_80027168`) is unchanged. No data-section fix was actually needed: this TU's `.data` is fully named/C-defined and the retail string pool is emitted in retail order by the existing `order_data` anchor, so every data section stays 100% matched throughout. Match cost of dropping the tactics (retail keeps a single `.data` base register live, which the clean forms can't reproduce): 8002838C 100→97.6, 80028690 99.6→81.4, 80027AB0 100→79.6, 80023B24 99.2→91.6, 80026C04 96.4→94.5 — all at or above their upstream-master values.

src/melee/lb/lbaudio_ax.c
```diff
@@ -2769,91 +2737,76 @@ void lbAudioAx_8002835C(void)
...
     PAD_STACK(8);
 
-    lbl_804D643C = data->offsets[0][0];
-    lbl_804D6440 = data->offsets[0x33][0];
-    lbl_804D6440 += data->offsets[1][0];
-    lbl_804D6440 += data->offsets[0x36][0];
+    lbl_804D643C = *(int*) (lbl_803BB300 + 0x11E4);
+    lbl_804D6440 = *(int*) (lbl_803BB300 + 0x137C);
+    lbl_804D6440 += *(int*) (lbl_803BB300 + 0x11EC);
+    lbl_804D6440 += *(int*) (lbl_803BB300 + 0x1394);
```
ribbanya: Use `M2C_FIELD` instead of direct pointer manipulation. Preferably modify the datatype for clean field access. Same goes for below this point.

ribbanya:
Could you add a lint rule to discourage defining function-like macros? It's almost always wrong, and in the cases where it's correct, a todo-comment requesting a human do it would be preferred.

ribbanya (changes requested):

src/melee/lb/lbaudio_ax.c
ribbanya: True, revert.

ribbanya (changes requested):

fjooord:
Will do — queued as a new deterministic rule in our standards set (`function_like_macro`: discourage defining function-like macros; where one is genuinely correct, require a @todo comment asking for human review). It'll post on future PRs the same way the other lint findings do. The instances already on this branch (`color_slot`, and grgreens' `BLOCK`/`VEC` on #2880) were expanded and deleted in the earlier rounds.

src/melee/lb/lbaudio_ax.c
fjooord: Done — and applied through the whole region below as well (aea554361d). Every raw `*(int*)(lbl_803BB300 + off)` / `*(char**)(…)` read is now clean access on the real objects: `offsets_arr_803BC4E4[…]`, `lbl_803BBCFC[slot]`, `s32_arr_803BB5D0[slot][1]`, `lbl_80433B44[…]`. For the path-prefix strings we modified the datatype: the fake combined pool literal is replaced by real `str_audio_us`/`str_audio` objects (named in symbols.txt by value), so the `strcpy` calls take named strings. `M2C_FIELD` turned out to be unnecessary — the typed forms hold or improve the match at every site (unit .text 97.53% → 99.40%, e.g. lbAudioAx_80027AB0 79.6% → 99.2%, lbAudioAx_80023B24 91.6% → 98.9%), all data sections stay 100% matched, and no other TU moves. The remaining ~1–2% on a few functions is retail addressing everything off a single `.data` base register, which we can't reproduce without the fake region-overlay typedefs.

src/melee/lb/lbaudio_ax.c
fjooord: Agreed — this was reverted in the previous pass (6e27f1f9fe): both sites use `strcpy(&lbl_803BB340[lbl_804D38D0], "smash2.sem")` with the inline literal again. The score the tactic had bought is now mostly recovered through the typed shapes from the other thread (lbAudioAx_80027AB0 79.6% → 99.2%, lbAudioAx_80028690 81.4% → 97.9%).

ribbanya (approved):

src/melee/lb/lbaudio_ax.c
fjooord: Follow-up so this thread reflects the final state: after the class-wide sweep, the subsequent typed-access round (aea554361d, per the discussion at line 2760) recovered most of the cost with clean shapes on the real datatypes — this function is now at 98.15% (sweep had it at master's 97.58%), `lbAudioAx_80027AB0` 99.15%, `80023B24` 98.92%, `80028690` 97.87%. All data sections still 100% matched.

src/melee/lb/lbaudio_ax.c
fjooord: Follow-up for the final state: with the typed shapes from the line-2760 thread (aea554361d), this function recovered from master's 81.37% to 97.87% — no offset math, no overlays; the path-prefix strings are real named objects (`str_audio_us`/`str_audio`) and `.data` stays 100% matched.

PR #2879, opened 2026-07-16
Melee decomp 7/11: TY trophies

fjooord:
## Summary

Exact-match decompilation of 5 file(s) (26 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Verification

- Published from the local branch commit `26ea87a9e0`.


src/melee/ty/tyfigupon.c
```diff
@@ -1366,50 +1373,43 @@ static void order_data_4A8(void)
 
 void _tyFigupon_80317A60(void)
 {
```
fjooord, line 1375: **`preship_reject_global_standard_no_string_literal_symbol_regression`**

> The local `strings` is initialized with the ordinary asset literal `"ToyFigurePonPanel_Top_joint"` and later treated as the base of unrelated data via `strings + 0x4EC`. This is a hand-packed string-blob/offset-access dodge matching the maintainer-rejected packed-string pattern; it also points far beyond the declared literal object.

**Match context:**

- match status unavailable

fjooord:
**`preship_reject_global_standard_assert_report_macros`**

> The OSPanic message is changed from `"0"` to an empty string. Panic/report text is observable diagnostic behavior, and the assert/report standard forbids dropping real report or panic information merely to alter object output.

**Match context:**

- match status unavailable

fjooord:
<details>

**`extern_in_c`** at `src/melee/ty/tydisplay.c:84`

> Added `extern` declaration `_tyDisplay_803B8988` in a .c file. Externs in .c files are not allowed; declare the symbol in the owning header or define the data in the TU that owns it. This TU owns the data (0x803B8988 is inside its own section ranges in splits.txt); define it here in binary order instead of externing it.

Evidence:

> _tyDisplay_803B8A34 is the second 0xAC portion of the same owned .rodata table. Defining the full table cleanly caused the measured exact-function regressions and keeping both generated lookup data and the definition duplicated the section, so the target was restored.

**`preship_reject_global_standard_literals_and_data_ownership`** at `src/melee/ty/tydisplay.c:84`

> The added `extern TyDspNameTables _tyDisplay_803B8988;` is an extern declaration in a .c file, which is banned outright. Lint additionally establishes that address 0x803B8988 is owned by this translation unit, so this is not a legitimate cross-TU extern or the accepted forward-declaration case with a later same-file definition.

Evidence:

> The source-owned definition experiment established correct section ownership and binary order, but it either duplicated .rodata or caused three existing exact matches to collapse. No header was edited and the prior source was restored under the checkpoint rule.

**`extern_in_c`** at `src/melee/ty/tydisplay.c:85`

> Added `extern` declaration `_tyDisplay_803B8A34` in a .c file. Externs in .c files are not allowed; declare the symbol in the owning header or define the data in the TU that owns it. This TU owns the data (0x803B8A34 is inside its own section ranges in splits.txt); define it here in binary order instead of externing it.

Evidence:

> _tyDisplay_803B8A34 is the second 0xAC portion of the same owned .rodata table. Defining the full table cleanly caused the measured exact-function regressions and keeping both generated lookup data and the definition duplicated the section, so the target was restored.

**`preship_reject_global_standard_literals_and_data_ownership`** at `src/melee/ty/tydisplay.c:85`

> The added `extern TyDspArchNames _tyDisplay_803B8A34;` is an extern declaration in a .c file. Lint establishes that 0x803B8A34 lies in this translation unit's own section ranges, making this a self-TU ownership/data-ordering dodge rather than a legitimate external dependency.

Evidence:

> The source-owned definition experiment established correct section ownership and binary order, but it either duplicated .rodata or caused three existing exact matches to collapse. No header was edited and the prior source was restored under the checkpoint rule.

**`preship_reject_global_standard_literals_and_data_ownership`** at `src/melee/ty/tydisplay.c:1496`

> The explicitly labeled `.data order hack`, `(void) "ScMenDisplay_cam_int1_camera";`, emits an otherwise unused string solely to influence data ordering. This is invented storage/data-order anchoring rather than authored behavior and is exactly the kind of score-motivated data-ordering dodge that may not ship.

Evidence:

> The source-owned definition experiment established correct section ownership and binary order, but it either duplicated .rodata or caused three existing exact matches to collapse. No header was edited and the prior source was restored under the checkpoint rule.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/ty/tylist.c:527`

> The diff preserves newly edited identical `if`/`else` assignments to `state_tail->x0C` and `state_tail->x0D`. This is a generated branch ladder with no differing result and no recorded local evidence justifying the duplicated control flow.

Evidence:

> Collapsing only the duplicated x278/x270/x274 assignments into one unconditional block changed exact _tyList_80313774 from 100% to 94.83274%. The exact-match source was restored as required.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/ty/tylist.c:561`

> A second identical `Toy_GetTrophyTotal() <= 0xA` branch assigns the same three entry pointers in both arms. This is generated duplicated control flow, not a meaningful authored conditional.

Evidence:

> Collapsing only the duplicated x278/x270/x274 assignments into one unconditional block changed exact _tyList_80313774 from 100% to 94.83274%. The exact-match source was restored as required.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/ty/tylist.c:618`

> The cursor creation conditional has byte-for-byte equivalent calls in both branches. This is another generated branch with no semantic selection and no permitted evidence-backed exception.

Evidence:

> Collapsing only the duplicated x278/x270/x274 assignments into one unconditional block changed exact _tyList_80313774 from 100% to 94.83274%. The exact-match source was restored as required.



</details>

src/melee/ty/tydisplay.c
```diff
@@ -81,6 +81,9 @@
 /* 4D6F28 */ static s32 _tyDisplay_804D6F28;
 /* 4D6F2C */ static HSD_GObj* _tyDisplay_804D6F2C;
 
+extern TyDspNameTables _tyDisplay_803B8988;
+extern TyDspArchNames _tyDisplay_803B8A34;
```
ribbanya: These shouldn't be `extern`.

src/melee/ty/tydisplay.c
```diff
@@ -1977,7 +2176,9 @@ static inline HSD_JObj* un_8031BF34_inline(void)
 
 void _tyDisplay_8031BF34(s32 arg0)
 {
-    TyDspArchiveHolder* a = &_tyDisplay_804A2DD0;
+#ifdef __MWERKS__
+    void Toy_80308250(TyDspArchiveHolder*, s16, s32);
+#endif
```
ribbanya: Not allowed.

src/melee/ty/tyfigupon.c
ribbanya: True.

src/melee/ty/tylist.c
```diff
@@ -57,6 +57,27 @@
...
+    /* 0x13 */ s8 x13;
+    /* 0x14 */ s8 x14;
+    /* 0x15 */ u8 pad_15[3];
+    /* 0x18 */ HSD_GObj* x18;
+} TyListGobjEntryInner;
+typedef struct TyListAuxState {
+    /* 0x00 */ u8 pad_0[0x18];
+    /* 0x18 */ HSD_GObj* gobj;
+    /* 0x1C */ u8 pad_1C[0x10];
+} TyListAuxState;
```
ribbanya, line 64: Move above the function declarations.

ribbanya (changes requested):

src/melee/ty/tydisplay.c
fjooord: Fixed in 308c09746c — the tables are now proper in-TU `static TyDspArchNames const` definitions (they're tydisplay's own .rodata; the externs had been resolving onto the anonymous initializer templates of `tyDisplay_8031BB34/94/F4`, which now copy from the named tables instead). `tyDisplay_8031C454`/`_8031C5E4` view the three adjacent tables as a `TyDspNameTables` block, matching how retail indexes across them from the first table — noted with a comment. Byte-identical: all previously-matching functions stay 100% and section data is unchanged. (A single composite `TyDspNameTables` object would read slightly cleaner but costs `tyDisplay_8031BB94`/`_tyDisplay_8031BBF4` their 100% — 99.75 each — because the compiler folds the interior offsets differently than retail; happy to switch if you prefer that form regardless.)

src/melee/ty/tydisplay.c
fjooord: Removed in 308c09746c — the toy.h prototype is used as-is; the local declaration turned out to be residue with no codegen effect (function stays 100%).

src/melee/ty/tyfigupon.c
fjooord: Agreed and fixed in 308c09746c — `strings + 0x4EC` was exactly `&_tyFigupon_803FEEFC` (already defined in this TU), so both sites now reference the camera desc directly and the offset trick is gone. Byte-identical.

src/melee/ty/tylist.c
fjooord: Done in 308c09746c.

ribbanya (approved):

PR #2880, opened 2026-07-16
Melee decomp 8/11: GR stages

fjooord:
## Summary

Exact-match decompilation of 29 file(s) (75 newly matched function(s)). Produced by the decomp-orchestrator pipeline; only runner-validated exact matches ship.

## Verification

- Published from the local branch commit `8bc1ad49f1`.


src/melee/gr/grcastle.c
```diff
@@ -1369,6 +1384,12 @@ void grCastle_801CF308(Ground_GObj* gobj)
     }
 }
 
+/// Preserve the original .sdata2 placement of the double-precision 0.5.
```
fjooord, line 1387: **`preship_reject_global_standard_literals_and_data_ownership`**

> The added no-op `grCastle_ForceSdata2Order` function exists solely to force placement of the `0.5` double literal in `.sdata2`. The comment explicitly identifies this as a data-ordering anchor rather than authored behavior, and it supplies no section/symbol ownership metadata or split-boundary evidence. This is the same class of data-ordering dodge rejected by the maintainer exhibits.

**Match context:**

- match status unavailable

src/melee/gr/grinishie1.c
```diff
@@ -35,8 +36,27 @@
 
 #include <baselib/gobj.h>
 
+static void sdata2_order(void)
```
fjooord, line 40: **`preship_reject_global_standard_literals_and_data_ownership`**

> The discarded float and conversion literals in sdata2_order are artificial anchors for sdata2 emission. No section or symbol ownership evidence is recorded in the diff, so using literals solely to manipulate data layout is not reviewable source.

**Match context:**

- match status unavailable

src/melee/gr/grinishie1.c
```diff
@@ -35,8 +36,27 @@
...
+    (void) 5.0f;
+    (void) 0.69999999999999996;
+    (void) 100000.0f;
+    (void) -30.0f;
+    (void) S32_TO_F32;
+}
+
 #define invalid_state(line)                                                   \
-    OSReport("%s:%d: oioi..\n", "i!=HATENA_APPEAR_CHECKLOOP", line);          \
+    OSReport("%s:%d: oioi..\n", "grinishie1.c", line);                        \
```
fjooord: **`preship_reject_global_standard_assert_report_macros`**

> The deterministic assert_idiom_downgrade lint error reports that this diff removes HSD_ASSERT-family uses while adding raw OSReport/assert-style code. The changed invalid_state expansion retains an open-coded report/control-flow idiom instead of the established project assert/report macro family. This hard lint error cannot be waived.

**Match context:**

- match status unavailable

src/melee/gr/grkongo.static.h
```diff
@@ -18,6 +18,20 @@
...
+/// spline table, as taken by #Ground_801C247C.
+typedef struct grKg_SplineChoice {
+    s32 file;
+    s32 spline;
+} grKg_SplineChoice;
+
+typedef struct grKg_SplineChoiceList {
+    grKg_SplineChoice pairs[3];
+    s32 terminator;
+} grKg_SplineChoiceList;
```
fjooord, line 31: **`preship_reject_global_standard_literals_and_data_ownership`**

> The diff introduces named constant storage through `extern const grKg_SplineChoiceList grKg_SplineChoices;` without recording section, symbol, split-boundary, or other ownership evidence. Moving the declaration into a header avoids the narrower `extern_in_c` ban, but it does not satisfy the mandatory requirement that newly named data have documented ownership evidence.

**Match context:**

- match status unavailable

src/sysdolphin/baselib/debug.h
```diff
@@ -23,6 +23,16 @@ ATTRIBUTE_NORETURN void HSD_Panic(char*, u32, char*);
 #define HSD_ASSERTREPORT(line, cond, ...)                                     \
     ((cond) ? (void) 0                                                        \
             : (OSReport(__VA_ARGS__), __assert(__FILE__, line, #cond)))
+/// HSD_ASSERTREPORT for code split out of a larger retail TU: the retail
+/// __FILE__ string and stringized condition belong to the original TU's
```
fjooord: **`preship_reject_global_standard_assert_report_macros`**

> The new HSD_ASSERTREPORTFILE macro is a fake assert/report variant that exposes the filename and stringized condition as caller-supplied data solely to preserve strings from a former merged TU. Its body manually expands OSReport plus __assert instead of retaining the established HSD_ASSERTREPORT idiom. The adjacent comment explicitly identifies split-TU data placement as the motivation, making this a data-ordering workaround rather than an authored project macro. The standard has no exception for replacing an existing macro form with a custom raw expansion.

Also at lines: 30, 33, 37

**Match context:**

- match status unavailable

src/melee/gr/grbigblue.c
```diff
@@ -451,8 +525,13 @@ void grBigBlue_801E6364(Ground_GObj* gobj)
...
 
+    cur = car_gobj->hsd_obj;
     i = 0;
-    cur = HSD_JObjGetChild(car_gobj->hsd_obj);
+    if (cur == NULL) {
+        cur = NULL;
+    } else {
+        cur = cur->child;
+    }
     rot_y = grBb_804DB304;
```
fjooord, line 537: **`preship_reject_global_standard_header_inlines`**

> The canonical `HSD_JObjGetChild(car_gobj->hsd_obj)` operation is replaced by a manually expanded null check and direct `child` field access. This is precisely the kind of manual header-inline expansion that the standard requires restoring to the shared helper.

**Match context:**

- match status unavailable

src/melee/gr/grbigblue.c
```diff
@@ -1509,224 +1628,246 @@ void grBigBlue_801E93D8(Ground_GObj* gobj)
...
+                    if (height != -3.4028235e38f) {
+                        f32 cam_right;
+                        f32 cam_left;
+                        f32 cam_bot;
+                        f32 speed;
+                        s32 collided;
+
+                        pos.y = height + grBb_804D69C8->xCC;
+                        speed = 140.0f * Ground_801C0498();
+                        collided = grBigBlue_801E8794(
```
fjooord, line 1672: **`numeric_literal_to_symbol`**

> Numeric literal replaced by address-style data symbol `grBigBlue_801E8794`. Keep float/constants inline unless the PR is explicitly scoped to evidenced data ownership. Keep literals inline unless data ownership evidence says otherwise.

Also at lines: 1674, 1698, 1739, 1808, 1817, 1856

**Match context:**

- match status unavailable

src/melee/gr/grbigblue.c
```diff
@@ -1509,224 +1628,246 @@ void grBigBlue_801E93D8(Ground_GObj* gobj)
...
+                                             cam_bot, cam_left, cam_right))
+                            {
+                                collided = 1;
+                            }
+                        }
+                        if (collided != 0) {
+                            pos.y = 30.0f + Stage_GetCamBoundsTopOffset();
+                        }
+                        if (pos.y == -3.4028235e38f) {
+                            OSReport("*** Not Set Position!(FFlyer)\n");
```
fjooord: **`assert_idiom_downgrade`**

> File diff removes HSD_ASSERT* and adds raw __assert/OSReport code. Keep the project assert/report idiom unless there is evidence the source really used the raw call. Use project assert/report macros (HSD_ASSERT*) when they represent the source.

**Match context:**

- match status unavailable

src/melee/gr/grfourside.c
```diff
@@ -334,10 +332,15 @@ void grFourside_801F3274(Ground_GObj* gobj)
                 if (fVar1 < 0.0f) {
                     fVar1 = -fVar1;
                 }
+                rand_max2 = (s32) fVar1;
+                rand_max = (s32) fVar1;
+                fVar1 = rand_max != 0 ? HSD_Randi(rand_max2) : 0;
                 temp_fVar1 = grFs_804D69D8->crane_iron_down_min;
+                if (temp_fVar1 < 0.0f) {
```
fjooord, line 339: **`preship_reject_global_standard_canonical_control_flow_and_macros`**

> The added branch that manually negates `temp_fVar1` is a raw absolute-value expansion. This standard requires canonical `ABS`/min/max forms unless recorded local evidence proves the canonical form cannot represent the source. The same added expansion appears again around line 358.

**Match context:**

- match status unavailable

fjooord:
<details>

**`preship_reject_global_standard_literals_and_data_ownership`** at `src/melee/gr/grbigblueroute.c:35`

> The discarded float sequence promotes and orders numeric constants solely to manipulate generated `.sdata2`. The diff records no section or symbol ownership evidence permitting named or synthetic data-ordering source. This is the same class of unfinished data-order dodge rejected by the maintainer for self-TU `.sdata2` anchors.

Evidence:

> A clean ownership attempt used the canonical metadata-backed grBb_Route_803E6200 symbol, which config/GALE01/symbols.txt records as global .data at 0x803E6200 with size 0x3C and splits.txt assigns to this TU. Diff-aware lint nevertheless rejected the new address-named definition. Adding or changing ownership metadata is outside this source-only repair lane, so the original candidate table was restored.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/gr/gricemt.c:1834`

> The change replaces a concise conditional traversal loop with `while (true)` plus two duplicated parent lookups, assignments, breaks, and an explicit null case. This is a generated branch-ladder shape rather than canonical structured traversal, and the diff contains no recorded local evidence that the canonical condition cannot represent the authored source.

Evidence:

> grIceMt_801FA500 was exact before repair. Two normal condition-controlled ascent loops with one successor selection were tested and measured at 79.91525% and 76.89831%, respectively. Per the exact-function checkpoint rule, the prior source was restored and revalidated at 100.00000%.

**`assert_idiom_downgrade`** at `src/melee/gr/grinishie1.c:59`

> File diff removes HSD_ASSERT* and adds raw __assert/OSReport code. Keep the project assert/report idiom unless there is evidence the source really used the raw call. Use project assert/report macros (HSD_ASSERT*) when they represent the source.

Evidence:

> Restoring the baseline report operand removed the diff-added raw report shape but regressed exact grInishie1_801FAD84 to 99.91089%; the checkpoint rule required restoring the prior source.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/gr/grinishie1.c:891`

> The existing fclamp0 helper calls are expanded into repeated conditional assignments. The standard requires canonical clamp/max forms instead of generated branch expansions, absent recorded evidence that the canonical form is wrong.

Evidence:

> Restoring fabsf_inline as part of the canonical grInishie1_801FC110 body changed the previously exact function to 97.70175%; the target was restored.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/gr/grinishie1.c:902`

> The existing fabsf_inline call is replaced by an open-coded ternary absolute-value expression. Raw absolute expressions must use the canonical ABS/helper form unless contrary source evidence is recorded.

Evidence:

> Restoring fabsf_inline as part of the canonical grInishie1_801FC110 body changed the previously exact function to 97.70175%; the target was restored.

**`preship_warn_global_standard_literals_and_data_ownership`** at `src/melee/gr/grkongo.c:705`

> The data object was moved after its first use and is apparently forward-declared through `grkongo.static.h`. This matches the accepted same-TU forward-definition counter-exhibit rather than the banned unresolved extern-anchor pattern, and the adjacent comment records concrete MWCC section-anchor/code-generation evidence. It is nevertheless an exact-match-sensitive declaration-order tradeoff that the maintainer has historically treated as undesirable style.

Evidence:

> Warning repair was not required. Lines 702-709 document that grKg_SplineChoices is declared by grkongo.static.h and defined in this same translation unit after its first use because the alternative definition order emits .rodata-section-anchor accesses. Whole-file checkdiff reports the owning function grKongo_801D651C as exact. No unresolved or cross-TU extern anchor was introduced.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/gr/ground.c:1919`

> A natural sentinel-copy loop was changed into `do { ... if (...) break; ... } while (true)`. This infinite-loop-plus-break form is a generated control-flow ladder, and the diff records no local evidence that a canonical while/do-while condition cannot represent the source.

Evidence:

> A condition-controlled `while (i < entry_count)` search with the not-found return after the loop compiled but lowered already-exact Ground_801C34AC from 100% to 93.73288%. The target was restored and revalidated at 100% as required.

**`preship_reject_global_standard_canonical_control_flow_and_macros`** at `src/melee/gr/grvenom.c:299`

> The added `do { ... } while (false)` is a generated branch-shaping construct whose only purpose is to manufacture `break`/`return` control flow. It replaces straightforward conditionals with a single-iteration fake loop, and the diff records no local evidence that a canonical conditional form cannot represent the source.

Evidence:

> A normal if-based implementation of the combined cases reduced exact grVenom_8020362C from 100% to 97.40446% in artifact melee_gr_grvenom-89f7d403.o. The checkpoint required restoration of the prior target shape.



</details>

config/GALE01/symbols.txt
ribbanya, line 1: See https://github.com/doldecomp/melee/pull/2877#discussion_r3598762309

src/melee/gr/grcastle.c
ribbanya: It's fine, just add a `@todo`.

src/melee/gr/grgreens.c
```diff
@@ -717,6 +710,24 @@ void grGreens_80213C10(Ground_GObj* gobj)
     }
 }
 
+/* These tables live at 0x803E7840/0x803E787C, after grGreens_80213C10's
+ * switch jumptable (0x803E7824): MWCC emits .data in definition/creation
+ * order, so they are defined after that function on purpose. */
```
ribbanya: Unnecessary comment.

src/melee/gr/grgreens.c
```diff
@@ -931,50 +937,63 @@ void grGreens_80214FA8(Ground_GObj* gobj)
...
-
-    if (arg2 > 0 && arg1 > 0 && getBlock(gp, arg2 - 1, arg1 - 1)->status == 3)
-    {
-        if (getBlock(gp, arg2 - 1, arg1)->status == 3) {
-            mpLib_800581DC(getBlock(gp, arg2 - 1, arg1 - 1)->x18,
-                           getBlock(gp, arg2 - 1, arg1)->x18);
-        } else if (getBlock(gp, arg2, arg1 - 1)->status == 3) {
-            mpLib_800581DC(getBlock(gp, arg2 - 1, arg1 - 1)->x18,
-                           getBlock(gp, arg2, arg1 - 1)->x18);
+#define BLOCK(i, j) (*(BlockRow**) &gp->gv.greens.x8_blocks)[i][j]
```
ribbanya: Remove and expand.

src/melee/gr/grgreens.c
```diff
@@ -1106,183 +1127,210 @@ void grGreens_802159B8(Ground* gp, int i, int j, int value)
...
+    grGreens_802159B8(ground, col, row, (s32) hit);
 }
 
 void fn_80215D50(Item_GObj* item_gobj, Ground* gp, HSD_GObj* gobj)
 {
     return;
 }
 
+#define BLOCK(gp, i, j)                                                       \
+    (((struct grGreens_BlockVars(*)[6])(gp)->gv.greens.x8_blocks)[i][j])
```
ribbanya: Remove and expand.

src/melee/gr/grgreens.c
```diff
@@ -1106,183 +1127,210 @@ void grGreens_802159B8(Ground* gp, int i, int j, int value)
...
             row = 0;
         }
     }
     return (s32) gobj;
 }
 
+#undef BLOCK
+
+#define BLOCK(gp, i, j) ((gp)->gv.greens.x8_blocks + (j))[(i) * 6]
+#define VEC(gp, i, j) ((gp)->gv.greens.x4 + (j))[(i) * 6]
```
ribbanya: Stop.

ribbanya (changes requested):

config/GALE01/symbols.txt
fjooord: Reverted all the meaning-named renames to master/address-style names in both symbols.txt and the sources (d03269db4a) — the remaining symbols.txt diff is only size fixes, static-scope corrections, and the string/data splits the section pairing needs. The dropped names are queued for a separate naming PR.

src/melee/gr/grcastle.c
fjooord: Added the @todo.

src/melee/gr/grgreens.c
fjooord: Removed.

src/melee/gr/grgreens.c
fjooord: Removed the macro and expanded the accesses inline; the object is unchanged — grGreens_802150C4 stays at 100%.

src/melee/gr/grgreens.c
fjooord: Removed and expanded; grGreens_80215D54 is byte-identical, still 100%.

src/melee/gr/grgreens.c
fjooord: Stopped — the redefinition layer is gone and all accesses are written out inline. Scores unchanged.

src/melee/gr/grbigblue.c
ribbanya: Correct.

src/melee/gr/grinishie1.c
ribbanya: The rule should be changed to allow these ordering functions as long as they have a todo comment.

src/melee/gr/grinishie1.c
ribbanya: Bad macro.

src/sysdolphin/baselib/debug.h
fjooord: Resolved in f61441fa3d, mirroring the ruling on the identical macro block in #2878: the `HSD_ASSERTMSGFILE`/`HSD_ASSERTREPORTFILE` macros are removed and debug.h is byte-identical to master again. grhomerun.c's four call sites now use direct `__assert(grHr_803E821C.filename, …)` calls — the same pattern grkongo.c already uses on master for a retail file string that lives in the TU's own data — with no change to the TU's output or to symbols.txt.

src/melee/gr/grgreens.c
```diff
@@ -152,6 +162,9 @@ grGr_StageData grGr_803E76D0 = {
 
 char grGr_803E7728[] = "grgreens.c";
 
+#undef __FILE__
+#define __FILE__ (&grGr_803E7728[0])
```
ribbanya: Not allowed.

src/melee/gr/grhomerun.c
```diff
@@ -770,69 +769,77 @@ void grHomeRun_8021E4F8(Ground_GObj* arg) {}
...
     } else if (idx == 3) {
-        if (grHr_804D6ADC - 1 != arg0 % grHr_804D6ADC) {
+        if (grHr_804D6ADC - 1 != arg % grHr_804D6ADC) {
             gobj_id = 8;
         } else {
             gobj_id = 9;
         }
     } else {
-        HSD_ASSERT(0x3D2, 0);
+        __assert(grHr_803E821C.filename, 0x3D2, "0");
```
ribbanya: None of these assertion regressions are allowed.

src/melee/gr/grkongo.c
```diff
@@ -729,6 +682,25 @@ void grKongo_801D651C(Ground_GObj* gobj)
...
+/// spline to take from that file's spline table.
+///
+/// Defined after #grKongo_801D651C (declared in grkongo.static.h) to
+/// match: at the reference's code for that function, the compiler
+/// addresses this object through the symbol itself rather than a
+/// .rodata section anchor, which MWCC only produces when the object
+/// is not yet defined at the point of use. Verified against the
+/// section-anchored form (lwz offsets 0x24..0x38 off the .rodata
+/// base) produced when the definition precedes the function.
+const grKg_SplineChoiceList grKg_803B7FD4 = {
```
ribbanya:
```suggestion
/* 3B7FD4 */ const grKg_SplineChoiceList grKg_803B7FD4 = {
```

src/melee/gr/grkongo.static.h
ribbanya: Its only usage in the dol is by `grKongo_801D651C`, what more evidence is needed?

src/melee/gr/grmutecity.c
```diff
@@ -1841,14 +1823,18 @@ DynamicModelDesc* grMuteCity_801F28A8(void)
     return NULL;
 }
 
+typedef struct grMc_StackPadArg {
+    u32 x[4];
+} grMc_StackPadArg;
+
+static inline void grMc_StackPad(grMc_StackPadArg arg) {}
```
ribbanya, line 1831: Add todo comment for fake function.

src/melee/gr/grpura.c
```diff
@@ -398,31 +398,21 @@ void grPura_80212CD4(HSD_GObj* arg0)
 {
     Ground* gp = arg0->user_data;
     Ground* gp2 = gp;
-    HSD_JObj* jobj = arg0->hsd_obj;
-    u32 i;
+    HSD_JObj* jobj = HSD_GObjGetHSDObj(arg0);
```
ribbanya, line 401:
```suggestion
    HSD_JObj* jobj = GET_JOBJ(arg0);
```

src/melee/gr/grvenom.c
```diff
@@ -27,35 +27,26 @@
...
-        HSD_GObj* arwing_gobj[3];
-        int arwing_type[3];
-    } arwing;
+    char x0[0x44];
+    StageCallbacks callbacks[1]; // at offset 0x44
+    char x58[0x1B8 - 0x58];
+    char x1B8_file[0x1DC - 0x1B8]; // 0x1B8: "%s:%d..." string
+    char x1DC_func[0x20];          // 0x1DC: function name or filename
+    char x1FC[0x2BC - 0x1FC];
+    char x2BC[0x30];
```
ribbanya: Mistaken decompilation of string pooling; revert.

src/melee/gr/grvenom.c
```diff
@@ -196,10 +187,41 @@ StageData grVe_803E54CC = {
...
+typedef struct grVe_ExtData {
+    char pad[0x20];
+    void* x20[3];
+    s32 x2C[3];
+    s32 x38[3];
+    char pad44[0x338 - 0x44];
+    s32 x338[5];
+} grVe_ExtData;
+
+#define VE_DATA ((grVe_ExtData*) &grVe_803E5348)
```
ribbanya: Expand and delete.

src/melee/gr/grvenom.c
```diff
@@ -212,43 +234,55 @@ typedef struct grVe_GroundData {
...
-    int* x38_ptr;
-    HSD_GObj** x20_ptr;
-    HSD_GObj** x20_next;
-    PAD_STACK(0x28);
+    s32* x2c_ptr;
+    s32* x38_ptr;
+    void** x20_ptr;
+    void** x20_next;
+    /* Objdiff checkpoint: this 0x20 pad preserves the exact 0x68-byte frame.
+     */
```
ribbanya: Remove yap comment.

src/melee/gr/grvenom.c
```diff
@@ -764,154 +803,158 @@ bool grVenom_80204544(Ground_GObj* arg)
...
-    Ground* gp;
-    s32 i;
-    Ground* gp_save;
-    HSD_GObj* lgobj;
-    HSD_LObj* lobj;
-    HSD_LObj* next;
-    f32 hi;
-    f32 lo;
-    PAD_STACK(8);
+    /* Objdiff checkpoint: the split 4+4 pads preserve the exact frame. */
```
ribbanya:
```suggestion
```

src/melee/gr/grvenom.c
```diff
@@ -764,154 +803,158 @@ bool grVenom_80204544(Ground_GObj* arg)
...
-                    if (lobj->position != NULL) {
-                        HSD_ForeachAnim(lobj->position, WOBJ_TYPE,
-                                        ALL_TYPE_MASK, HSD_AObjReqAnim,
-                                        AOBJ_ARG_AF, 0.0);
-                    }
-                    if (lobj->interest != NULL) {
-                        HSD_ForeachAnim(lobj->interest, WOBJ_TYPE,
-                                        ALL_TYPE_MASK, HSD_AObjReqAnim,
-                                        AOBJ_ARG_AF, 0.0);
+                if (HSD_JObjGetFlags(gp->gv.venom2.xC4) & 0x10) {
```
ribbanya:
```suggestion
                if (HSD_JObjGetFlags(gp->gv.venom2.xC4) & JOBJ_HIDDEN) {
```

src/melee/gr/grvenom.c
```diff
@@ -1042,37 +1085,32 @@ void grVenom_80204EFC(Ground_GObj* arg)
...
 
-static int grVe_803E5530[] = { -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
-
-static Vec3 grVe_803E5560[12] = { 0 };
-
-static int grVe_803E55F0[] = { 3, 3, 3, 3, 6 };
+/// Error message read through the stage data base pointer in
+/// #grVenom_80204F20 (`base + 0x2BC`); sits between #grVe_803E5530 and
+/// the TU's pooled assert strings in the reference object.
+static char grVe_803E5604[] = "arwin:user data error\n";
```
ribbanya: Use inlined literal instead.

src/melee/gr/grvenom.c
```diff
@@ -1261,11 +1299,16 @@ void grVenom_802053B0(Ground_GObj* gobj)
...
+
+/// Per-group map animation ids (cf. Corneria's grCn_803E217C).
+static int grVe_803E566C[5] = { 0x0A, 0x0A, 0x0B, 0x0C, 0x0D };
 
-static int grVe_803E566C[] = { 10, 10, 11, 12, 13 };
+/// Joint id per Arwing group (cf. Corneria's grCn_803E2190).
+static int grVe_803E5680[5] = { 4, 4, 4, 4, 3 };
 
-static int grVe_803E5680[] = { 4, 4, 4, 4, 3 };
+/// grVenom_802056B0
```
ribbanya: ???

src/melee/gr/grvenom.c
```diff
@@ -1301,139 +1344,117 @@ bool grVenom_80205750(Ground_GObj* arg)
...
 }
 
+/// Per-group animation flags (cf. Corneria's grCn_803E21B0); the
+/// trailing zero word is present in the reference object.
+static int grVe_803E56A0[6] = { 1, 1, 1, 1, 1, 0 };
+
 void grVenom_80205AD0(Ground_GObj* arg) {}
 
-/// @todo Only differs by register allocation.
+/// grVenom_80205AD4
```
ribbanya:
```suggestion
```

src/melee/gr/grvenom.c
```diff
@@ -1301,139 +1344,117 @@ bool grVenom_80205750(Ground_GObj* arg)
...
     gp->gv.venom.xC8 = grVe_804D6A34;
-    *(s32*) &gp->gv.venom.xEC = zero;
-    *(s32*) &gp->gv.venom.xE8 = zero;
-    *(s32*) &gp->gv.venom.xE4 = zero;
-    *(s32*) &gp->gv.venom.xE0 = zero;
-
+    *(s32*) &gp->gv.venom.xEC = 0;
+    *(s32*) &gp->gv.venom.xE8 = 0;
+    *(s32*) &gp->gv.venom.xE4 = 0;
+    *(s32*) &gp->gv.venom.xE0 = 0;
```
ribbanya: Change the type of the members to `s32` instead of casting pointers.

src/melee/gr/grvenom.c
```diff
@@ -1483,16 +1504,40 @@ s32 grVenom_80205E84(Vec3* pos)
     return 0;
 }
 
-/// @todo Incomplete decompilation.
-
-static int grVe_803E56A0[] = { 1, 1, 1, 1, 1 };
+/// grVenom_80205F30
+/// @todo First attempt at decomp; complex 592-instruction function.
```
ribbanya:
```suggestion
```

src/melee/gr/grvenom.c
```diff
@@ -1518,7 +1563,9 @@ void grVenom_80205F30(Ground_GObj* gobj)
     jobj = gobj->hsd_obj;
     sp94 = grVe_803B82D0;
     sp88 = grVe_803B82DC;
-    PAD_STACK(0x28);
+    /* Objdiff: 0x18 with padA8 reproduces the retail frame and instructions.
+     */
```
ribbanya: Keep the "Objdiff" comments to yourself, please.

src/melee/gr/grvenom.c
```diff
@@ -1765,8 +1823,11 @@ void grVenom_80205F30(Ground_GObj* gobj)
 
 void grVenom_80206870(Ground_GObj* arg) {}
 
+/// grVenom_80206874
```
ribbanya:
```suggestion
```

src/melee/gr/grvenom.c
```diff
@@ -1862,6 +1922,8 @@ void grVenom_80206BC4(Ground_GObj* gobj)
 
 void grVenom_80206BEC(Ground_GObj* arg) {}
 
+extern char grVe_803E5524[];
```
ribbanya: Revert named literal declaration. This is `__FILE__`.

src/melee/gr/gryorster.c
```diff
@@ -181,48 +181,52 @@ void grYorster_80202254(Ground_GObj* gobj)
 
 void grYorster_802022A0(HSD_GObj* gobj) {}
 
+static inline Ground* grYorster_GetGround(HSD_GObj* gobj)
+{
+    return gobj->user_data;
+}
```
ribbanya, line 187: Delete and use `GET_GROUND`. If that doesn't match, just use `gobj->user_data` at the call site.

src/melee/gr/gronett.c
```diff
@@ -473,13 +473,86 @@ bool grOnett_801E43D8(Ground_GObj* gobj)
     return false;
 }
 
+static inline f32 grOnett_Randf(void)
```
ribbanya: Rename to `HSD_RandF_noinline`

src/melee/gr/gronett.c
```diff
@@ -473,13 +473,86 @@ bool grOnett_801E43D8(Ground_GObj* gobj)
...
+{
+    return HSD_Randf();
+}
+
+static inline HSD_JObj* grOnett_GetJObjChild(HSD_JObj* jobj)
```
ribbanya: What is wrong with the `jobj.h` version?

src/melee/gr/grzebes.c
```diff
@@ -1553,149 +1563,185 @@ s32 grZebes_801DB088(Ground* gp, s32 arg1)
...
-                                f32 dx2 = dx * dx;
-                                f32 dy2 = dy * dy;
-                                f32 dist = dx2 + dy2;
-                                if (dist < best_dist) {
-                                    best_idx = j;
-                                    best_dist = dist;
+                            HSD_JObjRemoveAll(*jobjp);
+                            *jobjp = NULL;
+                            {
+                                HSD_JObj* jobj = (HSD_JObj*) (*gobjp)->hsd_obj;
```
ribbanya:
```suggestion
                                HSD_JObj* jobj = GET_JOBJ(gobjp);
```

src/melee/gr/grzebes.c
```diff
@@ -1553,149 +1563,185 @@ s32 grZebes_801DB088(Ground* gp, s32 arg1)
...
-                    break;
-                }
                 }
             }
         }
     }
     return result;
 }
 
+static inline f32 grZebes_801DB3CC_sqrt(f32 x)
```
ribbanya: Use `math_ppc` version.

ribbanya (changes requested):

src/melee/gr/grgreens.c
fjooord: Removed (02b61a6a41). Note: retail's asserts reference the TU's `grGr_803E7728` array; with canonical `__FILE__` a second pooled "grgreens.c" string appears — four exact functions go 100 → 99.9 and .data 100 → 99.23. Kept the compliant form.

src/melee/gr/grhomerun.c
fjooord: Restored to master's HSD_ASSERT body verbatim (02b61a6a41); grHomeRun_8021E500 back at the baseline score.

src/melee/gr/grkongo.c
fjooord: Applied as suggested.

src/melee/gr/grkongo.static.h
fjooord: Acknowledged — the grKongo_801D651C usage is sufficient evidence; flag withdrawn, typed struct stays.

src/melee/gr/grmutecity.c
fjooord: Added (02b61a6a41).

src/melee/gr/grpura.c
fjooord: Tried it — the cast alone perturbs MWCC's register allocation here (grPura_80212CD4 100 → 98.79, pure r4/r5/r6 swap), so we kept the 100% form per the 100%-line ruling. Happy to take the suggestion anyway if you prefer it over the match.

src/melee/gr/grvenom.c
fjooord: Reverted to the typed grVe_Data joints/arwing model (the #2844 form); grVenom_8020362C holds 100%.

src/melee/gr/grvenom.c
fjooord: Expanded and deleted; typed `grVe_Data*` accesses now. One small typed accessor inline remains — MWCC otherwise reorders the first type load (99.36 vs 100).

src/melee/gr/grvenom.c
fjooord: Removed — and swept the whole slice; no tooling comments remain.

src/melee/gr/grvenom.c
fjooord: Removed.

src/melee/gr/grvenom.c
fjooord: Applied as suggested.

src/melee/gr/grvenom.c
fjooord: Inlined; the named static is gone and .data stays 100%.

src/melee/gr/grvenom.c
fjooord: Removed the stray name comment.

src/melee/gr/grvenom.c
fjooord: Removed.

src/melee/gr/grvenom.c
fjooord: Done — `xD4`/`xD8` were genuinely ints (every use was an int store/load; 14 casts dropped), retyped to `s32` in gr/types.h. `xE0`–`xEC` are floats elsewhere (they serve as a Vec3 translate base), so they got f32/s32 union views per the existing xDC precedent, with `_int` stores at the zero/-1 sites. grvenom unit unchanged.

src/melee/gr/grvenom.c
fjooord: Removed.

src/melee/gr/grvenom.c
fjooord: Removed, slice-wide — none remain.

src/melee/gr/grvenom.c
fjooord: Removed.

src/melee/gr/grvenom.c
fjooord: Reverted — the extern is gone; the assert takes `__FILE__`.

src/melee/gr/gryorster.c
fjooord: Measured both suggested forms: `GET_GROUND` scores 86.58 and `gobj->user_data` 86.72 (the merge-base itself is 87.44), while the local inline holds 100 — the call boundary is load-bearing for the match. Kept the 100% form; say the word if you want the suggestion regardless.

src/melee/gr/gronett.c
fjooord: Renamed.

src/melee/gr/gronett.c
fjooord: The jobj.h inline adds a null-check retail doesn't have here (99.74 → 99.21); used direct `jobj->child` member access instead, same as grOnett_StartCar in this file (99.65).

src/melee/gr/grzebes.c
fjooord: Applied — as `GET_JOBJ(*gobjp)`, since that site holds an `HSD_GObj**` (the suggestion's bare `gobjp` looks like a typo).

src/melee/gr/grzebes.c
fjooord: Done; costs ~0.9 on the still-fuzzy 801DB3CC (92.49 → 91.56) — acceptable per the sub-100 rule.

src/melee/gr/grbigblue.c
fjooord: Investigated: retail's 801EB4AC has no assert call at all — removing `HSD_ASSERT(979, jobj)` is what completed the match (99.13 with it, 100 without). The 2328 site already uses HSD_ASSERTMSG with retail's string. No branch-introduced raw __assert remains in the file.

src/melee/gr/grinishie1.c
fjooord: Expanded and deleted at all four sites; byte-identical.

src/melee/gr/grinishie1.c
fjooord: Agreed — the rule change is queued in our standards set (anchors pass with a todo), and all four order anchors in this slice now carry the `/// @todo .sdata2 order hack` comment.

src/melee/gr/grbigblue.c
ribbanya: Shouldn't this be `HSD_ASSERTREPORT`?

src/melee/gr/grbigblue.c
```diff
@@ -2293,47 +2446,49 @@ void grBigBlue_801EB004(Ground_GObj* gobj)
...
+        HSD_ASSERTMSG(2328, child, "jobj");
         HSD_ASSERT(2329, start_jobj);
         HSD_ASSERT(2330, end_jobj);
     }
 
-    ((grBb_InitC4HalfBits*) &gp->gv.bigblue.x0_w)->lane = 0xFFFF;
+    ((grBb_InitC4HalfBits*) (gp + 0xC4))->lane = 0xFFFF;
 
-    ((grBb_InitC4WordBits*) &gp->gv.bigblue.x0_w)->lane = 4;
+    ((grBb_InitC4WordBits*) (gp + 0xC4))->lane = 4;
```
ribbanya: Revert. Use proper struct fields or `M2C_FIELD` for all pointer-offset accesses.

src/melee/gr/grbigblue.c
```diff
@@ -3237,62 +3409,20 @@ void grBigBlue_801ECB50(Ground_GObj* gobj)
...
+                    if ((st == 7 &&
+                         *(f32*) (car + 0xE0) < *(f32*) (target_car + 0xE0)) ||
+                        (st == 8 &&
+                         *(f32*) (car + 0xE0) > *(f32*) (target_car + 0xE0)))
                     {
-                        byte = p3[0xD4];
+                        byte = car[0xD4];
 #ifdef MWERKS_GEKKO
                         asm { rlwimi byte, st_val, 2, 24, 29 }
 #endif
```
ribbanya: Remove.

src/melee/gr/grbigblue.c
```diff
@@ -3308,15 +3438,23 @@ void grBigBlue_801ECB50(Ground_GObj* gobj)
...
+                    active_count = 1;
+                } else {
+                    u32 st = (p[0x114] >> 2) & 0x3F;
+                    p += 0x40;
+                    if (st == 1) {
+                        active_count = 2;
+                    } else if ((u32) ((p[0x114] >> 2) & 0x3F) == 1) {
+                        active_count = 3;
+                    }
+                }
```
ribbanya: These are accessing bitfields on `bp`.

src/melee/gr/grbigblueroute.c
```diff
@@ -664,10 +685,29 @@ void grBigBlueRoute_8020C85C(Ground_GObj* gobj)
...
+    }
+
+    grBigBlueRoute_SpawnRoute(gp->u.car.x108, gp, gobj);
+}
+
 static const Vec3 grBb_Route_803B83E0 = { 0.0f, 1.0f, 0.0f };
 
-/// @todo Currently 89.41% match - SYSTEMIC: float-constant hoisting order and
-/// register pressure differ (target spills rot to stack; we keep it in FPRs).
 #define RE_ENTRY (&((RouteEntry*) gp->u.car.car_info)[i])
```
ribbanya, line 706: Expand and delete.

src/melee/gr/grgreens.c
```diff
@@ -931,71 +931,125 @@ void grGreens_80214FA8(Ground_GObj* gobj)
 
 void grGreens_802150C4(Ground_GObj* gobj, int arg1, int arg2)
 {
+    typedef struct grGreens_BlockVars BlockRow[6];
     Ground* gp = GET_GROUND(gobj);
 
-    if (arg2 > 0 && arg1 > 0 && getBlock(gp, arg2 - 1, arg1 - 1)->status == 3)
+    if (arg2 > 0 && arg1 > 0 &&
+        (*(BlockRow**) &gp->gv.greens.x8_blocks)[arg2 - 1][arg1 - 1].status ==
```
ribbanya: The type definition of `x8_blocks` needs to be changed, or `grGreens_BlockVars` needs to be modified, so that this egregious casting is no longer necessary.

src/melee/gr/grgreens.c
```diff
@@ -1106,31 +1163,42 @@ void grGreens_802159B8(Ground* gp, int i, int j, int value)
...
     int j;
-    int row = -1;
-    int col = 0;
-    u8 operand_pad[4];
 
     for (i = 0; i < 5; i++) {
         for (j = 0; j < 6; j++) {
-            struct grGreens_BlockVars* block = getBlock(gp, i, j);
+            struct grGreens_BlockVars* block =
+                gp->gv.greens.x8_blocks + i * 6 + j;
```
ribbanya: This suggests that `x8_blocks` is a pointer to a 2D array. Definition should be modified accordingly.

src/melee/gr/grhomerun.c
```diff
@@ -200,81 +201,89 @@ void grHomeRun_8021CB1C(Ground_GObj* arg) {}
...
-    s16 i;
+    s32 i;
     f32 gobjs_num;
-    HSD_GObj** gobjs;
-    HSD_GObj** jobjs;
-    PAD_STACK(24);
+    Ground_801C2ED0(jobj2, gp->map_id);
 
-    Ground_801C2ED0(jobj, gp->map_id);
+    ((grHomeRun_MainGroundVars*) &gp->gv.homerun)->gobjs = HSD_MemAlloc(0x100);
```
ribbanya: What is this cast? Just modify the original type of `gv.homerun` to accommodate matching `grhomerun`.

src/melee/gr/grinishie1.c
```diff
@@ -84,6 +86,9 @@ static inline f32 fclamp0(f32 x)
     return ret;
 }
 
+#define INISHIE1_XC6(gp) ((s16*) &(gp)->gv.inishie1.xC4)[1]
+#define INISHIE1_XC8(gp) ((s16*) &(gp)->gv.inishie1.xC4)[2]
```
ribbanya: Expand and delete.

ribbanya (changes requested):

src/melee/gr/grbigblue.c
fjooord: Done — byte-identical (58895d620c).

src/melee/gr/grbigblue.c
fjooord: Done — the function is fully typed (`&gp->gv.bigblue.x0_w`/`.x2`/`.x3` bitfield views, `Ground* gp`) with M2C_FIELD covering the remaining raw offsets; byte-identical. Scope note: this covers the anchored function and the flagged regions — the file's ~100 historical byte-walk sites in unflagged, mostly-matched functions are untouched to avoid unforced regressions; say the word if you want a whole-file sweep.

src/melee/gr/grbigblue.c
fjooord: Removed — both asm blocks are gone, replaced by typed bitfield stores (details in the bitfield thread below).

src/melee/gr/grbigblue.c
fjooord: Modeled it: `state` is now a true `u8 state : 6; u8 state_lo : 2;` bitfield on the CarLane, the four manual `>>2 & 0x3F` scan reads use `.state` directly (byte-identical), and 801ECB50's stores are typed bitfield stores replacing the asm. Measured ladder for the full conversion: all-typed walks 94.15 (MWCC emits lbzu-form walks instead of retail's lbz+addi), M2C_FIELD notation 94.98, open-coded and-or 95.73, byte-walk + typed bitfield stores 96.36 — shipped the last (was 98.99 with the asm; pre-PR baseline is 95.46). The fully-typed form is a one-line change away if you prefer it over the score.

src/melee/gr/grbigblueroute.c
fjooord: Expanded and deleted; byte-identical.

src/melee/gr/grgreens.c
fjooord: Retyped in gr/types.h as `grGreens_BlockVars (*x8_blocks)[6]` — all the casts, `+ i*6 + j` arithmetic, and row-shift shapes collapsed to plain 2D indexing, the local BlockRow typedefs are gone, and the alloc/memzero sizes are expressed as `5 * sizeof(*x8_blocks)`. Every previously-exact function held, and grGreens_80215ED8 improves 92.69 → 96.45.

src/melee/gr/grgreens.c
fjooord: Yes — it's now defined exactly as that (see the thread above); the definition change is what made the casting disappear.

src/melee/gr/grhomerun.c
fjooord: Accommodated — `grHomeRun_MainGroundVars` moved into gr/types.h as a union member of `grHomeRun_GroundVars`, so every cast is gone and access is plain `gp->gv.homerun.main.…`; grHomeRun_8021CB20 improves 97.01 → 97.48. grhomerun.c is the only gv.homerun user tree-wide, so nothing else moves.

src/melee/gr/grinishie1.c
fjooord: Expanded and deleted; byte-identical.

src/melee/gr/grbigblue.c
ribbanya: False positive; that's a function.

src/melee/gr/grbigblue.c
```diff
@@ -2463,32 +2464,32 @@ void grBigBlue_801EB004(Ground_GObj* gobj)
...
     }
 
-    ((grBb_InitC4HalfBits*) (gp + 0xC4))->lane = 0xFFFF;
+    ((grBb_InitC4HalfBits*) &gp->gv.bigblue.x0_w)->lane = 0xFFFF;
 
-    ((grBb_InitC4WordBits*) (gp + 0xC4))->lane = 4;
+    ((grBb_InitC4WordBits*) &gp->gv.bigblue.x0_w)->lane = 4;
 
-    ((grBb_InitByteLo7*) (gp + 0xC6))->lo7 = 0;
+    ((grBb_InitByteLo7*) &gp->gv.bigblue.x2)->lo7 = 0;
```
ribbanya: Change the types of the fields instead of casting.

src/melee/gr/grbigblue.c
```diff
@@ -3418,11 +3414,8 @@ void grBigBlue_801ECB50(Ground_GObj* gobj)
                         (st == 8 &&
                          *(f32*) (car + 0xE0) > *(f32*) (target_car + 0xE0)))
                     {
-                        byte = car[0xD4];
-#ifdef MWERKS_GEKKO
-                        asm { rlwimi byte, st_val, 2, 24, 29 }
-#endif
-                        car[0xD4] = byte;
+                        ((struct grBigBlue_CarLane*) (car + 0xD4))->state =
+                            st_val;
```
ribbanya: Change the type fields instead of casting.

src/melee/gr/grbigblue.c
ribbanya: Yes, use the proper bitfields.

ribbanya (changes requested):

src/melee/gr/grhomerun.c
ribbanya: Why are they two different union members? This is not typical for HAL. Is there a split missing?

src/melee/gr/gronett.c
ribbanya: Fine, add a todo comment above the redefinition then.

src/melee/gr/grpura.c
ribbanya: 100% is preferred.

src/melee/gr/grbigblue.c
fjooord: Done as directed (85ada912a2) — the lane scans use the typed CarLane bitfields and typed lane pointers end to end; no byte-walks or cast-at-offset stores remain. Final number: 94.15 (was 96.36 with the byte-walk, 98.99 with the asm; pre-PR baseline 95.46).

src/melee/gr/grbigblue.c
fjooord: Done — the control word is real bitfields in gr/types.h now (`prev_lane : 7`, `cur_lane : 7`, `next_lane : 7`, …, union'd with the legacy views that other, already-matched functions in the file still use). 801EB004's view-typedefs are deleted and every access is a plain member. −0.5 on 801EB004 from the u32 container emitting word-width accesses where retail mixes lhz/lbz, still well above master (99.42 vs 97.12). Happy to convert the rest of the file off the legacy views in a follow-up if you want the full sweep.

src/melee/gr/grbigblue.c
fjooord: Gone — subsumed by the bitfield conversion above; `car`/`target_car` are typed CarLane pointers throughout.

src/melee/gr/grbigblue.c
fjooord: Confirmed false positive — `Ground_801C0498` is `f32 (void)` in ground.h, `type:function` in symbols.txt. Same lint misread as the gm_80169238 case on #2877 (the function-symbol exemption is already queued in the rule fix). Resolving.

src/melee/gr/grhomerun.c
fjooord: Investigated properly: no missing split — the two views are genuinely two different Ground-gobj subtypes sharing the gv slot, exactly like venom/venom2 and zebes/zebes2: `gv.homerun` is the stage gobj (gobjs/jobj_gobjs/rear state) and the target gobjs created by grHomeRun_8021E500 use their own vars. Remodeled as top-level `homerun`/`homerun2` members of the gv union (the nested union is gone, and the cast in this hunk was stale diff context — none remain on the tip). Object bytes unchanged.

src/melee/gr/grbigblue.c
ribbanya: Let's just keep the bitfield accesses for now and try to find a better match that way.

src/melee/gr/grbigblue.c
```diff
@@ -6,6 +6,7 @@
 #include "grfzerocar.h"
 #include "grmaterial.h"
 #include "ground.h"
+#include "m2c_macros.h"
```
ribbanya:
```suggestion
#include <placeholder.h>
```

ribbanya (changes requested):

src/melee/gr/grfourside.c
ribbanya: If it improves the match it's fine.

ribbanya (dismissed):
Still a lot to work out but let's just merge this pass for now.

PR #2882, opened 2026-07-16
Update style guide

Jacoby6000:
Based on conversations with some of the main contributors, I've updated the style guide to reflect my understanding of what the expected style is for newly written code.  I'm not married to any of this, and happy to make whatever adjustments necessary!

The document as a whole is [best viewed here](https://github.com/Jacoby6000/melee/blob/update-style-guide/.github/CONTRIBUTING.md)

I think with the increase of AI assisted PRs and the increasing renames, the more prescriptive style guide is worth getting hammered out before things get messy.  Should help streamline future contributions, generally speaking.

This is purely a documentation change, though I will likely be making future follow-up PRs with clang-tidy policies that try to follow whatever we decide on here, and then over time try to move the broader codebase to be conformant. 

Once we get all the tooling in to a good place, hopefully PR discussion can be more focused on what the code is/does, rather than how the code looks.

A doxygen style guide will be added in the next PR.

Let the bickering over small details commence!


.github/CONTRIBUTING.md
```diff
@@ -111,6 +150,93 @@ pre-commit run --all-files # Run the hook once on existing code
...
+| Target Type                             | Good  | Bad            |
+|-----------------------------------------|-------|----------------|
+| 8-bit unsigned integer                  | u8    | unsigned char  |
+| 16-bit unsigned integer                 | u16   | unsigned short |
+| 32-bit unsigned integer                 | u32   | unsigned int   |
+| 64-bit unsigned integer                 | u64   | unsigned long  |
+| 8-bit signed integer                    | s8    | char           |
+| 16-bit signed integer                   | s16   | short          |
+| 32-bit signed integer                   | s32   | int            |
+| 64-bit signed integer                   | s64   | long           |
```
r-burns: This would actually be a `long long`, a `long` is only 32 bits

.github/CONTRIBUTING.md
```diff
@@ -111,6 +150,93 @@ pre-commit run --all-files # Run the hook once on existing code
...
+
+| Target Type                             | Good  | Bad            |
+|-----------------------------------------|-------|----------------|
+| 8-bit unsigned integer                  | u8    | unsigned char  |
+| 16-bit unsigned integer                 | u16   | unsigned short |
+| 32-bit unsigned integer                 | u32   | unsigned int   |
+| 64-bit unsigned integer                 | u64   | unsigned long  |
+| 8-bit signed integer                    | s8    | char           |
+| 16-bit signed integer                   | s16   | short          |
+| 32-bit signed integer                   | s32   | int            |
```
r-burns: Sometimes you have to use int, because we currently don't have any typedefs that are int-typed - s32/u32 are signed/unsigned longs, and that can sometimes cause different codegen.

.github/CONTRIBUTING.md
```diff
@@ -111,6 +150,93 @@ pre-commit run --all-files # Run the hook once on existing code
...
+      ```c
+        bool intr = gm_8015CDC8(CKIND_LUIGI)
+      ```
+    - Bad:
+      ```c
+        bool b = gm_IsCKindUnlocked(CKIND_LUIGI)
+      ```
+
+## <a name="primitives"></a>Primitives
+- Prefer primitive types with explicit bit widths
```
r-burns: Personally I prefer to use regular old `int` and not specify the bit width explicitly unless it seems like the original developers would have cared to hardcode that. E.g. for bitmasks, you actually do care about the size of the field, but for generic counters, array indices, or loop index variables, it seems unusual to hardcode the bit width. This is a matter of opinion though, and I'm open to changing that style.

.github/CONTRIBUTING.md
```diff
@@ -111,6 +150,93 @@ pre-commit run --all-files # Run the hook once on existing code
...
+- When a type is an unknown 32 bit integer and likely a pointer, declare it as UNK_T
```
r-burns: I use UNK_T for any unknown types that seem to be 32 bits, even if it seems more likely to be a scalar than a pointer. To me the UNK_T just signifies that it is currently unknown, not necessarily that it is a pointer.

.github/CONTRIBUTING.md
```diff
@@ -47,35 +57,64 @@ pre-commit run --all-files # Run the hook once on existing code
...
+          u8 thing;
+          u32 something;
+          u32 something2;
+          u32 something3;
+          u8 well_known_something;
+          u8 bits
+      };
+      ```
+
+- If you copy structs from another project
```
r-burns:
```suggestion
- Do not copy structs from another project
```

Probably better advice for AI and beginners. You're only allowed to copy stuff if you're experienced and know the narrow cases where it's acceptable. This won't stop experienced contributors like me because we don't know how to read.

.github/CONTRIBUTING.md
ribbanya: We do have `bool` and `enum_t` and I think `ssize_t` lol, but yeah.

`int` is also semantically preferred in a lot of cases because it simply means "the native signed integer type." If anything it should be preferred over `s32` but Dolphin SDK and m2c disagree, so it's crept in all over.

.github/CONTRIBUTING.md
ribbanya: Agreed with @r-burns.

.github/CONTRIBUTING.md
ribbanya: It expands to `M2C_UNK` when using m2c, which I think is internally a `s32`, but I intended it mostly for struct pointers whose layouts are undefined.

ribbanya (changes requested):
- [x] https://github.com/doldecomp/melee/pull/2882#discussion_r3599471311
- [x] https://github.com/doldecomp/melee/pull/2882#discussion_r3599474678
- [x] https://github.com/doldecomp/melee/pull/2882#discussion_r3599480211
- [x] https://github.com/doldecomp/melee/pull/2882#discussion_r3599497922

ribbanya (changes requested):
Please add an explanation of our header format as it's a little unintuitive.

- Every TU header (meaning things that actually appear in `splits.txt`) **must** contain its own symbols according to the split, and **only** those members. Extra inlines, types, etc. are not allowed. Symbols must appear in address order. `static` symbols should appear at the top of the C file, not the header.
- A TU may define a `*.static.h` file. This is a temporary file to aid in m2c decompilation runs, so it should be considered as part of the C file, and entirely private. It should be deleted when the TU is linked (marked as `Matching`).
- `types.h` for each module contains relevant types primary to that module. It's difficult to discern where a type truly belongs, so we're not very strict about that. Types should be only `struct X` and/or `union X` definitions, not `typedef`s.
- One `forward.h` corresponds to each `types.h` and is responsible for providing a `typedef` for each struct and union, and typedefs of function pointers. It is also the place to define `enum`s, global `const` simple types like integers that don't get emitted as symbols, and `#define` constants. The purpose of this separation is to avoid circular inclusions.
- If an `inline` function is shared by multiple TUs but its function body isn't a real symbol, it goes in `inlines.h` for the most relevant module. An exception is made if that inline function contains an assert which proves its source file.
- Documentation belongs in `*.dox` and should be kept out of the C and H files with minimal exceptions. `@todo` comments are allowed anywhere.

.github/CONTRIBUTING.md
```diff
@@ -111,6 +150,93 @@ pre-commit run --all-files # Run the hook once on existing code
...
+
+## <a name="enums"></a>Enums
+- Enum declarations should include a clear means of determining the value for each entry
+  - May be a comment prefix with the hexadecimal value. There must be enough left-padded 0s so that each comment is the
+    same width.
+  - May be an explicit value decl such as `FOO_BAR = 4`
+- Each enum value for a given type should have a unique prefix that makes them easy to identify as a group.
+- Enum value names should be descriptive when well understood.
+- Enum value names should contain their offset when not well understood.
+- Enum value names should be in SCREAMING_SNAKE_CASE
```
ribbanya: This is not necessarily true. It's pretty inconsistent right now, including in assert messages, but there are definitely cases where split camel case (similar to function names) is preferred.

ribbanya (changes requested):

.github/CONTRIBUTING.md
Jacoby6000: Agreed on all counts. I included this as a carry-over from the original style guide. Happy to remove it!

.github/CONTRIBUTING.md
Jacoby6000: Ah gotcha. I'll find some examples... I'll also add a foreward explaining some naming convention caveats, like when there are assert strings push us towards breaking convention to be closer to the original

.github/CONTRIBUTING.md
Jacoby6000: So, generally speaking, is the table below backward? For anything except bitmasks we should be using the platform-native types? [signed|unsigned] <char, short, int, long, long long>?

My original intuition here was that, because this is a decompilation project, we're more concerned with what the compiler spat out than what the developers may have written, and as such the specific bit width is more important than the platform-native types.  If you were to get things to a place where a modern compiler could build them and you used explicit bit-widths, you'd wind up to something closer to the original than if you use the platform native types.   Maybe this line of thought doesn't have as much value as I thought it did originally. I'm still new to this :D 


From the other thread:

> Sometimes you have to use int, because we currently don't have any typedefs that are int-typed - s32/u32 are signed/unsigned longs, and that can sometimes cause different codegen.

Does this imply that there should not be any hard and fast rules around this? Seems like its up to the compiler whether or not a long or int should be used


As an aside: I don't fully understand why a 32 bit signed long would have any different codegen than a signed 32 bit int. Is that just a compiler quirk? Is there a doc about this somewhere I can read?

.github/CONTRIBUTING.md
Jacoby6000: I always interpret it as meaning "This thing is probably a pointer to something, but I have no clue what to", because the definition is:

```
#ifdef M2CTX
#define UNK_T M2C_UNK
#else
#define UNK_T void*
#endif
```

I (as a newbie) would expect something that is likely a scalar to be written as an s32/u32.  The name of the member (whether or not it is concrete) would then tell me the level of confidence in that type in case it actually was something else.

But, yeah M2C_UNK is `s32`.  It's a bit strange that they map to different types depending on M2CTX.

Based on the [current m2ctx readme](https://github.com/doldecomp/melee/blob/master/tools/m2ctx/README.md), I'm inclined to not go by the M2CTX definition. `s32` seems like an alternate definition in place to make m2ctx happy, while the `void*` definition looks like what it should be interpreted as by a human, generally.

I could go either way on this, just giving an outside perspective

.github/CONTRIBUTING.md
```diff
@@ -47,35 +57,64 @@ pre-commit run --all-files # Run the hook once on existing code
...
+- Structs should include explicit information about offsets
+    - Top-level Struct offsets begin at 0.
+    - Nested structs continue from their parents offset.
+    - Struct members should include the offsets of their members as an inline comment before the type declaration.
+    - Even when the struct member is well understood and named, the offset comment should be retained.
+    - Struct offset comments must all be the same width, using 0s for left-padding.
+- Struct definitions may contain child structs only if their associated members are not pointer references.
+- When a struct members name/purpose is unknown, it should be written as `x<Hex-Offset>`.
+- When a section of a struct is only functioning as padding, it should be written as `x<Hex-Offset>_pad` and use a u8 array.
+- Struct member names should be snake_case.
```
Jacoby6000, line 95: Especially curious about thoughts on these details.  Just to clarify, this codifies the rule that all struct members should always have an inline comment with their offset, even when the struct is well understood, because this should help with reverse engineering functions with a lot of pointer math.

Ths `x<Hex>` convention is somewhat spotty in the codebase, but it is the convention I have found the most useful when going through things, so I put it in here. Want to make sure this is something y'all agree with.

Also, the child-struct/offset/nesting rule on line 83 I also want to make sure I have correct

Jacoby6000:
> If an inline function is shared by multiple TUs but its function body isn't a real symbol, it goes in inlines.h for the most relevant module. An exception is made if that inline function contains an assert which proves its source file.

For functions that wind up defined in inlines.h, we should probably have a separate naming convention associated.

Existing conventions:
* local/non-exported methods are lowerCamelCase with no file prefix: `fooBar`
* exported methods are UpperCamelCase with a file prefix `baz_FooBar`


This means that the methods would be named `inlines_Foo`, 

Using `gr/inlines.h` as a reference, there is a method:

```
static inline int rand_range(int a, int b)
```

I think a good convention would communicate multiple things at the callsite. 
* This is an inline method
* This is a shared method imported from another file

If we were to change the above function to be named
```
gr_randRange(s32 value)
```

Note that it has the _folder_ prefix for the inline, rather than the filename prefix, and it is _lower_ camel case, rather than upper camel case.

ribbanya:
Updated https://github.com/doldecomp/melee/pull/2882#pullrequestreview-4723407304 slightly.

ribbanya:
> For functions that wind up defined in inlines.h, we should probably have a separate naming convention associated.

Maybe. Some of the convention, especially for `it/inlines.h`, are from asserts (like `"It_PKind_Start <= itGetKind(igobj) && itGetKind(igobj) < It_PKind_Terminate"`), so we need to navigate that as well.

It's the same problem as filenames. HAL used all-lowercase for filenames usually, and our convention is more readable split camel-case.

.github/CONTRIBUTING.md
ribbanya: I personally think right-space-aligned `/* +C0 */` should be the standard as it's the nicest to look at. Bitfields are addressed as `/* +C0:7` */ etc.

Members with `x<Hex>` should only be for members without names, not used as a prefix, but yeah, it should be used in that case.

Ideally, structs wouldn't have inline child types. It's easier to reason about where a type comes from and what uses it, if it's not embedded in a larger struct. So whether pointer or not, it should be considered placeholder, imo.

.github/CONTRIBUTING.md
ribbanya: Sometimes `int` can be collapsed to use byte instructions, but this never happens with `long`. It's most important to write what was originally written, which is more likely to be `int` than `s32` in HAL's code at least. But we can't know for sure, so yes, we can't make a hard rule about it.

I think it's a little backwards, yeah. I'd sooner use `short` than `s16`, but `s16` is way more common in the codebase. For strings, `char` should always be used over `u8`, and if it's raw data it should be `u8`. `s8` is kinda rare and in my experience is often an `int` in disguise.

I don't know of any documentation for this.

.github/CONTRIBUTING.md
ribbanya: Yeah, my conception of `UNK_T` vs. scalar aligns with yours.

Jacoby6000:
> Documentation belongs in *.dox and should be kept out of the C and H files with minimal exceptions. @todo comments are allowed anywhere.

I interpret "Documentation" to mean all comments. Does your statement imply a restriction on inline code comments?
I don't usually like inline comments but I can imagine some useful cases, like for complex methods someone took the time to grok.

There is also the occasional priceless commentary that I would hate to forbid
```
    // Unholiest of Permuter slop
    HSD_GObjObject_80390A70(
        data->x4,
        (((((((((((HSD_GObj_804D784B & 0xFFFFFFFFFFFFFFFFu) &
                  0xFFFFFFFFFFFFFFFFu) &
                 0xFFFFFFFFFFFFFFFFu) &
                0xFFFFFFFFFFFFFFFFu) &
               0xFFFFFFFFFFFFFFFFu) &
              0xFFFFFFFFFFFFFFFFu) &
             0xFFFFFFFFFFFFFFFFu) &
            0xFFFFFFFFFFFFFFFFu) &
           0xFFFFFFFFFFFFFFFFu) &
          0xFFFFFFFFFFFFFFFFu) &
         0xFFFFFFFFFFFFFFFFu) &
            0xFFFFFFFFFFFFFFFFu,
        cobj);
```

ribbanya:
"Documentation" refers to doc comments. So `///` and `/**`, not all `//` and `/*`. Comments are fine, and comments which explain the nuances of how an implementation works in line with the code is encouraged.

Jacoby6000:
> >  For functions that wind up defined in inlines.h, we should probably have a separate naming convention associated.
>
> Maybe. Some of the convention, especially for it/inlines.h, are from asserts (like "It_PKind_Start <= itGetKind(igobj) && itGetKind(igobj) < It_PKind_Terminate"), so we need to navigate that as well.
> It's the same problem as filenames. HAL used all-lowercase for filenames usually, and our convention is more readable split camel-case.

Right, for stuff like this there will be a foreward/caveats section explaining that rules are broken sometimes to align with assert/debug strings found embedded in the binary

.github/CONTRIBUTING.md
Jacoby6000: I looked around the codebase.  There really is no rhyme or reason to the enum naming convention as it stands.  Same deal for strings embedded in the binary.

.github/CONTRIBUTING.md
```diff
@@ -71,35 +94,48 @@ pre-commit run --all-files # Run the hook once on existing code
...
-          /* 0x00 */ u32 x0;
-          /* 0x04 */ u8 x4_pad[12];
-          /* 0x10 */ u8 well_known_something;
-          /* 0x11 */ u8 some_bitmask
+      struct Player {
+          /* +C00 */ u32 x0;
+          /* +C04 */ u8 x4_pad[12];
+          /* +C10 */ u8 well_known_something;
+          /* +C11 */ u8 foo_bar;
+          /* +C12 */ u8 x13
```
ribbanya: The `C0` thing was just a random hex number. This is what I meant by right-space-aligned:
```suggestion
          /*  +0 */ u32 x0;
          /*  +4 */ u8 x4_pad[12];
          /* +10 */ u8 well_known_something;
          /* +11 */ u8 foo_bar;
          /* +12 */ u8 x12;
```

.github/CONTRIBUTING.md
```diff
@@ -165,34 +201,49 @@ pre-commit run --all-files # Run the hook once on existing code
...
+Generally speaking, `int` or `unsigned int` should be the preferred type for
+whole numbers, unless there is some specific context about a piece of code or
+struct that dictates a specific bit width. In practice, `int` is more likely to
+be used throughout a codebase, and the goal of this decmopilation project is to
+try to match the original source.
+
+When explicit fixed-width numbers are necessary, use the following types:
+- Booleans: `bool`
+- Bitmasks: Use the smallest unsigned explicit width int type that fits the
+  mask. (`u8`, `u16`, `u32`, `u64`)
```
ribbanya: This is heavily match-dependent and should not have a guideline attached.

.github/CONTRIBUTING.md
```diff
@@ -255,6 +304,56 @@ pre-commit run --all-files # Run the hook once on existing code
...
+- One forward.h corresponds to each types.h and is responsible for providing a
+  typedef for each struct and union, and typedefs of function pointers. It is
+  also the place to define enums, global const simple types like integers that
+  don't get emitted as symbols, and #define constants. The purpose of this
+  separation is to avoid circular inclusions.
+- If an inline function is shared by multiple TUs but its function body isn't a
+  real symbol, it goes in inlines.h for the most relevant module. An exception
+  is made if that inline function contains an assert which proves its source file.
+- Documentation belongs in *.dox and should be kept out of the C and H files
+  with minimal exceptions. @todo comments are allowed anywhere.
```
ribbanya:
```suggestion
- Every TU header (meaning things that actually appear in `splits.txt`) **must** contain its own symbols according to the split, and **only** those members. Extra inlines, types, etc. are not allowed. Symbols must appear in address order. `static` symbols should appear at the top of the C file, not the header.
- A TU may define a `*.static.h` file. This is a temporary file to aid in m2c decompilation runs, so it should be considered as part of the C file, and entirely private. It should be deleted when the TU is linked (marked as `Matching`).
- `types.h` for each module contains relevant types primary to that module. It's difficult to discern where a type truly belongs, so we're not very strict about that. Types should be only `struct X` and/or `union X` definitions, not `typedef`s.
- One `forward.h` corresponds to each `types.h` and is responsible for providing a `typedef` for each struct and union, and typedefs of function pointers. It is also the place to define `enum`s, global `const` simple types like integers that don't get emitted as symbols, and `#define` constants. The purpose of this separation is to avoid circular inclusions.
- If an `inline` function is shared by multiple TUs but its function body isn't a real symbol, it goes in `inlines.h` for the most relevant module. An exception is made if that inline function contains an assert which proves its source file.
- Documentation belongs in `*.dox` and should be kept out of the C and H files with minimal exceptions. `@todo` comments are allowed anywhere.
```

ribbanya (changes requested):

.github/CONTRIBUTING.md
Jacoby6000: Oh, duh. I don't know what I was thinking

.github/CONTRIBUTING.md
```diff
@@ -241,32 +269,34 @@ techinical quirks surrounding various integer types.
...
+        /*  +0 */ CKIND_CAPTAIN,
+        /*  +1 */ CKIND_DONKEY,
+        /*  +2 */ CKIND_FOX,
         ...
-        /* 0x20 */ CHKIND_POPO,
-        /* 0x21 */ CHKIND_NONE,
-        /* 0x22 */ CHKIND_MAX = CHKIND_NONE
+        /* +20 */ CHKIND_POPO,
+        /* +21 */ CHKIND_NONE,
+        /* +22 */ CHKIND_MAX = CHKIND_NONE
```
ribbanya: This should probably be more like the following, cause it's not an "offset" from anything it's just a value:
```suggestion
        /*  0x0 */ CKIND_CAPTAIN,
        /*  0x1 */ CKIND_DONKEY,
        /*  0x2 */ CKIND_FOX,
        ...
        /* 0x20 */ CHKIND_POPO,
        /* 0x21 */ CHKIND_NONE,
        /* 0x22 */ CHKIND_MAX = CHKIND_NONE
```

Decimal would look nicer, but the utility of matching m2c/objdiff output probably outweighs that. Plus `clangd` tooltips give you the decimal value.

.github/CONTRIBUTING.md
```diff
@@ -47,33 +80,90 @@ pre-commit run --all-files # Run the hook once on existing code
...
+          /*   +1E */ short x1E;
+          /*   +10 */ u8 well_known_something;
+          /* +11:0 */ bool enable_rumble: 1
+          /* +11:1 */ bool x13_1: 1
+          /* +11:2 */ bool x13_2: 1
+          /* +11:3 */ bool is_invisible: 1
+          /* +11:4 */ bool x13_4: 1
+          /* +11:5 */ bool is_metal: 1
+          /* +11:6 */ bool x13_6: 1
+          /* +11:7 */ bool x13_7: 1
```
ribbanya: Wrong spacing for bitfields. Also, this byte alignment is not possible on a halfword.
```suggestion
      struct Player {
          /*  +0   */ u32 x0;
          /*  +4   */ u8 x4_pad[0xC];
          /* +10   */ short x10;
          /* +12   */ u8 well_known_something;
          /* +13:0 */ bool enable_rumble: 1
          /* +13:1 */ bool x13_1: 1
          /* +13:2 */ bool x13_2: 1
          /* +13:3 */ bool is_invisible: 1
          /* +13:4 */ bool x13_4: 1
          /* +13:5 */ bool is_metal: 1
          /* +13:6 */ bool x13_6: 1
          /* +13:7 */ bool x13_7: 1
```

ribbanya: Actually one more thing, because of a quirk of m2c it's better for padding to be `char pad_4[0xC]`. Not sure if that's been patched, but this lets it guess at fields when it otherwise wouldn't.

ribbanya: Yeah it's the same as before:
https://github.com/matt-kempster/m2c/blob/6da8dca10a9bffd0fcbd65777528f78703bc2523/m2c/c_types.py#L286-L295

Jacoby6000: Gotcha. I'll make the updates

.github/CONTRIBUTING.md
```diff
@@ -128,18 +128,18 @@ pre-commit run --all-files # Run the hook once on existing code
...
-          /* +11:0 */ bool enable_rumble: 1
-          /* +11:1 */ bool x13_1: 1
-          /* +11:2 */ bool x13_2: 1
-          /* +11:3 */ bool is_invisible: 1
-          /* +11:4 */ bool x13_4: 1
-          /* +11:5 */ bool is_metal: 1
-          /* +11:6 */ bool x13_6: 1
-          /* +11:7 */ bool x13_7: 1
+          /*  +0   */ u32 x0;
+          /*  +4   */ u8 pad_4[0xC];
```
ribbanya: It needs to also be `char` for m2c to recognize it.

ribbanya (changes requested):

.github/CONTRIBUTING.md
Jacoby6000: Ah I thought it would resolve the type through the define

ribbanya: It doesn't, I think because it's a `typedef` which isn't through the preprocessor.

Jacoby6000: There is a typedef resolution phase... Maybe it happens after, or maybe it would recognize an `s8` but not a `u8`.

I wonder how much additional stuff would be found by just correcting this in the codebase. I've definitely seen padding that doesn't follow m2c rules. Very often using u8

ribbanya (approved):

PR #2891, opened 2026-07-17
Document shared shell spin effect spawning

MarkMcCaskey:
These functions are very similar but it's not possible to share code without hacks: we need macros, manually inlined code, or adjusting inline depth pragmas to make it actually shared.


src/melee/it/items/inlines.h
```diff
@@ -18,4 +20,26 @@ static inline void itGamewatch_InitSpawnItem(SpawnItem* spawn,
...
+#define ITEM_SPAWN_SHELL_SPIN_EFFECT(gobj, reset_interval, offset, timer,     \
+                                     pos, jobj, prepare_facing, facing_dir)   \
+    do {                                                                      \
+        (jobj) = GET_JOBJ(gobj);                                              \
+        (pos) = (offset);                                                     \
+        prepare_facing;                                                       \
+        (pos).x *= (facing_dir);                                              \
+        efAsync_Spawn((gobj), &GET_ITEM(gobj)->xBC0, 2, 1029, jobj, &(pos));  \
+        (timer) = (reset_interval);                                           \
+    } while (0)
```
ribbanya: I'm not sure if this is a good precedent to set. On the one hand it's clearly a nightmare to read and to work with. On the other hand, it's a common placeholder that can be fixed up later, so compared to spamming todo comments it might be preferable. @PsiLupan, @r-burns, @BR-?

ribbanya (changes requested):

src/melee/it/items/inlines.h
MarkMcCaskey: Makes sense, perfectly happy to not ship this or not ship this as-is by the way! Refactors should make the code better and if it's unclear if it's helping then best to not ship it.

Perhaps I/we just need a place to document near-matches for code restructuring to revisit later. I was able to get a bigger shared helper where 2/3 matched but the third always had register allocation mismatches. So there probably is a real way to do this but I couldn't figure it out right now.

ribbanya: Let's just go with todo-comments anywhere that function-like macros are the only way to pull out code. Say what code is shared. If the shared code is self-contained in an existing function, just refer to that function in a comment.


If the duplicated code isn't cleanly contained in its own function already, just have all instances of the duplicated code point to a single arbitrary instance so that it's simpler to investigate later. You can also use scope blocks to section it off. Like:
```c
void foo() {
    // some unique code...

    { /// @todo Duplicated code used elsewhere
        // some duplicated code...
    }

    // some more unique code...
}

void bar() {
    // some "bar" code...

    { /// @todo Shared code with #foo
        // some "foo" code...
    }

    // some more "bar" code...
 }

void baz() {

    // some "baz" code...

    { /// @todo Shared code with #foo
        // some "foo" code...
    }

    // some more "baz" code...
}
```

ribbanya (approved):

PR #2893, opened 2026-07-17
Melee decomp 9/11: FT root

fjooord:
## Summary

Decompilation match + standards work for the fighter-root (`ft/`) units from the melee decomp orchestrator session (July 2026): 13 functions newly matched, plus data-section pairing fixes (ftcoll `.sdata2`/`.data`, ftcpuattack `.sdata2`, ftafterimage `.sdata`/`.data`, ftdevice `.bss` via a byte-conserving `symbols.txt` region). All changes passed a standards review/repair pipeline (deterministic review_lint + adversarial LLM review, per-file repair validated by object build + objdiff comparison) and a full-build zero-regression gate at every total/unit/section/function measure.

## Verification

- Rebuilt from scratch on `a079f32198` and verified against a fresh master baseline: 0 broken matches, 0 fuzzy regressions, 0 metric regressions (section existence and exact byte conservation included).
- Passed the upstream `check-issues` lint locally (same container as CI's Issues job) at exact master parity.

## Coordination

- The previously disclosed `ftkirby.c`/`ftparts.c`/`ftparts.h` overlap with the Kirby PR (10/11) resolved itself when #2894 merged: master was merged into this branch and all three files dropped out of the diff.

## Review notes

Unresolved review findings from the session pipeline are posted as comments — those are the spots that need human judgment; everything else passed the automated standards gate.


src/melee/ft/ftcoll.c
```diff
@@ -591,94 +573,128 @@ bool ftColl_80076ED8(Fighter* fp0, HitCapsule* hit0, Fighter* fp1,
...
+                            entry->x0 = 1;
+                            entry->kind = kind;
+                            entry->gobj = gobj;
+                            entry->hit0 = hit0;
+                            entry->hit1 = hit1;
+                            entry->pos = hit0->hurt_coll_pos;
+                            entry->x20 = temp_dmg;
+                            entry->size_of_xC = len;
+                            ++dmg_log1_idx;
+                        } else {
```
fjooord: **`preship_reject_global_standard_header_inlines`**

> The existing `tiplog` helper is deleted and its full body is manually expanded into `ftColl_80076ED8`. The standard requires authored helper boundaries to be restored when duplicated expanded bodies expose them; manual inline expansion is also an evidence-gated matching tactic, and no evidence is recorded.

**Match context:**

- match status unavailable

src/melee/ft/ftcoll.c
```diff
@@ -1176,14 +1200,18 @@ bool ftColl_80077C60(Item* item, HitCapsule* hit, Fighter* fp,
...
             }
-        } else {
+            goto retfalse;
+        }
+
+        {
+            DmgLogEntry* entry;
             bool inner_ret;
+            bool should_log;
             int mode;
```
fjooord, line 1205: **`preship_reject_global_standard_canonical_control_flow_and_macros`**

> The added `goto retfalse` and label at line 1328 preserve generated-style control flow without recorded evidence that structured C was tested and failed. This confirms the `m2c_goto_label` lint finding; its moved-code downgrade is not an exception to the mandatory canonical-control-flow rule.

**Match context:**

- match status unavailable

src/melee/ft/ftcpuattack.c
```diff
@@ -28,6 +28,73 @@
 #include <melee/mp/mplib.h>
 #include <melee/mp/types.h>
```
fjooord, line 30: **`preship_reject_global_standard_literals_and_data_ownership`**

> The added `sdata2_order` function is an explicit data-ordering dodge: it consists entirely of discarded numeric literals and `S32_TO_F32`, with no runtime purpose or recorded section-ownership evidence. It manufactures literal references to influence `.sdata2` ordering rather than reconstructing authored source. This is the same class of regression as the rejected self-TU `.sdata2` anchors in exhibit-2657-grkongo-sdata2-externs.

**Match context:**

- match status unavailable

src/melee/ft/ftcpuattack.c
```diff
@@ -1325,27 +1401,32 @@ void ftCo_800B77E8(Fighter* fp)
                                          target->cur_pos.x, target->cur_pos.y))
         {
             can_attack = 1;
```
fjooord, line 1403: **`preship_reject_global_standard_canonical_control_flow_and_macros`**

> The patch repeatedly replaces ordinary `if`/`else` assignment flow with `do { ... break; ... } while (0)` constructs. This generated-looking single-iteration control flow is introduced for the 22, 62, 102, 58, 57, 27, 53, 16, 60, and 32-unit checks through line 1860. No local evidence establishes this as authored source, so the generated branch shape may not ship.

**Match context:**

- match status unavailable

fjooord:
**`preship_reject_global_standard_literals_and_data_ownership`**

> The newly defined address-style constants `ftColl_804D82EC` and `ftColl_804D8314` promote ordinary `1.0F` and `0.01F` literals into named `.sdata2` storage. The diff records no section or symbol metadata proving source ownership, and later substitutes these symbols into arithmetic. This is a data-ordering dodge.

Also at line: 2173

**Match context:**

- match status unavailable



---

**`preship_reject_global_standard_no_string_literal_symbol_regression`**

> The diff creates three `UNUSED static char` symbols for ordinary assert text and later replaces inline `HSD_ASSERTREPORT` string literals with those symbols. Asset/report/assert strings are required to remain inline absent maintainer-approved data-ownership evidence. The `UNUSED` storage also exposes these as layout-oriented dummy data.

Also at line: 2948

**Match context:**

- match status unavailable

config/GALE01/symbols.txt
```diff
@@ -25970,8 +25970,10 @@ gFtDataList = .bss:0x804598B8; // type:object size:0x84 scope:global data:4byte
 ft_8045993C = .bss:0x8045993C; // type:object size:0x30 scope:global data:4byte
 ft_8045996C = .bss:0x8045996C; // type:object size:0x84 scope:global data:4byte
 lbl_804599F0 = .bss:0x804599F0; // type:object size:0x78
-ft_80459A68 = .bss:0x80459A68; // type:object size:0x30 scope:global data:4byte
+ft_80459A68 = .bss:0x80459A68; // type:object size:0xC scope:global data:4byte
 ...bss.0 = .bss:0x80459A68; // type:label scope:local
+ftDevice_BuryThings = .bss:0x80459A74; // type:object size:0x18 scope:global data:4byte
```
fjooord, line 25990: Possible issue on naming, just flagging

src/melee/pl/pltrick.c
```diff
@@ -256,8 +256,8 @@ void pl_80038144(HSD_GObj* attacker_gobj, HSD_GObj* victim_gobj, s32 x18d4_int,
                     victim_fp->player_id, victim_fp->x221F_b4, ev.x2073);
     }
 
-    if (attack_instance == 0 ||
-        victim_fp->dmg.x18ec_instancehitby != attack_instance)
+    if ((u16) attack_instance == 0 ||
+        victim_fp->dmg.x18ec_instancehitby != (u16) attack_instance)
```
r-burns: this seems like a regression

config/GALE01/symbols.txt
ribbanya: It's correct. The symbol was already named that by humans, it just wasn't in `symbols.txt` for whatever reason (unsplit symbol, presumably).

src/melee/ft/ftcoll.c
```diff
@@ -570,6 +533,25 @@ static inline bool inlineB1(HitCapsule* hit0)
...
+        fp1->dmg.x1838_percentTemp =
+            (float) (fp1->dmg.x1838_percentTemp + dmg);
+        if (var_r24_3 > (int) fp1->dmg.x183C_applied) {
+            fp1->dmg.x183C_applied = var_r24_3;
+        }
+        var_r0_3 = 1;
+    } else {
+        var_r0_3 = 0;
+    }
+    return var_r0_3;
```
ribbanya: Try simplifying to direct return statements.

src/melee/ft/ftcoll.c
ribbanya: Provide evidence.

src/melee/ft/ftcoll.c
ribbanya: Evidence?

src/melee/ft/ftcoll.c
```diff
@@ -2154,50 +2189,56 @@ float ftColl_80079C70(Fighter* fp, Fighter* attacker, HitCapsule* hit,
...
+                               (ftd->x114 * (x118 * (float) (u32) x28)))) +
+                 ftd->x120)) +
+               (float) (u32) hit->x2C)));
     } else {
         s32 count;
-        float damage;
-        float x24_f;
 
-        count = ftColl_GetDamageCount(fp, ftd);
+        if (fp->x2225_b7) {
```
ribbanya, line 2211: Mark expanded inline with a todo comment.

ribbanya (changes requested):

src/melee/ft/ftcoll.c
fjooord: Measured. Restoring the `tiplog` helper is codegen-neutral: with master's `static void tiplog` restored and called from `ftColl_80076ED8`, the function emits the identical 356 instructions (99.69 either way; MWCC fully inlines the call). The expansion was not load-bearing, so the helper boundary is restored in 887f49e4ea. Note MWCC also emits the unreferenced out-of-line copy of the plain-static helper, same as master's build of this TU already does for `getEnvDmg`; unit .text match is unchanged at 99.18.

src/melee/ft/ftcoll.c
fjooord: Measured, load-bearing. `ftColl_80077C60` is exact (100) with the `goto retfalse`. Replacing it with a direct `return false;` drops the function to 99.07 (457 insns; the shared return-false tail no longer lands at the end of the function where the target has it). Master's structured `} else { ... }` shape for this function also mismatches in this TU: 98.62. For context, master's ftcoll.c uses the same `goto retfalse` idiom today in `ftColl_80076ED8` (2 occurrences), and this PR removed both of those while keeping this single one where it is required. Happy to take the 99.07 penalty if you prefer the structured form.

src/melee/pl/pltrick.c
fjooord: You were right — measured it and the whole retype was unnecessary. Reverted in 887f49e4ea: `pl_80038144` is back to master's `u16` signature and pltrick.c/pltrick.h/ftcoll.h are now byte-for-byte master. The retype only existed to serve `ftColl_8007861C`'s exact match (with an `int` local passed to a `u16` param, MWCC inserts a clamp at the call site: 62 -> 65 insns, regalloc shifts, match breaks). Typing that function's own param/local as `u16` instead (matching master's types) avoids the clamp and holds `ftColl_8007861C` at 100 with no cross-module edit at all. pltrick.o code/data sections verified byte-identical before and after.

src/melee/ft/ftcoll.c
fjooord: Done in 887f49e4ea — direct `return true;` / `return false;` verified codegen-neutral (all users unchanged, `ftColl_80076ED8` emits identical bytes, unit scores unchanged).

src/melee/ft/ftcoll.c
fjooord: Done in 887f49e4ea — added `/// @todo Restore expanded inline.` above both expanded formula sites in `ftColl_80079C70`.

src/melee/ft/ftcoll.c
fjooord: Correction to the above — a later pass (8762546412) found a structured form that IS free: the goto/label pair is now a plain `if`/`else` (the item path in the else arm, single fall-through `return false;`) and `ftColl_80077C60` holds 100%. The earlier numbers stand for the other variants (direct `return false;` at the site: 99.07; the flat else-shape: 98.62) — but the shipped tree has no goto and no penalty. The flag was right that no evidence had been recorded; it's recorded now.

src/melee/ft/ftcoll.c
fjooord: Addendum (8762546412): also marked the `ftColl_GetDamageCount` expansion with its own todo, and measured the re-inline while at it — 100 → 99.72, so the expansion stays under the todo for now.

config/GALE01/symbols.txt
fjooord: Thanks for confirming — the entry stays as-is.

ribbanya (dismissed):

ribbanya (approved):

PR #2895, opened 2026-07-17
Melee decomp 11/11: IT items

fjooord:
## Summary

Decompilation match + standards work for the item units (`it/`) from the melee decomp orchestrator session (July 2026): 10 functions newly matched across `itspawn`, `itkyasarin` (unit reaches 100%), `itlinkhookshot`, and `itsamusgrapple`, with `.sdata2` pairing for itspawn and itlinkhookshot restored to 100%. No config changes owed (verified by address scan of the session's config diff against these units' split ranges). All changes passed a standards review/repair pipeline (deterministic review_lint + adversarial LLM review, per-file repair validated by object build + objdiff comparison) and a full-build zero-regression gate at every total/unit/section/function measure.

## Verification

- Rebuilt from scratch on `a079f32198` and verified against a fresh master baseline: 0 broken matches, 0 fuzzy regressions, 0 metric regressions (section existence and exact byte conservation included).
- No headers or out-of-manifest files shipped.
- Passed the upstream `check-issues` lint locally (same container as CI's Issues job) at exact master parity.

## Review notes

Unresolved review findings from the session pipeline are posted as comments — those are the spots that need human judgment; everything else passed the automated standards gate.


src/melee/it/items/itlinkhookshot.c
```diff
@@ -60,6 +60,12 @@ ItemStateTable it_803F6998[] = { {
 
 f32 it_804D6D48;
```
fjooord, line 61: **`preship_reject_global_standard_literals_and_data_ownership`**

> The added `sdata2_order` function is explicitly labeled an `.sdata2 order hack` and contains an otherwise useless `0.0F` literal. This is a source-level data-ordering anchor rather than recovered authored code. Data-ordering dodges are rejected even when they improve text or data matching.

**Match context:**

- match status unavailable

fjooord:
<details>

**`preship_reject_global_standard_literals_and_data_ownership`** at `src/melee/it/itspawn.c:21`

> The newly added `sdata2_order` function is explicitly labeled an `.sdata2 order hack` and references otherwise unused numeric literals solely to influence data ordering. This invents source-level literal references without recorded section-ownership evidence and is the same class of data-ordering dodge as the rejected sdata2 anchors.

Evidence:

> Removing sdata2_order caused six previously exact functions to become non-exact through relocation/data-reference differences. Final restoration returned all six to PASS. Proper .sdata2 ownership or symbol metadata requires an edit outside the queue item's source-only lane.



</details>

src/melee/it/items/itkyasarin.c
```diff
@@ -270,40 +270,43 @@ bool itKyasarin_UnkMotion4_Anim(Item_GObj* gobj)
     return false;
 }
 
+static inline s32 itKyasarin_Randi(s32 range)
```
ribbanya, line 273: Why was this created?

src/melee/it/items/itlinkhookshot.c
ribbanya: It's fine. Should be considered a controlled/acceptable hack.

src/melee/it/itspawn.c
```diff
@@ -20,6 +20,14 @@ ItemPickTable monster;
 ItemPickTable it_804A0E50;
 RandomItemSpawner it_804A0E30;
 
+/// @todo .sdata2 order hack
+static void sdata2_order(void)
+{
+    (void) 4503601774854144.0;
```
ribbanya:
```suggestion
    (void) S32_TO_F32;
```

src/melee/it/itspawn.c
```diff
@@ -241,23 +256,22 @@ void it_8026CB9C(s32* counts, u64 mask, f32 weight)
         it_kind++;
         mask >>= 1;
     }
-    it_804A0E30.x4.size = cnt;
-    it_804A0E30.x4.x4 = HSD_MemAlloc(cnt * 4);
-    it_804A0E30.x4.xC = HSD_MemAlloc(cnt * 4);
+    spawner->x4.size = cnt;
+    *(item_kinds = &spawner->x4.x4) = HSD_MemAlloc(cnt * 4);
+    *(weights = &spawner->x4.xC) = HSD_MemAlloc(cnt * 4);
```
ribbanya, line 263: Suspicious pointer math here and L310. What is the justification?

ribbanya (changes requested):

src/melee/it/items/itlinkhookshot.c
fjooord: Thanks — keeping the anchor with its `@todo` marker per this ruling.

src/melee/it/items/itkyasarin.c
fjooord: It reproduces retail's argument materialization: with the wrapper inlined at the four call sites, MWCC emits the exact retail instruction stream for `itKyasarin_UnkMotion4_Coll`/`itKyasarin_UnkMotion5_Coll` (the unit is at 100%). We measured deleting it and calling `HSD_Randi` directly — the emitted `.text` changes (same size, different stream), breaking the exact matches. Happy to take the penalty if you'd rather have the direct calls.

src/melee/it/itspawn.c
fjooord: Applied in 57bf90f4c0 (with the `placeholder.h` include) — object sections byte-identical.

src/melee/it/itspawn.c
fjooord: These are combined assign+store spellings, not reads into foreign data — `item_kinds`/`weights` point at the TU's own table fields (`it_804A0E30`/`it_804A0E50`) and feed the fill loops below. We measured both plain spellings at both sites (`item_kinds = &field; *item_kinds = alloc;` and `field = alloc; item_kinds = &field;`): each changes the emitted `.text` (same size, different instruction stream); `it_8026CB9C` is an exact match and `it_8026CD50` is at 99.95%, so the current form is load-bearing. Can take the penalty if you prefer the plain spelling.

ribbanya: It's fine.

src/melee/it/items/itkyasarin.c
ribbanya: What changes to the stream?

fjooord: Concretely (`itKyasarin_UnkMotion4_Coll` prologue, with-wrapper = retail vs direct `HSD_Randi`):

```
- stwu r1,-88(r1)      ; retail frame 0x58
+ stwu r1,-64(r1)      ; frame shrinks to 0x40
- mr   r31,r3          ; gobj kept in r31, saves ordered r31,r30,r29,r28
+ mr   r28,r3          ; gobj moves to r28, save order changes
- lwz  r29,44(r3)      ; ip in r29
+ lwz  r30,44(r3)      ; ip in r30
```

…and the reallocation (r29→r30, r28→r29, r30→r31) plus shifted spill offsets (`stfd f0,64(r1)` → `40(r1)`, etc.) ripple through both `_Coll` bodies. The inlined wrapper's call expansion is what claims the extra stack words and pins the allocator to retail's assignment; without it MWCC picks the smaller frame + different registers, so every instruction that touches those registers/offsets mismatches.

ribbanya: Ok just say 'regswap' next time. Keep the improved version but add a todo comment above the redefinition.

ribbanya (approved):

PR #2905, opened 2026-07-18
Refactor CPU code for lots of code reuse

MarkMcCaskey:
Removing some of the dontinline helpers is a good signal that some of the new code structure is perhaps more faithful to the original. There are lots of other refactors in various other areas too.


MarkMcCaskey:
Investigating more similar clean ups in this file so back in draft mode.

MarkMcCaskey:
There's definitely a lot more in this file and I'll leave codex running a bit but I need to step away for a while. Feel free to merge if the changes seem reasonable

ribbanya (dismissed):

MarkMcCaskey:
Okay calling it here at -1000 lines. Some of these helpers are more helpful than others. There were a lot of near misses here for large semantically meaningful helpers that were too large to be inlined or had other issues. I think this is definitely worth time to clean this code up more in the future.

ribbanya (approved):

PR #2913, opened 2026-07-19
Check for completed but unlinked units in CI

ribbanya:


r-burns (dismissed):
This is erroring out on a case that is nontrivial, and I don't see why this needs to be an error anyway

ribbanya:
Is it non-trivial? `master` has only one such case and I'm almost finished with it, I just got side-tracked by this PR. I don't really see a reason not to automate the check, since people are frantically checking the wiki for it anyway.

PR #2917, opened 2026-07-19
Share Corneria and Venom Arwing code + Improve Corneria matches

MarkMcCaskey:
From Codex:
> Introduce a common Star Fox ground-var view and typed inline helpers for four duplicated Arwing stage-object callbacks. Decode Venom's raw table indexing while preserving all four functions as 100% matches.


src/melee/gr/stage.h
```diff
@@ -24,6 +24,41 @@
 /* 224B50 */ float Stage_GetBlastZoneLeftOffset(void);
 /* 224B68 */ float Stage_GetBlastZoneTopOffset(void);
 /* 224B80 */ float Stage_GetBlastZoneBottomOffset(void);
+
+static inline bool Stage_IsOutsideBlastZone(Vec3* pos)
```
MarkMcCaskey, line 28: These helpers didn't seem usable in other maps. Seemingly because these check all 4 directions and most stages don't do that

src/melee/gr/grcorneria.c
```diff
@@ -869,20 +869,29 @@ void grCorneria_801DDE88(HSD_GObj* gobj)
...
+        {
             lbAudioAx_800237A8(0x55731, 0x7F, 0x40);
             gp->gv.arwing.xD8 = 1;
         }
         break;
     case 1:
-        if (pos.x < -10.0f || 2000.0f < pos.x) {
+        if (pos.x < *(volatile const f32*) &grCn_ArwingMinX ||
+            *(volatile const f32*) &grCn_ArwingMaxX < pos.x)
+        {
```
MarkMcCaskey, line 910: This code is kind of crazy but it's the result of Codex iterating on this for hours

ribbanya: It's slop.

ribbanya: To elaborate a little, I could believe that it might be marked as `volatile`, but at the very least the decorators are totally unnecessary because MWCC will place any `const f32` in `.sdata2` no matter what.

ribbanya: And if there is a de-duplicated literal, that indicates a missing split, not these gymnastics.

src/melee/gr/grcorneria.c
```diff
@@ -2808,6 +2812,18 @@ DynamicsDesc* grCorneria_801E2EE4(enum_t arg)
...
+#ifdef __MWERKS__
+#define GRCN_SDATA2 __declspec(section ".sdata2")
+#else
+#define GRCN_SDATA2 __attribute__((section(".sdata2")))
+#endif
+/// These are distinct symbols in the DOL. Volatile reads below prevent MWCC
+/// from folding them back into anonymous literals in a different order.
+GRCN_SDATA2 const f32 grCn_804DB260 = 107.0f;
+GRCN_SDATA2 const f32 grCn_804DB264 = 106.0f;
+GRCN_SDATA2 const f32 grCn_804DB268 = -3.4028235e38f;
```
ribbanya: Can you not just declare them as `volatile`?

MarkMcCaskey: At least when I last tried declaring the const volatile and accessing it directly did not match, but maybe declaring the const volatile and accessing them another way could

ribbanya:
To be clear I'm not saying we can't merge the slop. It's just probably not as sensible as it could be while still matching.

MarkMcCaskey:
Only `grCorneria_801E25C4` left in the TU. I'm going to leave Claude iterating on it but it seems tricky. The function is not big and not having a lot of luck so far. If that fails, can just make a decomp.me pad with it and ship this. Can probably clean up some of the hacks used to get matches here either before or after merging.

MarkMcCaskey:
From Claude:
> The permutation is now fully understood, but not cleanly fixable. The sweep (120 declaration permutations plus ~60 structural variants) decoded MWCC's coloring rule: callee-saved ranks are param homes first (r4's home before r3's), then user locals in declaration order, with a pair inversion when one def nests inside another's statement. The target order [v, joints, i, joint0, joint1, gobj] corresponds exactly to declaring the owner copy last — but a plain copy of the r3 param declared that late flips MWCC's home-coloring knob and wraps that web to r31 instead (the "anomaly" that defeated all previous sweeps was this rotation).
>
> I got within 8 instructions-rows of perfect: an opaque chained copy go = (tmp = gobj); in the caller plus Codex's wide setup inline reproduces the exact target permutation, the full 93-instruction stream, and no tail split. The last wall is intrinsic to the trick: the chain's dead inner aliases r3 and reserves it until the first call, so the preheader scratch registers shift off r3 and the copy spells mr instead of addi. Making the inner live releases r3 — but only in the flat structure, where MWCC then splits joint0/joint1 into the freed loop registers after the loop. The two requirements are mutually exclusive in every source shape I could construct, and the chain is an unshippable match hack regardless. So E25C4 is terminal for clean source — but it's now terminal with a complete mechanistic explanation rather than a shrug, all recorded in CLAUDE.md and memory for the decomp.me kit if you want the community to take the last swing.

I made a decomp.me link here https://decomp.me/scratch/IsulI

PR #2919, opened 2026-07-19
Share Kongo barrel motion helpers

MarkMcCaskey:
From Codex:
> Reuse common symmetric clamping and angle wrapping in Kongo and Old Kongo's large barrel callbacks. Reshape the existing Old Kongo inline body around the complete deceleration operation; both callbacks remain 100% matches.


src/melee/gr/inlines.h
```diff
@@ -54,6 +54,27 @@ static inline void Ground_InitTargetStage(HSD_GObj* (*create_gobj)(int) )
...
+        }
+    }
+}
+
+static inline void Ground_WrapAngle(f32* angle)
+{
+    if (*angle > 6.283185307179586) {
+        *angle -= 6.283185307179586;
+    } else if (*angle < -6.283185307179586) {
+        *angle += 6.283185307179586;
```
ribbanya: Please use `M_TAU` from `math.h`.

ribbanya (changes requested):

src/melee/gr/groldkongo.c
```diff
@@ -234,7 +234,6 @@ bool grOldKongo_8020F880(Ground_GObj* gobj)
 }
 
 #define DegToRad(a) ((a) * 0.017453292F)
```
ribbanya: Remove this too.

ribbanya (changes requested):

ribbanya (approved):

PR #2928, opened 2026-07-20
Replace `gv` and `fv` with `u` according to `Ground` asserts

ribbanya:


ribbanya (commented):
Manually scanned all the changes, didn't catch any false positives.

PR #2936, opened 2026-07-21
Clarify Star Fox Arwing behavior + name magnifier

MarkMcCaskey:
Probably worth double checking some of these names with science in the running game. I haven't set that up yet but seems worthwhile. Though looking through the diff, most of these renames seem reasonably well supported by the code structure.


MarkMcCaskey:
**AI disclosure:** This comment was posted by OpenAI Codex at Mark's request.

I wrote a detailed evidence note for the naming and less-obvious structural choices in this PR, with citations to the code on the PR's `master` base:

https://gist.github.com/MarkMcCaskey/aacf303e24e51ed6bf894f995d3a2df1

~Mark has not had a chance to review this document yet~. It should be treated as an AI-generated review aid, and any conclusions in it are open to correction.

Looks good at a high level to me. But I wouldn't be able to catch subtly wrong conclusions in this reasoning anyway.

ribbanya (commented):
I would prefer if a human went through in Dolphin and verified these things. I'm not personally comfortable slapping an "Approve" review on this without verifying, which would take hours I don't wish to commit. We need to develop a more streamlined process for vibe naming passes like this.

PsiLupan:
I'm suspect of the MagnifyColors. Partly because the ifMagnify skipping about randomly in the calls, and partly because there's only 6 slots in a given match. 

It seems like the kind of thing an AI would hallucinate because of later Smash games have 8 player matches.

MarkMcCaskey:
> I would prefer if a human went through in Dolphin and verified these things. I'm not personally comfortable slapping an "Approve" review on this without verifying, which would take hours I don't wish to commit. We need to develop a more streamlined process for vibe naming passes like this.

For sure! Sorry for sending big and hard to review PRs. Happy to close this / split out anything easier to confirm here.

No pressure to merge anything you don't feel comfortable with and I appreciate the time you've already given to reviewing my PRs!

I'll be more cognizant of the review burden going forward -- thanks!

ribbanya:
Well, the problem is that I think parts of this are useful and probably accurate but I have no way to be sure. Honestly this kind of thing would be a good candidate for modders and enthusiasts who want to help and have time and are familiar with Dolphin, but don't know how to code.

MarkMcCaskey:
> Well, the problem is that I think parts of this are useful and probably accurate but I have no way to be sure. Honestly this kind of thing would be a good candidate for modders and enthusiasts who want to help and have time and are familiar with Dolphin, but don't know how to code.

Sure but I don't want to dump hours of work on you. This PR can be split up into reviewable chunks and spread out over a longer period of time.

As mentioned the interface changes here are a bit suspect (they actually got renamed twice 😬) so I think it's worth breaking this down.

But yeah, let's find a better way to do this because I completely agree with you and I don't want to add any unnecessary stress or burn you out. I deal with massive AI PR reviews at work and I know it sucks, so I'm genuinely sorry for doing it here!

PR #2937, opened 2026-07-21
mndiagram: match 3 functions via named HSD_ASSERT strings

Richard7840:
Redefine HSD_ASSERT around the baselib/jobj.h include so the inlined jobj asserts reference named .sdata strings (mnDiagram_804D4FA8 "jobj.h", mnDiagram_804D4FB0 "jobj") instead of anonymous pool entries, matching retail. Closes mnDiagram_CursorProc, mnDiagram_DrawCellValue and mnDiagram_DrawNameHeaders byte-exact, and names the two .sdata symbols. Same pattern already used in mnvibration.c.

### DIFF
--- a/src/melee/mn/mndiagram.c
+++ b/src/melee/mn/mndiagram.c
@@ -3,6 +3,19 @@
 #include "gm/gm_1A36.h"
 #include "mn/types.h"
 #define MNDIAGRAM_SOURCE
+
+#include "baselib/debug.h"
+extern char mnDiagram_804D4FA8[];
+extern char mnDiagram_804D4FB0[];
+#undef HSD_ASSERT
+#define HSD_ASSERT(line, cond)                                                \
+    ((cond) ? ((void) 0)                                                      \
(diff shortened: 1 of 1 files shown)


Richard7840:
Note on the decomp-dev **No changes** report — this is expected, same as the recent "Name and reuse … accessors" PRs (#2874, #2873, #2872). The three functions were already byte-identical to retail; the only divergence was that their inlined `jobj.h` `HSD_ASSERT` relocations pointed at anonymous `.sdata` pool strings instead of the named `mnDiagram_804D4FA8` ("jobj.h") / `mnDiagram_804D4FB0` ("jobj") symbols. This PR names those symbols via the `mnvibration.c` `#undef HSD_ASSERT` pattern, so objdiff now reports the functions as fully matched even though no bytes move.

ribbanya:
Thanks for the attempt, but this is a regression. 'Anonymous pool strings' are preferred to named definitions.

PR #2939, opened 2026-07-22
Fix external stage id docs

MarkMcCaskey:
First PR made with agentic runtime inspection, testing and validating the ideas with code injection using @Jacoby6000's https://github.com/Jacoby6000/dolphin/blob/feature/dap-server/Tools/dap/README.md tool.

Claude's more verbose justification for the details here https://gist.github.com/MarkMcCaskey/3645a0880e7293d3518714d686bdd2b5

I found this by accident by automating loading into a stage and we found out there's a mismatch between external and internal ids. This PR is a deeper dive into that concept to clean up the code here.


ribbanya (commented):
Is there anywhere in the DOL that explicitly maps these? Or just runtime checks?

PsiLupan:
The cutscene mapping seem like sufficient evidence to me at a glance.

MarkMcCaskey:
> Is there anywhere in the DOL that explicitly maps these? Or just runtime checks?

The internal vs external mismatch was identified with code injection to load a specific map and the internal ID didn't line up. The enum was built out by trying all the IDs with the load map code and then reading back memory of a known location for the internal id. So not sure if this is statically encoded but this mapping was created by dynamically editing and reading back memory.

I wonder if it'd be helpful to distill experiments into an easily reproducible form. I think dolphin has something called gecko codes or something and that might be a better way to be able to store this logic for the longer term. Perhaps we could just put that in an emulator and confirm rather than relying on layers of external tools like this.

Jacoby6000:
The code injection tools in the dap branch are mainly there for doing stuff during runtime, where it may be useful to be able to change execution in the middle of a running game, avoiding having to restart every time you change something. 

Gecko would be great for reproducibility I think. Gecko with some EXI and an accompanying launch script that collects the output.

ribbanya:
Gecko codes just overwrite RAM with arbitrary data, which can be assembly code:
https://wiigeckocodes.github.io/codetypedocumentation.html#C0

MarkMcCaskey:
> The code injection tools in the dap branch are mainly there for doing stuff during runtime, where it may be useful to be able to change execution in the middle of a running game, avoiding having to restart every time you change something. 
> 
> 
> 
> Gecko would be great for reproducibility I think. Gecko with some EXI and an accompanying launch script that collects the output. 

Yeah I think the tool you built is ideal for exploration, I'd just want the agent to finish with a self-contained easily reproducible experiment artifact I think

MarkMcCaskey:
Okay I set up a way to reproduce and verify the finding. Gecko code here https://github.com/MarkMcCaskey/melee-runtime-audits/blob/main/pr2939/kiosk.gecko.ini use left/right dpad to advance the stages or press A to autoplay.

selected_stage = .sdata:0x804D49E8 can be monitored as the external ID (displayed in game with devtext as ext)
The internal ID can also be monitored at 0x8049E750 (stage_info = .bss:0x8049E6C8 + 0x88) details [here](https://github.com/MarkMcCaskey/melee-runtime-audits/blob/main/pr2939/README.md#deriving-the-read-addresses-from-the-dol)

There's also a python script for headless verification but the interactive gecko code + dolphin memory inspector is probably the best way to verify.

MarkMcCaskey:
Happy to provide a more in-depth repro or cover anything else, just let me know!

lukechampine (approved):

PR #2945, opened 2026-07-23
Add mechanical clean up of Star Fox stages

MarkMcCaskey:
Splitting #2936 into reviewable and manageable chunks. This PR just removes unnecessary casts and simplifies boolean expressions into a more idiomatic / natural form where doing so doesn't affect the match; purely mechanical.

From Codex
> Drop casts where the source and destination already use compatible object pointer types in Corneria and Venom. Use Venom's leading joints member for the StageData payload because a bare struct pointer makes MWCC abort, while the member expression preserves the exact relocation.

### DIFF
--- a/src/melee/gr/grcorneria.c
+++ b/src/melee/gr/grcorneria.c
@@ -240,7 +240,7 @@ StageData grCn_803E1F08 = {
     grCorneria_801E2EE4,
     grCorneria_801E2EEC,
     0x00000001,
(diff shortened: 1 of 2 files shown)


MarkMcCaskey:
It looks like I don't have the ability to tag this PR myself but it is done with AI

MarkMcCaskey:
Actually let me roll in the mechanical boolean / type updates as well

PR #2947, opened 2026-07-23
Finish matching and link Flat Zone

MarkMcCaskey:
From Codex:
> Use direct Ground GObj field access to recover the retail inline stack layout in grFlatzone_802174EC and grFlatzone_802176BC. Preserve the expected random animation result register by incrementing it separately.

For linking, from Claude:
> All of grflatzone's data was already decompiled; this updates symbols.txt to mirror the compiled object (local scopes, MWCC @N literal names, real sizes), removes the stale commented-out assert-string definitions (the compiler emits "jobj.h"/"jobj" from the jobj.h inline asserts naturally), and marks the object Matching. All sections verified byte-identical against the extracted object.


MarkMcCaskey:
If there are any issues with the linking but we want to keep the matches feel free to drop that final commit and take the matching commit

PR #2954, opened 2026-07-24
Match it_8026CD50

MarkMcCaskey:
From Codex:
> Preserve the row pointer and copied display value lifetimes in fn_802FF218 so MWCC reproduces the target callee-saved allocation.


MarkMcCaskey:
I ran Codex and Claude concurrently and it looks like Codex got the same matches as Claude but more slowly. Closing this PR as a dupe

PR #2956, opened 2026-07-24
Match ftCo_CaptureWaitHi_Coll and ftCo_CaptureWaitLw_Coll

TheTom:
## Summary

Replaces incorrect `ftCo_CapturePulledHi/Lw_Coll` stubs in the CaptureWait coll handlers with the real collision bodies used by sibling capture states in the same TU.

### What matched

- `ftCo_CaptureWaitHi_Coll` (0x800DBA8C)
- `ftCo_CaptureWaitLw_Coll` (0x800DBE9C)

### Why this is correct

Both wait colls follow the established capture-coll pattern elsewhere in `ftCo_Attack100.c`:

1. Load the fighter.
2. Skip if `x2226_b2` is set.
3. Call `ft_80083C00` (hi) / `ft_8008403C` (lw) with the wait-specific landing callbacks (`fn_800DBAC4` / `fn_800DBED4`).

The previous implementations simply called the CapturePulled colls, which do not match the target code.

### How verified

```text
export WINEPREFIX="$HOME/.wine-melee-pr" WINEDEBUG=-all
ninja build/GALE01/src/melee/ft/chara/ftCommon/ftCo_Attack100.o
build/tools/objdiff-cli diff \
  -1 build/GALE01/obj/melee/ft/chara/ftCommon/ftCo_Attack100.o \
  -2 build/GALE01/src/melee/ft/chara/ftCommon/ftCo_Attack100.o \
  -o - --format json
```

Result: both functions report `match_percent` 100.0 (size 0x38 each).

TU remains NonMatching (not flipped).

### Notes

- Legal US 1.02 `main.dol` is not included in this PR.
- No global field renames, no data-section matching, no Matching flag changes.

> AI assistance was used for review, patch extraction, and objdiff verification. The coll bodies match the existing human-written sibling pattern in this file.

PR #2960, opened 2026-07-24
Match gm_80166378

TheTom:
## Summary

Matches `gm_80166378` (match end / player standings fill) at 100%.

### What matched

- `gm_80166378` (size 0x714 / 1812)

### Why this is correct

MWCC was not coloring the loop-invariant float constants the same way as the target. The fix:

1. Declares locals in reverse register order: `f32 d, c, b, a` so assignment coloring lands on `f27`..`f30` (with the existing stack bias on `f31`).
2. Materializes each constant via assign-exprs (`a = 0.031f`, `b = 100.0f`, `c = 60.0f`, `d = 10.0f`) and reuses `c`/`d` for the second scale so both multiplies share the same FPRs as the original.

Semantics are unchanged; this is pure FPR allocation control.

### How verified

```text
export WINEPREFIX="$HOME/.wine-melee-pr" WINEDEBUG=-all
ninja build/GALE01/src/melee/gm/gm_1601.o
  -1 build/GALE01/obj/melee/gm/gm_1601.o \
  -2 build/GALE01/src/melee/gm/gm_1601.o \
  -o - --format json
```

Result: `gm_80166378` reports `match_percent` 100.0 (size 1812).

TU remains NonMatching (not flipped). Independent of #2956 / #2957 / #2959 (branched from current `upstream/master` @ `607b232ac`).

### Notes

- Legal US 1.02 `main.dol` is not included in this PR.
- No global field renames, no data-section matching, no Matching flag changes.
- Touch: `src/melee/gm/gm_1601.c` only.

> AI assistance was used for FPR coloring exploration, patch application, and objdiff verification. No new globals or field names were introduced.

### DIFF
--- a/src/melee/gm/gm_1601.c
+++ b/src/melee/gm/gm_1601.c
@@ -3101,7 +3101,8 @@ void gm_80166378(lbl_8046B6A0_24C_t* arg0_raw)
             sp48_y = (a = 0.031f) * ((f32) cnt * fn_8016B5B0());
             fn_80166A8C((Vec3*) &sp48_y, (Vec3*) &sp48_x);
             arg0->player_standings[i].xE = *(u16*) &sp48_x;
-            arg0->player_standings[i].x34 = (s8) ((b = 100.0f) * pl_80040948(i));
+            arg0->player_standings[i].x34 =
+                (s8) ((b = 100.0f) * pl_80040948(i));
             arg0->player_standings[i].x38 = pl_80040900(i);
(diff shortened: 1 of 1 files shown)


ribbanya (approved):

TheTom:
Rebased onto current master so this no longer regresses the already-merged `fn_802B805C` match (that broken-match report was just the branch being behind). Match for `gm_80166378` is unchanged.

PR #2965, opened 2026-07-24
Match fn_80190ABC

TheTom:
## Summary

Matches `fn_80190ABC` (tournament HUD text refresh) at 100% code match.

### What matched

- `fn_80190ABC`

### Why this is correct

Upstream case 5 was a register-allocation tie-break (shared `@todo`). The match:

1. Hoists `u8* start` so cases 5 and 6 share the same base for `state->x3`.
2. Reuses the existing `table` local for `&state->x2` in case 5 so that pointer colors correctly.
3. Ends case 5 with a dummy `switch (state->x4) { default: break; }` so `state` stays live and the compiler does not fold `start` into an update-form load. (Load-bearing: removing it drops match to ~97.7%.)

Semantics unchanged; no new globals.

### How verified

```text
ninja build/GALE01/src/melee/gm/gmtou_0.o
  -1 build/GALE01/obj/melee/gm/gmtou_0.o \
  -2 build/GALE01/src/melee/gm/gmtou_0.o \
  -c functionRelocDiffs=data_value \
  fn_80190ABC
```

Result: `match_percent` 100.0 under `functionRelocDiffs=data_value`.

TU remains NonMatching. Single-file, independent of other open PRs.

### Notes

- Legal US 1.02 `main.dol` is not included.
- No global field renames, no data-section matching, no Matching flag change.
- Touch: `src/melee/gm/gmtou_0.c` only.

> AI assistance was used for the case-5 live-range pins and objdiff verification. No new globals or field names were introduced.

### DIFF
--- a/src/melee/gm/gmtou_0.c
+++ b/src/melee/gm/gmtou_0.c
@@ -138,7 +138,6 @@ typedef void (*lbl_803D9FD8_fn)(s32*, u32, u32);
     fn_80195CCC, fn_80194F30, fn_8019610C, fn_8019610C,
 };
 
-/// @todo Only differs by a register allocation tie-break in the last case.
 void fn_80190ABC(int mode)
 {
     struct Lbl804799B8_t* state = &lbl_804799B8;
@@ -146,6 +145,8 @@ void fn_80190ABC(int mode)
     TmData* tm;
     s32 cur_opt;
     s32 opt;
+    /* Shared by cases 5/6 for matching base register coloring. */
+    u8* start;
 
     tm = gm_GetTournamentData();
     cur_opt = tm->cur_option;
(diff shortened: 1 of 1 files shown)


ribbanya (approved):

ribbanya:
@TheTom Thanks for the contributions, nice matches!

You don't need to merge `master` into your PR branches. I can just do that myself. Each commit to `master` creates a new merge head, so if you merge master into multiple PRs at once it just creates a bunch of empty merges that have to re-run the CI workflow.

TheTom:
Thanks @ribbanya! Good to know. I'll stop merging master into the PR branches and leave that to you so we don't thrash CI with empty merges.

TheTom:
No problem at all @ribbanya, and thank you for the guidance. We'll leave master merges on open PRs to you from here on.

PR #2968, opened 2026-07-24
Match ftCo_800B4AB0

TheTom:
## Summary

Matches `ftCo_800B4AB0` (CPU attack selection) at 100% code match.

### What matched

- `ftCo_800B4AB0`

### Why this is correct

1. Pin MSL `sqrtf` expansion stack temps via `sqrtf_store` + `sqrt_tmp[4]` so each call site owns the retail 4-byte slots.
2. Array-only weight selection for induction coloring.
3. Stage upper/lower multiplies for FPR order.

No new globals. Semantics unchanged.

### How verified

```text
ninja build/GALE01/src/melee/ft/ftcpuattack.o
  -1 build/GALE01/obj/melee/ft/ftcpuattack.o \
  -2 build/GALE01/src/melee/ft/ftcpuattack.o \
  -c functionRelocDiffs=none \
  ftCo_800B4AB0
```

Result: `match_percent` 100.0 (instruction stream). Under `functionRelocDiffs=data_value` one SDA21 symbol-name display differs (`ftCo_804D3B70` vs local `@479`) with identical `li r5,0` encoding.

TU remains NonMatching. Single-file, independent of other open PRs.

### Notes

- Legal US 1.02 `main.dol` is not included.
- No global field renames, no data-section matching, no Matching flag change.
- Touch: `src/melee/ft/ftcpuattack.c` only.

> AI assistance was used for the sqrtf stack-slot pins, weight-selection regalloc, and objdiff verification. No new globals or field names were introduced.

### DIFF
--- a/src/melee/ft/ftcpuattack.c
+++ b/src/melee/ft/ftcpuattack.c
@@ -140,6 +140,22 @@ typedef struct ftCo_CollData {
 } ftCo_CollData;
 
 /// @todo Fake helper forcing a fresh load of the scale field.
+
+/// MSL sqrtf expansion writing through a caller-provided stack slot so each
+/// call site owns a distinct 4-byte temp (retail 0x28..0x34 below sp3C).
+static inline float sqrtf_store(float x, volatile float* y)
+{
+    if (x > 0.0f) {
+        double guess = __frsqrte((double) x);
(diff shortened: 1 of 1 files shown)


ribbanya (approved):
It's actually 100% with `--reloc-diffs name_address`.

PR #2971, opened 2026-07-25
Fix the lbshadow bezier constant and lay out its .sdata2 pool

MarkMcCaskey:
I'm going to keep grinding on this and see if I can get matches here. 

From Claude:
> The cubic bezier tangent wrote bez1 as (-4.0f * t) + 1.0f, but retail's .sdata2 holds +4.0f and contains no -4.0f at all, and the target emits fnmsubs where the negated form emits fmadds.  Writing it as 1.0f - (4.0f * t) restores both, and drops the pool from eight constants to retail's seven.
>
> MWCC orders the literal pool by first appearance in the source, so the recovered constant alone reshuffles it.  Add the usual sdata2_order anchor - as itzako, grrcruise and friends already do - to pin the pool to retail's order.  .sdata2 goes from 86.79% to 94.34%; the remainder is a second 0.0f and 1.0f that retail holds separately, which one translation unit cannot reproduce.
>
> lbShadow_8000E9F0 keeps one two-instruction scheduling transposition (fnmsubs against the t2 multiply); every other function and .data are byte-identical to before.


MarkMcCaskey:
Closing this as it's a bit of a dead end for now

PR #2973, opened 2026-07-25
Match mnItemSw_802351A0

TheTom:
## Summary

Matches `mnItemSw_802351A0` in `src/melee/mn/mnitemsw.c` at 100%.

### What matched

- `mnItemSw_802351A0` (size 0x720 / 1824)

### Why this is correct

The cursor-position setup and the user-data initialisation each need to be emitted
as a unit before the surrounding code touches the same registers. Pulling them into
two local static inlines gives that grouping without changing semantics, and lets
the remaining body keep retail's register assignment.

### How verified

```text
python3 configure.py && ninja
python3 configure.py progress
```

Against current `master` this goes from 19433 to 19434 matched functions and
3268728 to 3270552 matched code bytes (+1824, the exact size of this function).
`report.json` reports `mnItemSw_802351A0` at 100.0.

### Notes

- Touches `src/melee/mn/mnitemsw.c` only; no header, struct or global changes.
- No data-section matching; the TU stays `NonMatching`.
- `clang-format` clean. New static inlines use lowerCamelCase per the style guide.
- Branched directly off `master` with no merge commits, per the note on #2965.

> AI assistance was used for the inline grouping and objdiff verification. No new globals or field names were introduced.

### DIFF
--- a/src/melee/mn/mnitemsw.c
+++ b/src/melee/mn/mnitemsw.c
@@ -836,11 +836,8 @@ HSD_GObj* mnItemSw_802351A0(s32 arg0)
     HSD_JObjReqAnimAll(jobj, 0.0f);
     HSD_JObjAnimAll(jobj);
 
-    user_data = HSD_MemAlloc(sizeof(MnItemSwData));
-    if (user_data == NULL) {
-        OSReport("Can't get user_data.\n");
-        __assert("mnitemsw.c", 0x3D7, "user_data");
-    }
+    user_data = HSD_MemAlloc(sizeof(*user_data));
+    HSD_ASSERTREPORT(983, user_data, "Can't get user_data.\n");
 
     GObj_InitUserData(gobj, 0, HSD_Free, user_data);
 
--- a/src/melee/mn/mnitemsw.c
+++ b/src/melee/mn/mnitemsw.c
@@ -756,17 +756,14 @@ HSD_JObj* mnItemSw_80235020(u8 arg0, MnItemSwData* arg1)
     return jobj;
 }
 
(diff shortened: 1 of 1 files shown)


src/melee/mn/mnitemsw.c
```diff
@@ -77,10 +77,22 @@ static struct MnItemSwAnimTable mnItemSw_AnimTable = {
...
-static u8 mnItemSw_ItemOrder[32] = {
-    0x05, 0x12, 0x0A, 0x1E, 0x0D, 0x18, 0x03, 0x0E, 0x17, 0x1B, 0x01,
-    0x09, 0x08, 0x07, 0x15, 0x04, 0x06, 0x02, 0x0F, 0x00, 0x11, 0x0B,
-    0x1F, 0x1A, 0x14, 0x19, 0x10, 0x16, 0x13, 0x1D, 0x0C, 0x00,
+struct MnItemSw_803ED438 {
+    /* 0x00 */ u8 item_order[32];
+    /* 0x20 */ char cant_get[0x18];
+    /* 0x38 */ char file[0x0C];
+    /* 0x44 */ char user_data_name[0x0C];
+};
```
ribbanya: This pattern is a regression.

ribbanya (changes requested):

TheTom:
Good catch, thanks. Small oversight on my part. That struct renames the existing symbols and matches .data, neither of which it should be doing. Fix incoming.

TheTom:
Updated. Dropped the struct and the .data changes, kept the existing names. Still 100% on the function, and the data sections are untouched now.

ribbanya (approved):

PR #2976, opened 2026-07-25
Reuse some existing inline helpers more

MarkMcCaskey:
Reusing existing inline helpers in more places.

See commits for info about each change.


ribbanya (changes requested):
What happens if you try to force usage of the new helpers by dismantling inline functions at the would-be call sites?

MarkMcCaskey:
> What happens if you try to force usage of the new helpers by dismantling inline functions at the would-be call sites?

Opus 5 gave up when trying but its reasoning sounded made up and just confidently incorrect so I'm trying again with Fable

MarkMcCaskey:
> > What happens if you try to force usage of the new helpers by dismantling inline functions at the would-be call sites?
> 
> Opus 5 gave up when trying but its reasoning sounded made up and just confidently incorrect so I'm trying again with Fable

Nice, that worked!

MarkMcCaskey:
Looks like CI flaked, networking failed in some python setup or something

ribbanya (approved):

PR #2978, opened 2026-07-26
Reuse menu sfx inlines

MarkMcCaskey:
There are existing helper functions that wrap this function with its 3 arguments. If the base function were named and those 3 values were defined as an enum then I don't know how much gain there really is to using this, but this inline already exists and has semantic names so perhaps we want to use them. This also imports a `mn` helper to other areas of the code.

Happy to replace these helper functions instead with the named raw function + named args or just close this PR if this isn't the right direction to be going.


ribbanya (approved):
Hard to say. The fact that these exist in the TU suggests that they're either de-duplicated header functions, as the doc comments suggest, or were just written in-place and then inlined. I think it's fine for now, can look into other options later. Should be easy to bulk-replace either way.

PR #2979, opened 2026-07-26
Use named pad button constants instead of raw hex masks

MarkMcCaskey:
I noticed in the code near what's being touched in #2978 that there's a lot of magic numbers being used for input code that could be much more semantically meaningful. Claude claims to have found a misnamed existing constant, I'll verify that before taking this out of draft mode. Actually this seems well supported by the diff and the fact that LR+Start in the menu code also works in game.

From Claude:
> Rename PAD_LRAST to PAD_LR_START: bit 34 is derived from L+R+Start with no A held, so the old name described the wrong combo. Add PAD_LRA_START for the previously unnamed bit 35, which is the L+R+A+Start combo.
> 
> Spell the derived-bit combos in gm_EvaluateAllControllerInputs and all raw hex button-mask tests with the existing PAD_*, HSD_PAD_*, and MenuInput_* constants. Sites reading the u64/u32 controller-map masks use PAD_*; sites reading HSD_Pad*Status fields use HSD_PAD_* unless the file also needs the synthetic stick-direction bits that family lacks (gmcamera, gmresultplayer); menu popup handlers use MenuInput_*.


ribbanya:
> Rename PAD_LRAST to PAD_LR_START: bit 34 is derived from L+R+Start with no A held

Where?

MarkMcCaskey:
> > Rename PAD_LRAST to PAD_LR_START: bit 34 is derived from L+R+Start with no A held
> 
> Where?

In src/melee/gm/gm_1A36.c:119-123
```c
        gm_801A3820(i, PAD_TRIGGER_L | PAD_TRIGGER_R | PAD_BUTTON_START,
                    PAD_LR_START);
        gm_801A3820(
            i, PAD_TRIGGER_L | PAD_TRIGGER_R | PAD_BUTTON_A | PAD_BUTTON_START,
            PAD_LRA_START);
```

At least that part of the diff made it seem likely correct to me

PR #2983, opened 2026-07-26
Use HitElement and hit/hurt capsule state enum names

MarkMcCaskey:
These are harder to verify at a glance. I'll leave this in draft mode until I actually verify that these seem more than just close enough. Okay found a bit more to clean up and seems alright but I think this one needs more thorough review than the other rename passes I shipped recently.

From Claude:
> Replace 47 raw literals across 19 TUs: dmg.x1860_element and HitCapsule element comparisons spell HitElement_*, hitbox state stores and checks spell HitCapsule_*, hurt-capsule state checks and raw ftColl_8007B0C0 arguments spell HurtCapsule_Enabled/Intangible (the parameter is already declared HurtCapsuleState, and ftYs_Guard.c already mixes named and raw forms).
>
> Also annotate ftColl_803C0CAC with the same per-element index comments its value-identical twin it_803F1384 already carries.


src/melee/ft/ftcoll.c
```diff
@@ -714,9 +714,25 @@ bool ftColl_80076ED8(Fighter* fp0, HitCapsule* hit0, Fighter* fp1,
...
+    /* [HitElement_Sleep]    */ -1,
+    /* [HitElement_Catch]    */ -1,
+    /* [HitElement_Ground]   */ 1000,
+    /* [HitElement_Cape]     */ 1000,
+    /* [HitElement_Inert]    */ -1,
+    /* [HitElement_Disable]  */ -1,
+    /* [HitElement_Dark]     */ 1046,
+    /* [HitElement_Scball]   */ -1,
+    /* [HitElement_Lipstick] */ -1,
+    /* [HitElement_Leadead]  */ 0,
```
ribbanya: Start an enum in `ef/forward.h` for the effect IDs. Just label them `Ef_Id_Unk1000` etc. for now, so they can be identified and refactored later.

ribbanya (changes requested):

src/melee/ft/ftcoll.c
ribbanya: Also rename these arrays using local naming convention (`hit_effect_ids` for example).

ribbanya (approved):

PR #2997, opened 2026-07-31
Improve efAsync dispatch and data matching

MarkMcCaskey:
This is the second largest function in the decomp and it's not system code so giving it a go. It's pretty close but let's see if I can crack it or at least clean it up.


MarkMcCaskey:
Data can be cleaned up more but matching progress is hitting a ceiling.

AI summary of where we ended up at the last attempt:
> 0x41B (4 rows, Vec3 slot) — the biggest discovery. The target’s Vec3 at sp+0x14 is a compiler aggregate temp, not a user local: MWCC’s frame is ordered [outgoing args][temp pool][inline-billing][user locals][saves], and no user-declared object can enter the temp pool. I proved our compiler can produce such temps — our own efsync compile elides its case-0x4CC copy into a bottom temp at 0x70, and grFZeroCar_801CAFBC (a matched TU) shows the phenomenon too — but the identical source shape never elides in efasync, even as a 100%-faithful clone. Swept ~15 spellings (scopes, decl positions, arrays, helpers at every depth, jobj.h consumers, struct-return/comma temps — those land in a different, high temp class). Both efsync and efasync (both NonMatching upstream) share this exact unsolved residual class, so cracking it would fix two TUs.
>	•	0x439/0x43E (14 rows, scale-X cursor) — the single shared helper-local is load-bearing: splitting cursors breaks Y/Z too, expansion triggers an address-CSE the helper’s param indirection prevents, and param-order/alias/rank levers are all neutral or worse. This is the classic two-reg-swap class the community has cracked before — a decomp.me kit is the right next move.
>	•	0x438 (2 rows) and 0x419 (3 rows) — per-web allocator context, not variable identity; every reshaping regresses or breaks the delicate (void)-mention balance that fixed 40D/46A/46C.

ribbanya (approved):

ribbanya (changes requested):
Please add a concise doc comment describing the stack allocation behavior.

ribbanya (approved):

PR #3001, opened 2026-08-01
Improve matches in mncharsel

MarkMcCaskey:
Working on some other big functions, matching and cleaning them up.


MarkMcCaskey:
Matching stalled here for now, the `mnCharSel_803F0A48` reinterpret is odd. There's probably an explanation I don't understand but it seems like the same parts of static memory are referenced both from a base offset and directly at different places in the code.

MarkMcCaskey:
oops my bot that was looking at mwcc force pushed everything after allegedly figuring out the fix for the extra floating point register

PR #3003, opened 2026-08-01
Improve fn_8018B090 match (88.2% -> 99.0%)

MarkMcCaskey:
Working on another big function match / clean up

From Claude:
> Rewrite the bracket-slide GObj proc with the existing BracketEntry/BracketEntrySlot types instead of u8* pointer arithmetic. Key match fixes: negate inside the slide-Y helpers so MWCC fuses fnmadds/fnmsubs, switch on the entry mode byte instead of an if-chain, drop the entry pointer local in favor of direct lbl_80473AB8[idx] member paths, compute the slide amount as a single division, load JObjs after the mn_8022F470 calls, rewrite the unrolled slot loops as real 4-trip loops, and copy the case-24 bounce table with a struct assignment.
>
> Also retype lbl_803D9E1C as the s16 coordinate pair table it is (removing the out-of-bounds lbl_803D9D20.x0 casts) and match the original linkage of lbl_803D9DAC/lbl_803D9DD0.


MarkMcCaskey:
I'll come back to these big functions later and try to complete the match but probably not for a bit; need better tooling first.

PR #3004, opened 2026-08-01
Match it_8026C75C register allocation

MarkMcCaskey:
Testing out my vibe decompile of mwcc 1.2.5 and seeing if v0 is sufficient to get some 100% matches on things that got stuck before.


MarkMcCaskey:
It turns out there's a big difference between being able to understand things and being able to use that information to solve things. Not sure if there are any issues with sharing the decomp, but I'll make it public again later after I fill it out more.

There's definitely concrete info in there like how exactly things get scored/weighted. Claude when looking at the code just now called out the reg alloc as being https://en.wikipedia.org/wiki/Chaitin%27s_algorithm

Not sure how much is already known here, I didn't look too closely, but I think with a lot more work we can build code that can do more and maybe solve some cases.

PR #3010, opened 2026-08-02
Match and link vi1201v1 and vi1201v2

MarkMcCaskey:
Trying to finish up this TU with some new knowledge


MarkMcCaskey:
Let me know if you want me to propagate the names in the struct in v2 to those new globals or if I should keep iterating on code quality here before shipping.

PR #3016, opened 2026-08-06
Match and link groldpupupu

MarkMcCaskey:


ribbanya (approved):
@MarkMcCaskey You should teach it to run `ninja apply` whenever it links something, to prevent these CI failures (caused by `ninja diff`).

MarkMcCaskey:
> @MarkMcCaskey You should teach it to run `ninja apply` whenever it links something, to prevent these CI failures (caused by `ninja diff`).

Oh yeah that's way easier, thanks for the reminder!

PR #3021, opened 2026-08-08
Match and link vi0401

MarkMcCaskey:
From Claude:
> The register-permutation residual resolves with idiomatic structure: the second camera and model loop use their own locals (cam_gobj2, gobj2, jobj2), the loop index lives in a one-field struct, and fn_8031D168 is defined at its address-order position. The carrier also lets MWCC share the index zero with the gxlink_prios high-word store, which the plain u64 store spelling otherwise materializes twice.


MarkMcCaskey:
`idx.i` is a funny way to get this match. There's probably a real code structure that maps to this. When we figure out what that is can bulk update all these weird structs at once.

ribbanya:
Some kind of iterator struct?

MarkMcCaskey:
> Some kind of iterator struct?

My local setup is so janky, running ninja apply now but it's slow because I didn't install wine and I'm running it through Docker lol

MarkMcCaskey:
> Some kind of iterator struct?

I think it's just a way to move the register allocator in the right way, I think it's kind of arbitrary. Just a trick I suggested once that happened to work and the agent really held on to it.

ribbanya:
I meant in the original code, in response to this:
> There's probably a real code structure that maps to this. When we figure out what that is can bulk update all these weird structs at once.

MarkMcCaskey:
> I meant in the original code, in response to this:
> 
> > There's probably a real code structure that maps to this. When we figure out what that is can bulk update all these weird structs at once.

It could be but I think I've used the struct trick in a number of other places. Here's Claude's response to why it used it here:
> Yes — it's purely a register-allocation fix. The instruction stream is byte-identical with or without the struct; the only thing that changes is which physical registers MWCC assigns. Here's the mechanism, as established from the allocator captures this session:
>
> How MWCC picks registers: the allocator numbers every live range ("web") and colors them by popping a simplify stack. Pop order is driven by those web numbers, and the numbers come in fixed strata: parameters, then named locals (reverse declaration order), then second-region temps, then strength-reduction induction variables, then CSE/late compiler temps — with aggregate-promotion webs numbered above the SR IVs. First-popped webs claim registers descending from r31; later pops reuse the lowest-numbered already-claimed free register.
>
> The problem: the retail binary has i = r30 and the compiler-created byte-offset IV (slwi r29, r30, 2) = r29 — which requires i's web to pop before the IV's. But a plain named local always numbers in the locals stratum, which sits below the SR-IV stratum, so i always popped after the IVs and the four webs {i, both byte-IVs, the fn_8031D168 address temp} came out permuted (r29↔r30). We proved this ceiling is absolute for a multi-use named counter: all 9! declaration orders, block scoping, u32, (void) i, opaque assignment chains, embedded defs — every variant was measured dead in the simulator and/or real compiles.
>
> What the struct does: wrapping the counter as struct { int i; } idx; makes it an aggregate, so MWCC's aggregate-promotion/scalarization pass re-creates its web — and those webs number in a stratum above the SR IVs. Same lifetime, same interference graph, same instructions; just a higher web number, so it pops first in its pass, claims r30 fresh, and the whole four-web permutation cascades into the target coloring. Critically it stayed fully register-allocated — no memory home, no frame change (the prologue and stream stayed byte-identical; final diff had zero real rows).
>
> Bonus: with the carrier, the plain cam_gobj->gxlink_prios = 0x2A9; store recycles idx.i's zero into the u64 high word natively — which eliminated the extra li insert that started this whole effort, letting us drop the ugly u32-pointer-cast store spelling and the dead opacity variable entirely.
>
> Precedent: mnCharSel_8025FDEC shipped the same one-field-carrier shape for a coloring cycle. Caveat: it's not universal — in lb_8001044C a counter carrier got memory-homed and cost stack bytes, so it has to be measured per site. Here it happened to be exactly right, which is decent evidence the original source really did keep this index in a small struct.

So Claude maybe agrees that it was real here. But I'm not sure. This case does sound more compelling but I suspect this is just within easy reach in our bag of tricks and there's more tricks to be discovered.

MarkMcCaskey:
Okay apparently it takes 45 minutes for me to run ninja apply, let me see if I can just install wine. I saw it's being deprecated from homebrew.

Issue #2933, opened 2026-07-20
Tracking: Scratches for the <10B remaining TUs


jurrejelle:

g-guthrie:
Working on `gm_801BFCFC` in `melee/gm/gm_1BFA` now; I’ll follow up with the verified match.

ribbanya:
Closing this because the only remaining TU with a scratch is `lbshadow`, and that scratch is for the wrong function. Refer to [the wiki](https://github.com/doldecomp/melee/wiki/Translation-Units#text-in-progress) for an updated list of nearly-complete TUs.

PR #3030, opened 2026-08-09
Link `ftdynamics` and `ground`

MarkMcCaskey:
Looking into more close matches from #2933.

This PR is a follow up on #3025


MarkMcCaskey:
Yeah good to merge, lbshadow is a nightmare. It's at a point where 2 distinct concepts share a register but there's no idiomatic way to express that that doesn't regress the function in major ways. Any sensible notion of a union doesn't work there but that's what the assembly effectively is doing anyway.

What source generates the assembly of lbshadow is beyond what my tools can understand. The agent tried tons of obscure pragmas even haha.

MarkMcCaskey:
Just in case anyone else wants to look into this before I do again, here's the handoff doc Codex just generated for lbshadow's function:

# Handoff: `lbShadow_8000F38C`

Last verified: 2026-08-09 (America/Los_Angeles)

This is the focused continuation document for the last unmatched function in
`src/melee/lb/lbshadow.c`. It supersedes the lbshadow portion of
`MATCHING_HANDOFF_GROUND_LBSHADOW.md`; that older document remains useful for
the history of the preceding ground matches.

## Executive summary

The committed function is structurally exact and scores **99.8540%**. Target
and candidate have the same 411-instruction graph, the same `0x148` stack
frame, and the same register assignment everywhere except one instruction:

```asm
              target                         candidate
li            r24, 0                         r24, 0
copy zero     addi r28, r24, 0               li r28, 0
```

The target initializes the `fallback` light pointer by copying the zero in the
`lobj` register. The candidate rematerializes zero. This choice happens before
register allocation: direct caller-scope source loses the dependency during
value numbering, while an inline-origin version retains the dependency but
changes web ownership and stack billing. The allocator cannot turn the second
`li` into an `addi` after the dependency has been erased.

Do not describe this as a generic register-allocation mismatch. The physical
colors are already right in the committed source (`lobj` is `r24`, `fallback`
is `r28`). The remaining problem is preserving one IR dependency without
changing the origin/ranking of either web.

## Repository state

- Repository: `/Users/mark/etc/melee`
- Branch: `match-followups-0808`
- Verified branch tip: `0f9f159e0` (`Link ground.c`)
- lbshadow checkpoint: `448544c70` (`Improve lbShadow_8000F38C`)
- Remote PR: <https://github.com/doldecomp/melee/pull/3030>
- Compiler: MWCC GC/1.2.5n, `-O4,p`
- Tooling repository: `/Users/mark/etc/mwcc`


## Exact source and residual

The important committed source is the nested block near the beginning of
`lbShadow_8000F38C`:

```c
{
    HSD_LObj* lobj = (HSD_LObj*) (arg0 - arg0);
    HSD_LObj* fallback = lobj;
    {
        HSD_GObj* lgobj;
        for (lgobj = HSD_GObjGXLinkHead[4]; lgobj != NULL;) {
            nextGx = lgobj->next_gx;
            lobj = (HSD_LObj*) lgobj->hsd_obj;
            while (lobj != NULL) {
                if (lobj->flags & 3) {
                    fallback = lobj;
                }
                if (HSD_LObjGetFlags(lobj) & 0x400) {
                    nextGx = NULL;
                    break;
                }
                if (lobj == NULL) {
                    lobj = NULL;
                } else {
                    lobj = lobj->next;
                }
            }
            lgobj = nextGx;
        }
    }
    // ...
}
```

Verify it with:

```sh
build/testfn_any.sh \
  src/melee/lb/lbshadow.c \
  build/GALE01/obj/melee/lb/lbshadow.o \
  lbShadow_8000F38C
```

Expected output:

```text
======== lbShadow_8000F38C ========
=== lbShadow_8000F38C: 99.8540%
  [  25] REPLACE REAL T: addi r28, r24, 0x0  C: li r28, 0x0
```

The target window begins at object offset `0x9f8`:

```asm
lwz   r3, HSD_GObjGXLinkHead@sda21(0)
li    r24, 0
addi  r28, r24, 0
lwz   r3, 0x10(r3)
```

The candidate is identical except for the third instruction. Preserve the
unusual `arg0 - arg0` initializer and the nested scopes unless an experiment
is deliberately testing their effect. Preserve the duplicated
`if (fp2->x21FC_flag.b7)` later in the function; cleaning it changes matched
control flow.

The immediately preceding checkpoint `a689b376a` had a second FPR/literal
residual. Commit `448544c70` solved it with:

```c
lightPos.z = upVec.y = lbShadow_804D7B88;
```

That chained assignment gives the named `1.0F` constant the correct FPR web
rank and store order. Do not split it back into two assignments.

## What allocator capture proved

The direct committed source was captured at:

```text
/private/tmp/mwcc-lbshadow-f38c-0808
```

The full light-scan inline-helper variant was captured at:

```text
/private/tmp/mwcc-lbshadow-helper-0809.EApFlm
```

The paths are scratch storage and may eventually disappear, so the essential
results are recorded here:

- In the direct/caller-scope form, the important GPR webs are approximately
  `fallback` v40 -> `r28` and `lobj` v41 -> `r24`.
- Both have degree 12 and the target physical colors. The candidate's second
  zero is already an independent `li` in optimized pcode before allocation.
- In the full scan-helper form, `lobj` and `fallback` are approximately v56
  and v57. They again have degree 12 and nearly identical neighbor sets, with
  the expected mutual interference because both values are live during the
  search.
- Inline provenance preserves the copy operation, producing the wanted
  `li; addi` pair, but the scan's `lobj` web colors `r29`; `fallback` remains
  `r28`. The resulting function has 15 register-operand differences even
  though its control-flow and instruction graph are exact.
- Later in the target, `r29` belongs to the fighter pointer. `r24` is reused by
  several non-overlapping values after the light scan: `hasObj`, shadow
  intensity, and the viewing-rectangle loop counter `i`. The target's `lobj`
  web participates in this sequential register ownership. A helper-local web
  is allocated in a different origin/rank stratum and does not naturally join
  that sequence.

This establishes two independent constraints:

1. Inline-origin lowering is the known way to retain the zero-copy dependency.
2. Caller-origin lifetime ordering is the known way to give `lobj` target
   ownership of `r24`.

The missing source shape must satisfy both at once.

## Union experiments

Two distinct union ideas were tested.

### `lobj` and `fallback` in one union

This cannot model the target lifetimes. The scan needs the current `lobj` and
the previous qualifying `fallback` simultaneously, and the allocator graph
correctly contains an interference edge between them. A union asserts shared
storage where the program needs two live values. MWCC therefore keeps the
aggregate in memory/stack-homes it; it does not create two coalescible scalar
webs.

Tested forms included a same-type pointer union and a `u64` union with two
32-bit halves. They grew or shifted the stack arena and did not reproduce the
target registers.

### `lobj` and later `hasObj` in one union

This is semantically more plausible because those lifetimes do not overlap and
the target reuses `r24` for both roles. It still fails in MWCC: accessing
different union members makes the storage type-punned/addressable, prevents
scalar replacement, adds an 8-byte home/billing band, and leaves the helper's
internal formal/web in the wrong allocation stratum.

The conclusion is not that lifetime sharing is wrong. It is that a C aggregate
is the wrong way to request it from this compiler. The successful shape, if
found, will use one scalar web or compatible inline provenance without taking
an address or introducing aggregate storage.

## Experiments completed

The following directions have been tested. Avoid repeating them without a new
mechanical hypothesis.

### Direct source spellings

All of these keep the exact frame and target colors but are canonicalized to
two independent `li` instructions:

- declaration initializers and later assignment statements;
- `fallback = lobj`, both chained-assignment directions, and comma forms;
- pointer casts and integer-zero/algebraic-zero spellings;
- block-scope and loop-header initializers;
- constant-returning or identity inline helpers on the RHS;
- zero-offset member/address expressions;
- `HSD_LObjGetNext(NULL)` (this also added an 8-byte home);
- the `register` keyword, which is inert here at `-O4,p`.

### Volatile, aggregate, and aliasing shapes

- volatile pointer objects and volatile pointees introduce loads/stores or
  homes;
- struct/union aliasing and type-punning stack-home the shared object;
- `memcpy` and address-taking forms likewise destroy the exact frame;
- `u64` direct/split carriers either fold to two `li` instructions or add an
  8-byte billing band;
- a dynamic conditional `u64` zero path can emit `li; addi`, but it retains the
  branch scaffolding and is not instruction-equivalent.

### Inline shapes

- A full light-scan inline helper preserves the exact copy instruction and
  instruction graph, but colors `lobj` `r29` instead of `r24` (15 operand
  differences).
- A pointer-only helper retains caller colors but post-inline optimization
  folds the identity copy.
- A caller-owned local passed by address/value to a helper adds stack billing
  or still puts the internal web in `r29`.
- Splitting the first fighter-reset loop into the caller and putting the
  remainder in a helper preserves the copy, but the `noLight` formal adds a
  copy/web and rotates the saved registers.
- A whole-function inline wrapper tested on 2026-08-09 preserved `li; addi`,
  proving again that inline provenance is the trigger. It shrank the frame from
  `0x148` to `0x138`, merged the two early zero webs, and changed saved-register
  ownership throughout. Its pair was `li r27,0; addi r26,r27,0`, not the target
  `r24/r28` pair.
- Making `noLight` a caller-owned formal and increasing `PAD_STACK` inside the
  inline did not help. Unused inline-local padding is eliminated; caller-owned
  stack billing is what matters.

### Optimizer pragmas

Disabling `opt_propagation`, `opt_common_subs`, or `opt_lifetimes` for the whole
function perturbs many otherwise exact instructions and still does not produce
the required combination.

On 2026-08-09, tiny identity helpers were compiled under individually scoped
`opt_propagation off`, `opt_dead_assignments off`, and `opt_common_subs off`.
In all three cases, the caller's later optimization still folded the copy to
`li`, while merely introducing the inline helper increased the frame to
`0x150`. This rules out “freeze the copy in a one-line helper” using the normal
optimizer switches.

`#pragma disable_registers` is a broad function-level register-optimization
switch used for setjmp-like cases. It cannot reserve/exclude only `r29` and is
not a targeted allocator mask.

### Inline assembly

- Named inline assembly using `addi`, `subi`, or `la` is normalized and then
  constant-folded.
- Raw `opword` can force bytes but changes the compiler's stack arena/register
  graph and is not an acceptable source match.

## Exact donor patterns

A read-only scan (`build/twinscan_zerocopy.py`) found 121 target-object sites
with adjacent `li rA,0; addi rB,rA,0`. The useful matched donors show that MWCC
emits the pair for real add/loop/pair lowering, not for an ordinary pointer
identity after GVN.

- `HSD_GObj_80391304`: chained array stores in a loop,
  `HSD_GObjGXLinkHead[i] = HSD_GObj_804D7820[i] = 0`, produce a zero loop value
  and an `addi` copy used as the byte offset.
- `PickAutoName`: `name_idx = 0; char_idx = name_idx;` produces
  `li r9,0; addi r5,r9,0`. Both are integer scalars that later diverge/mutate.
- `mnCount_8025035C`: initial values for two loop-owned integer webs produce
  `li r25,0; addi r31,r25,0`.
- `lbArchiveRelocate`: `Locate`, an inline relocation loop, produces
  `li r5,0; addi r6,r5,0` for its count and byte-offset webs.
- `HSD_GObj_80390CFC`: a conditional `u64` value emits the low/high zero halves
  as `li; addi`.
- `AXFXDelaySettings` and `OSInitAlloc`: loop counters plus derived offsets or
  chained zero stores produce the same lowering.

The especially interesting contrast is `PickAutoName`: its plain integer copy
survives while the analogous pointer copy in lbshadow does not. That suggests
the next pass should test the compiler's pointer-vs-integer value-numbering
boundary, not more superficial pointer assignment spellings.

## Ranked next experiments

1. **Use 32-bit integer scalar carriers for the scan.** Model `lobj` and
   `fallback` as `uintptr_t`/`intptr_t` (or repository-equivalent `u32`) through
   the search and cast only at dereference/call boundaries. PowerPC pointers are
   32-bit, so the casts should be code-free. The hypothesis is that two mutable
   integer webs will lower like `PickAutoName` and retain the initial `addi`
   while remaining caller-origin and eligible for the target `r24/r28` colors.
   Measure the first 40 instructions before polishing every use. If register
   colors or schedule change broadly, stop.
2. **Find a smaller integer-derived offset formulation.** Keep `lobj` typed as
   a pointer but make `fallback` begin as a genuinely used integer-derived
   zero/offset web that later becomes the pointer carrier without added
   instructions. `lbArchiveRelocate` and `HSD_GObj_80391304` are the best donor
   sources. Avoid undefined pointer arithmetic.
3. **Extend inline scope only across sequential `r24` owners.** If integer
   carriers fail, construct one helper that contains the scan plus later
   `hasObj`/intensity/`i` roles so those non-overlapping webs share the same
   inline allocation stratum. Do not pass `noLight` as a new formal; the prior
   split proved that formal rotates the early saved registers. Capture the
   allocator after the first structural result rather than sweeping syntax.
4. **Improve the MWCC tooling.** The existing semantic register-web alignment
   and inline stack-slot ownership work in `/Users/mark/etc/mwcc` is directly
   applicable. A cross-capture report should identify which caller web must be
   given the surviving inline copy node and whether the required web numbering
   can arise from a legal source transformation.

## Fast stop conditions

Stop an experiment early if any of these occur:

- frame differs from `0x148` after accounting for an intentional temporary pad;
- instruction count or control-flow graph changes;
- the `li; addi` pair appears but `lobj` is `r29` and no later `r24` consumer is
  in the same source/allocation region;
- the second instruction is already `li` in optimized pcode—allocator tuning
  cannot recover the lost dependency;
- a union/struct/volatile object becomes addressable or receives a stack home.

## Before committing an exact solution

1. Format only `src/melee/lb/lbshadow.c`.
2. Run `python3 tools/check/main.py --fix --quiet src/melee/lb/lbshadow.c`.
3. Run `git diff --check`.
4. Re-run `build/testfn_any.sh` for every function in the TU, not only F38C.
5. Compare the complete lbshadow object, including `.data`, `.sdata`, and
   relocations. The currently known target/candidate tails differ in size
   (`.data` approximately `0x78/0x74`, `.sdata` approximately `0x18/0x12`), so
   an exact function alone is not yet proof that the TU links.
6. Commit the exact function separately with the allocator/value-numbering and
   stack-layout explanation. Then set `melee/lb/lbshadow.c` to `Matching` in
   `configure.py`, verify the full link/diff, commit the link gate separately,
   and push `match-followups-0808` only with explicit authorization for this PR.

## Commit-message material for the eventual match

The final commit should record:

- the original single residual and score (`99.8540%`);
- why the target `addi` represents a surviving semantic dependency rather than
  a late register allocator preference;
- how the winning source preserved caller-owned sequential `r24` lifetime
  reuse while retaining inline/add lowering for `fallback`;
- the exact frame before/after (`0x148`), absence of stack homes, and unchanged
  instruction/control-flow graph;
- the key negative result that unions impose storage identity while the target
  requires two simultaneous scalar webs.

This was a hard-won one-instruction case. Keep the explanation in the commit;
do not reduce it to “fix register allocation.”

PR #3033, opened 2026-08-10
Match grCorneria_801E25C4 and link grcorneria

MarkMcCaskey:
One more for #2933


MarkMcCaskey:
Nice, thanks! Much cleaner. `ninja apply` is super slow on my machine was still running and forgot it in the last PR

ribbanya:
Hm, that's odd, it should all be native tooling. What is your environment?

MarkMcCaskey:
> Hm, that's odd, it should all be native tooling. What is your environment?

I'm on a first gen ARM64 Mac laptop, the M1. So maybe it's a combination of Rosetta + wine overhead. Wine is much faster than docker though.

PR #3038, opened 2026-08-11
Link `lbrefract`

Danny-Dasilva:
> Matches `lbRefract_800222A4` by recovering the contiguous refraction data layout and simplifying its initialization loop.
>
> `lbRefract_800222A4`: 100.000%


src/melee/lb/lbrefract.c
```diff
@@ -379,11 +379,33 @@ HSD_TObjDesc tobjdesc1 = {
     NULL,
 };
 
+static char lbRefract_filename[12] = "LbRf.dat";
+static char lbRefract_symbol[12] = "lbRefData";
+
 /// @todo Only differs by register allocation.
 void lbRefract_800222A4(void)
 {
+    struct lbRefract_DataLayout {
```
r-burns, line 390: This looks unsafe, but I wonder if your other modifications are what actually fixed the match and we can drop this

ribbanya: It seems to require a struct layout. Added a comment and reverted what I could of it without breaking the match.

PR #3045, opened 2026-08-12
Rename and document some undocumented variables

EIREXE:
Through my own effort to understand melee movement, I have had to manually figure out the meaning of certain variables.

Additionally, some variable names in ftCo_DatAttrs were wrong/unclear:

- `walk_init_vel` -> `walk_accel_stick_based`: Actually not a velocity at all, it's the stick based fraction of the movement acceleration
- `walk_accel` -> `walk_accel_base`: Changed to clarify it's the stick-position-independent part of the walk acceleration
- `dash_run_acceleration_{a, b}` -> `dash_run_acceleration_{stick_based, base}`: Renamed for clarity


ribbanya (changes requested):
I'll take a closer look shortly. Couple things I notice for now:
* `_stick_based` should be changed; it's very clunkily worded.
* Rather than inline comments for the default values, it should be incorporated into doc comments, ideally with notation of which specific game version the value is from.

ribbanya (approved):
There are still some naming inconsistencies, especially with abbreviations, but those can be addressed later.

PR #3050, opened 2026-08-13
Partially match `hsd_803AAA48` and `fn_803B26CC`

TheTom:
Two more in the memory card code, including the big 5920-byte state machine.

> The command sequences shared five near-identical rewind/close blocks that MWCC needed to see as common tails; deduplicating them into file-local inline helpers and typing the s32 handle parameters as CardState* reproduces retail's register assignment throughout. The cmd[] pointer stores follow the file's existing idiom.

### DIFF
--- a/src/sysdolphin/baselib/hsd_3A94.c
+++ b/src/sysdolphin/baselib/hsd_3A94.c
@@ -592,12 +592,12 @@ s32 fn_803AA790(void)
             break;
         case 1:
         case 2:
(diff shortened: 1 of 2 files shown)


ribbanya (commented):
Some of these changes are arguably regressions, but I'll merge anyway for now.

TheTom:
Fair call. The rewrite dropped some of the descriptive local and helper names the file already had (file_count, file_size, the blocks_before/chunk_size helpers). Will send a follow-up restoring the naming where it does not disturb the match.

PR #3055, opened 2026-08-13
Match grCastle_801CF868 and link grcastle

MarkMcCaskey:
Heads up this includes a CI fix to force checkout so the compare can be done even if the CI tree is dirty.


ribbanya:
> Heads up this includes a CI fix to force checkout so the compare can be done even if the CI tree is dirty.

Why would the CI tree be dirty?

PR #3075, opened 2026-08-14
Box of matches + link ftcoll + link texpdag

strabitz:
AI assisted. Notes:

> Single-member anonymous structs (e.g. `struct { T v; } x;`) move a carrier out of the scalar mint band into the aggregate band. Used in mpIsland_8005A728, ftColl_80076ED8, ftColl_8007A06C.
> Zero-byte use anchors (e.g. `(void) x;`, `(void) arr[i];`) emit no instruction but count as a use, which keeps a carrier alive past copy-propagation or places a definition. Used in mpIsland_8005B004, it_8026CF04, ftCo_800AEA8C.
> Rebinding a dead parameter or local as a scratch carrier. Used in lbMemory_80014FC8, ftCo_80095D5C, ftColl_8007BE3C.
> `static inline` boundaries relocate only the webs you move. Used in it_8026D018, mpIsland_8005A728, ft_800852B0.
> `PAD_STACK` retuning wasn't used as a fix on its own but pairs with a structural change in four of these.
> Declaration and include order: grMuteCity_801F1328 is matched by moving one `#include` and nothing else.


strabitz:
Dropped fn_8001EBF0 since it was just merged into main a few minutes ago

PR #3077, opened 2026-08-14
Improve hsd_803B51C8

g-guthrie:
## What changed

- replace four fixed-size JPEG metadata `memcpy` calls with typed aggregate copies so MWCC emits the target's inline word-copy sequences
- hoist the quantization divisor once per coefficient loop instead of reloading the global for every coefficient
- limit the chroma quantization table pointer lifetime so it reuses the target register
- align the stack frame and width/height register allocation with the original function

## Why

`hsd_803B51C8` was one of the largest remaining fuzzy mismatches. The fixed-size copies and repeated global loads produced substantially different instruction sequences from the original JPEG encoder.

The local target-object comparison improves from 81.44211% to 95.6%.

## Validation

- `uvx pre-commit run --files src/sysdolphin/baselib/hsd_3B34.c`
- exact MWCC 1.2.5n compile with project flags
- local objdiff comparison against the exported `hsd_803B51C8` target object: 95.6%

### DIFF
--- a/src/sysdolphin/baselib/hsd_3B34.c
+++ b/src/sysdolphin/baselib/hsd_3B34.c
@@ -33,6 +33,18 @@ typedef union JpegMetadata {
     } data;
 } JpegMetadata;
 
+typedef struct JpegComment {
+    u8 data[0x15];
+} JpegComment;
+
+typedef struct JpegHuffmanDC {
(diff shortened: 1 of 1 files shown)


ribbanya:
`hsd_803B51C8` is matched in #3223.

PR #3078, opened 2026-08-14
Link ft_0852

g-guthrie:
## What changed

- Mark `melee/ft/ft_0852.c` as matching so it is linked from source.
- Keep the otherwise indirect-only `ft_8045993C` BSS symbol emitted so the following globals retain their original addresses.

## Why

The current decomp.dev report at `7eb151126e626ffc342e285896ac2f1fd78d9092` reports all three functions, the full 836-byte text section, and the full 448-byte data section at 100%. The last remaining function was matched in #3075, but the unit was left `NonMatching` in `configure.py`.

## Validation

- Confirmed no open PR or issue references `ft_0852` or `ft_800852B0`.
- Confirmed the live decomp.dev unit report is 100% for code and data.
- `git diff --check` passes.
- The first CI link attempt identified the missing `ft_8045993C` BSS emission and its resulting 0x30-byte address shift; the follow-up uses the repository's existing no-codegen address-reference pattern to retain it.
- A local full link could not be run because this checkout intentionally has no `orig/GALE01/sys/main.dol`; project CI is the authoritative clean-ROM link check.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -576,7 +576,7 @@ config.libs = [
             Object(Matching, "melee/ft/ft_0819.c"),
             Object(Matching, "melee/ft/ft_081B.c"),
             Object(Matching, "melee/ft/ft_084E.c"),
-            Object(NonMatching, "melee/ft/ft_0852.c"),
+            Object(Matching, "melee/ft/ft_0852.c"),
             Object(Matching, "melee/ft/ftdata.c"),
             Object(Matching, "melee/ft/ftmotionstates.c"),
             Object(Matching, "melee/ft/chara/ftCommon/ftCo_Init.c"),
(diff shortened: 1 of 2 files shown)


ribbanya:
Assigning `static` to BSS entries has an impact on their order. Also, most of the bss entries here are at an offset of +4 (rather than +8) meaning they cannot be split apart from each other. If you move a data member into the same TU as a function that uses it, and that function expects it to be `extern`, it will start referencing that member relative to the data section, breaking the match.

I suspect that most or all of the data members (including bss) live alone in a TU without a text section, similar to `gmscdata`, but I'm not sure.

ribbanya:
Well okay, that works too...

PR #3081, opened 2026-08-14
Improve fn_801701C0 match to 98.94%

g-guthrie:
## Summary

- load the three `x3` flag bytes by byte offset before rematerializing the player record
- restore the target instruction order and exact 6,344-byte function size
- improve `fn_801701C0` from 98.51009% to 98.93884% in the relocation-stripped MWCC comparison

## Validation

- exact MWCC 1.2.5n comparison: 98.93884%, size 6,344 bytes
- clang-format pre-commit hook
- changed-file style check
- full source style sweep
- `git diff --check`

The local retail-DOL matrix was unavailable because this checkout does not contain the required user-owned game binary; the pull request CI will run that matrix.

Generated with Codex.

### DIFF
--- a/src/melee/gm/gm_16F1.c
+++ b/src/melee/gm/gm_16F1.c
@@ -1075,8 +1075,8 @@ int fn_801701C0(void* arg0, int arg1, int arg2)
                     p++;
                 }
             }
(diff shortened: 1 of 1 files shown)


ribbanya:
`fn_801701C0` has a higher score (99.99%) in #3223.

PR #3083, opened 2026-08-14
Improve hsd_803B51C8 matching

g-guthrie:
## What changed

- model the JPEG comment and Huffman tables as fixed-size aggregate types so MWCC emits the original inline copy sequences
- cache the quantization scale once per transformed block instead of reloading it for every coefficient
- align local layout, integer widths, and dimension declaration order with the target function

## Why

`hsd_803B51C8` used generic `memcpy` calls and repeatedly loaded the quantization scale inside three unrolled loops. Those source shapes differed substantially from the original compiler output.

The aggregate assignments reproduce the target's paired loads/stores and counted AC-table loops, while the cached scale reproduces the target's single load before each unrolled division block.

## Matching improvement

- stripped function match: **81.44211% -> 94.7579%**
- candidate size: **2592 -> 2664 bytes**
- target size: **2660 bytes**

## Validation

- `pre-commit run --files src/sysdolphin/baselib/hsd_3B34.c`
- `python3 tools/check/main.py --quiet src/sysdolphin/baselib/hsd_3B34.c`
- `git diff --check`
- exact MWCC 1.2.5n compile and address-verified objdiff comparison against `hsd_803B51C8`

### DIFF
--- a/src/sysdolphin/baselib/hsd_3B34.c
+++ b/src/sysdolphin/baselib/hsd_3B34.c
@@ -1392,8 +1392,8 @@ s32 hsd_803B51C8(s32 arg0, s32 arg1, s32 arg2, char* arg3, s32 arg4)
     s32 scratch_r6_5;
     s32 scratch_r7_2;
     s32 scratch_r7_4;
-    s32 work_r24;
     s32 work_r25;
+    s32 work_r24;
     s32 work_r26;
     s32 work_r3;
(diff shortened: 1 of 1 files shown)


g-guthrie:
Superseded by #3077, which improves the same function further (+425 bytes vs +402 bytes in the decomp.dev report). Keeping the stronger single contribution open.

PR #3089, opened 2026-08-14
Fix Item_8026AD20 argument

g-guthrie:
## What changed

- Add the retail fifth integer argument to `Item_8026AD20`'s declaration and definition.
- Pass `1` from the forward-throw path and `0` from the backward-throw path in `ftCo_80095EFC`.

## Why

The retail `ftCo_80095EFC` call sites materialize `r6` immediately before both `Item_8026AD20` calls. The current four-argument prototype omitted those instructions even though the callee does not consume the value.

## Impact

This restores both missing call-site instructions and improves the remaining unmatched function without changing `Item_8026AD20`'s body behavior.

## Validation

- Exact MWCC GC/1.2.5n compile of `ftCo_ItemThrow.c` and `item.c`
- Function scorer: removed two missing instructions from `ftCo_80095EFC` (penalty 1410 to 1210)
- `python3 tools/check/main.py --fix --quiet`
- `clang-format 22.1.0 --dry-run --Werror`
- `git diff --check`

### DIFF
--- a/src/melee/ft/chara/ftCommon/ftCo_ItemThrow.c
+++ b/src/melee/ft/chara/ftCommon/ftCo_ItemThrow.c
@@ -554,16 +554,19 @@ void ftCo_80095EFC(Fighter_GObj* gobj)
                                                         ftCo_MS_LightThrowF]
                                      .x8;
                     float throw_speed = throw_scale * base_throw_speed;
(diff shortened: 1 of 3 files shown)


jurrejelle:
Item_8026AD20 and ftco_itemthrow are linked now

PR #3097, opened 2026-08-15
Add minor clean ups to itlinkhookshot

MarkMcCaskey:
Failed attempt at matching, found some clean ups. removing `test_comp` allows removing the extra noinline wrapper


MarkMcCaskey:
Not sure if this is worth merging given how small of a change this is but sharing in case it is

PR #3101, opened 2026-08-15
match and link itlinkhookshot.c

lukechampine:
Thank you Sol Ultra :)

Remaining item functions: 0 🎉 
Remaining item TUs: 0 🎉


src/melee/it/items/itlinkhookshot.c
```diff
@@ -1205,6 +1207,62 @@ void it_802A49B0(ItemLink* link_0, Vec3* arg1, itLinkHookshotAttributes* arg2,
...
+{
+    volatile f32 y;
+
+    if (x > 0.0F) {
+        f64 guess = __frsqrte((f64) x);
+        guess = 0.5 * guess * (3.0 - guess * guess * x);
+        guess = 0.5 * guess * (3.0 - guess * guess * x);
+        guess = 0.5 * guess * (3.0 - guess * guess * x);
+        *(&y + 6) = (f32) (x * guess);
+        return *(&y + 6);
```
r-burns, line 1240: this is very unsafe lol, but I suppose we can drop the offsets for nonmatching builds

lukechampine: oof wow, let me see if Sol can find something less insane

PR #3104, opened 2026-08-15
Resolve memory card names & structures and add missing gxends()

EIREXE:
GXEnd is a noop in melee, so there were some missing ones that the decomp mixed.

Additionally I also resolved the names and structure for some of the memory card & save related variables, in theory it should be matching.


ribbanya (approved):
Failing CI will be addressed in a subsequent PR.

PR #3110, opened 2026-08-16
Replace all `inline` with `static inline`

ribbanya:
Kind of a lazy stopgap at the moment. Need to eliminate `inline` keyword where possible, and move `static` keyword to the declaration (not the definition).

https://github.com/r-burns/doldecomp-melee/pull/2#issuecomment-5305776649

### DIFF
--- a/src/MSL/ansi_fp.c
+++ b/src/MSL/ansi_fp.c
@@ -6,7 +6,7 @@
 #define __LO(x) (((s32*) &x)[1])
 
 /// @todo somehow make this work with math.h
-inline int _fpclassify(double x)
+static inline int _fpclassify(double x)
 {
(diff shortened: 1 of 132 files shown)


r-burns:
I think you're going to have some trouble with this, there are a lot of cases where they definitely just used regular inline. I think that's probably the more common usage, tbh, I haven't seen any confirmed cases of `static inline`.

r-burns (approved):

ribbanya:
`HSD_JObjSetupMatrix` is the only instance where `static` wouldn't match. I tried `WEAK` instead but that didn't work. It's emitted to `lbcollision` so I'm not sure exactly what they did, maybe it really is just plain `inline`.

PR #3116, opened 2026-08-16
Clean up ifmagnify

MarkMcCaskey:
follow up to #3114, clean up the TU

this clean up removes the local variable name `teams_enabled = gm_8016B168();`, that’s pretty semantically useful if that’s a confirmed name. I won’t extend this PR into naming symbols just to keep the scope focused but probably worth another pass for that.

Also the edge logic is still magic numbers instead of an enum. More work needed here for sure.


src/melee/if/ifmagnify.c
```diff
@@ -263,8 +264,7 @@ static inline void ifMagnify_GetCornerColors(GXColor* colors, Vec3* world_pos)
         } else {
             x_class = 2.0f;
         }
-        color_ids =
-            (u8*) &ifMagnify_803F984C[(s32) x_class + ((s32) y_class * 4)];
+        color_ids = ifMagnify_803F984C[(s32) x_class + ((s32) y_class * 4)];
```
r-burns:
```suggestion
        color_ids = &ifMagnify_803F984C[y_class][x_class];
```

does this work?

MarkMcCaskey: It didn't when I first tried, but given this commit ended up in a separate PR. I can do a broader pass over the whole TU. Let me take another pass at this before this PR merges

MarkMcCaskey: Mentioned in a commit message below about why this doesn’t work:
> A true 2D [4][4][4] table cannot match: MWCC strength-reduces each subscript separately (y<<4 + x<<2), while the DOL has the linear form ((y*4 + x) << 2).

That explanation sounds plausible to me but maybe there’s a way to make this work

PR #3117, opened 2026-08-16
Finish matching and link remaining Kirby TUs

MarkMcCaskey:
Building off of @ribbanya's work, finish up the Kirby TUs. I ended up removing the Kirby copy TU. It seems the alignment requires that split to be undone due to an alignment to 4 bytes required between them which, as I understand it, is not possible with 2 TUs that require 8 byte alignment.


ribbanya:
Might want to reference https://github.com/doldecomp/melee/compare/master...ribbanya:melee:wip/ftkirby

MarkMcCaskey:
> Might want to reference [master...ribbanya:melee:wip/ftkirby](https://github.com/doldecomp/melee/compare/master...ribbanya:melee:wip/ftkirby)

I’m glad I opened this PR early! If you have a lot of work here, happy to move on to another area. Looks like you’re much further along

ribbanya:
I'm not, it's just data splits. Claude should be able to take what it can from my branch.

MarkMcCaskey:
Matching work done, I don't have time to do a human review of all the changes in this diff right now so I'll leave it in draft mode until I do. I'll review it tonight if I'm able and then clean up or mark it ready to review.

ribbanya (approved):
Huh, that is actually pretty shocking that it linked the whole thing.

Needs a little data cleanup but it's fine to merge for now.

MarkMcCaskey:
> Huh, that is actually pretty shocking that it linked the whole thing.
> 
> Needs a little data cleanup but it's fine to merge for now.

Yeah it wasn't able to link it split as two apparently because of that offset. I think I'm missing some context but it seemed it had to be remerged to be able to get both to link.

ribbanya:
It just recognized that it was split on a 4-byte boundary instead of 8-byte (which the linker requires).

MarkMcCaskey:
> It just recognized that it was split on a 4-byte boundary instead of 8-byte (which the linker requires).

Oh yeah, I just meant I was missing context about why we tried to split it out originally. Though it looks like it was successfully split into many pieces.

PR #3129, opened 2026-08-17
Match `Toy_803067BC` (toy)

Danny-Dasilva:
`Toy_803067BC` 96.30 → 100.00.

The table address is held in an `s32` and converted once, rather than casting the
pointer through a scalar inside each address expression:

```c
    base = (s32) _Toy_sbss_804D6E64;
    ...
    src = (s16*) ((arg0 << 1) + base);
```

The conversion is needed because the target adds the offset first
(`add r4, r3, r29` vs `add r4, r29, r3` without the conversion), and MWCC only emits that order for an addition of two
integers. Any pointer form puts the pointer operand first, which costs one
instruction per branch.


src/melee/ty/toy.c
```diff
@@ -1353,14 +1352,11 @@ void Toy_803067BC(s32 arg0, s32 arg1)
...
-    src = base + arg0;
-    j = count;
-    dest = Toy_sbss_804D6EDC + count;
-    if (count != 0) {
-        for (; j != 0; j--) {
-            *dest-- = *src;
-            src += 3;
-        }
+    src = (s16*) ((arg0 << 1) + (s32) base);
+    dest = (s16*) ((s8*) Toy_sbss_804D6EDC + (offset = count << 1));
```
r-burns: Let's try to clean up this pointer math first

ribbanya (changes requested):
https://github.com/doldecomp/melee/pull/3129#discussion_r3794120164

Danny-Dasilva:
Cleaned up — both lines now use the same idiom, and the `(s8*)` byte reinterpretation is gone:

```c
    src = (s16*) ((arg0 << 1) + (s32) base);
    dest = (s16*) ((offset = count << 1) + (s32) Toy_sbss_804D6EDC);
```

Two parts of it resist further cleanup, and I wanted to show the evidence rather than just assert it.

**`src` can't be `&base[arg0]`.** MWCC canonicalizes *pointer + int* to put the pointer first, and the target needs the scaled operand first — it's a one-instruction difference every time:

```
target:  add r4, r3, r29     (3-term sum, scaled operand first)
ours:    add r4, r29, r3     (2-term sum)
```

This also rules out `* 2` and `sizeof(s16)` on that line (they strength-reduce differently) and requires the cast to be to an integer type. `(u32)` works identically to `(s32)` if you'd prefer it.

**The embedded `(offset = ...)` has to stay.** It forces the shift result into a named variable's register so it coalesces with `dest`'s home register. Nine spellings without it were tried — array indexing, plain pointer addition, byte-offset addition, both operand orders — and all land at 25–27 differing rows:

| spelling | rows |
| --- | ---: |
| `(s16*)((offset = count << 1) + (s32) ARR)` | **0** |
| `(s16*)((s32) ARR + (offset = count << 1))` | **0** |
| `(s16*)((offset = count * 2) + (s32) ARR)` | **0** |
| `&ARR[offset = count]` | 27 |
| `ARR + (offset = count)` | 27 |
| `&ARR[count]` | 27 |
| `ARR + count` | 27 |
| `(s16*)((count << 1) + (s32) ARR)` | 25 |

Happy to take `* 2` instead of `<< 1` on the `dest` line if you prefer it — it also matches. I kept `<< 1` so the two lines read as the same shape, since `src` is forced to it.

One other thing while I'm here: the `M2C_FIELD` in the first branch can be expanded to `*(s16*) ((s8*) Toy_sbss_804D6EDC + offset)` at no cost to the match, if you'd rather not have the m2c marker in the file. Rewriting it as `Toy_sbss_804D6EDC[i]` costs 6 rows, so the byte stepping there is load-bearing too. Left it alone to keep this diff small — say the word.

r-burns:
You are still casting the pointer to a scalar type and then back, which is very unsafe. I think you should instruct your model to never do that. The rest of its output sounds like a lot of hallucinations to me.

Danny-Dasilva:
> You are still casting the pointer to a scalar type and then back, which is very unsafe. I think you should instruct your model to never do that. The rest of its output sounds like a lot of hallucinations to me.

Makes sense, running a very tiny model for this on a loop. I'll get this cleaned up

Danny-Dasilva:
Alright ... I updated this to do one conversion instead of two.

MWCC always puts the pointer operand first when the source is pointer math. The target has it the other way round `add r4, r3, r29`, offset first (I put this in the PR description),  and only an addition of two plain integers gives that. I tried `&base[arg0], base + arg0, arg0 + base`, and a `(s8*)` byte offset. Both normalize to pointer-first. 

If you'd rather not have the round trip, I can leave this at 99.73 with plain `&base[arg0]` (I believe it's only a couple byte mismatch anyway). If you have any creative ways to get the order to work let me know, would be a good learning opportunity

ribbanya (approved):
I'm personally fine with this 🤷

It does suck but there are far worse things in the codebase. Surely `Toy_sbss_804D6EDC` is a struct with an s16 array as a field, or something? All of the types are definitely very wrong and that's promoting a lot of hackery.

Xkower8181:
I've tried poking around myself to get something cleaner and technically
`dest = (s16*) &(((s8*) Toy_sbss_804D6EDC)[offset = count << 1]);`
still matches.
However I'm still fighting the reg swap for the add opcode on `src = (s16*) ((arg0 << 1) + base);`
`src = (s16*) ((arg0 << 1) + (s32)base)` does match if we want to keep `base` as `s16*`, which I suppose does help with portability later down the line (if we change s32 into intptr_t for 64-bit compatibility, but I suppose making the code less jank would also do the trick).

r-burns (approved):
Yeah, just wanted to make sure you took a crack at matching this properly with an array access before merging. Definitely plenty on worse stuff in the codebase, and we'll want to fix those at some point, but don't need to do that immediately.

PR #3093, opened 2026-08-15
Improve mnCharSel matching

g-guthrie:
## Summary

- improve matching across `mnCharSel_802640A0`, `mnCharSel_CursorThink`, `mnCharSel_8025C020`, and `mnCharSel_8025DB34`
- recover direct character-select state carriers, pointer walks, stack layout, and integer widths
- update three color-index accessor return types to preserve caller-side integer promotion without changing their generated code

## Matching impact

- `mnCharSel_8025C020`: 97.282555% to 98.919395% (4,516 bytes)
- `mnCharSel_8025DB34`: 87.95396% to 89.04039% (4,952 target bytes)
- additional matching improvements are included for `mnCharSel_CursorThink` and `mnCharSel_802640A0`

## Validation

- exact MWCC 1.2.5n isolated comparisons for the two large functions above
- full-TU compiler comparison confirmed `gm_80169264`, `gm_80169290`, and `gm_801692BC` remain 44 bytes and code-identical after the signature correction
- `clang-format --dry-run --Werror`
- repository source checks
- `git diff --check`

No runtime behavior change is intended; this is a decompilation-accuracy update.

PR #3154, opened 2026-08-18
Match and link mncharsel

MarkMcCaskey:
Time to select a character


MarkMcCaskey:
Just seeing #3093 now, I'll pause work here a bit and pick it up later -- looks like there's minimal overlap so far but there's lots of other matching work to be done. I'll take a look at `lbaudio_ax` for now

ribbanya (dismissed):
Pretty crazy.

MarkMcCaskey:
> Pretty crazy.

Yeah, this one was a real grind. Didn't expect to finish it in this PR but forward progress never really stopped, it was just slow at times.

PR #3163, opened 2026-08-20
Match fn_80193FCC (gmtou_0.c)

Danny-Dasilva:
Matches fn_80193FCC. Uses the named .sdata2 float instead of 0.0F literals and indexes the array directly instead of walking a pointer.


ribbanya (approved):
Normally I would reject a regression from a literal to a named constant but the rest of the floats in the file have to be fixed anyway. I think the sdata2 split is also wrong.

PR #3164, opened 2026-08-20
Use float literals in gmtou_0 and fix its sdata2 symbols

Danny-Dasilva:
Follow-up to #3163. Restores plain float literals in gmtou_0.c (removing the named-constant externs, including the one #3163 added) and localizes the file's sdata2 literal pool in symbols.txt, which the split had promoted to named globals. All matched functions in the unit still score 100.


ribbanya (approved):
Nice, this is perfectly executed. If you didn't know, you can pass `--reloc-diffs all` to `configure.py` to check this kind of thing in `report.json`.

PR #3168, opened 2026-08-21
Match 70 ftCo_0A01 functions via sdata2 pool localization

Danny-Dasilva:
The split had promoted ftCo_0A01's anonymous literal pool to 77 named symbols and its layout is permuted relative to the target, which capped 70 otherwise-matching functions below 100. An sdata2_order() seeder (existing repo idiom) reproduces the target pool layout byte-for-byte and the pool symbols are localized in symbols.txt, taking those 70 functions to 100. Same follow-up class as #3164.


ribbanya (approved):
... Is it referring to bytes as "functions"? It does match sdata2, thoogh.

I guess it's calling each float a function.

PR #3175, opened 2026-08-22
Clean up `build.yml`

ribbanya:


ribbanya:
Well, looks like you can't use variables for container names in a lot of places, so nevermind.

PR #3181, opened 2026-08-22
Link `lbshadow`

JaviSoto:
Matches `lbShadow_8000F38C`.

The remaining diff was register allocation. Keeping the null-initialized `nextGx` value live across the first fighter loop, using self-assignments for the loop and light locals, and reusing `lobj` as the later boolean flag gives MWCC the same register lifetimes as the target.

99.85401 → 100.00 (1,644 bytes)


ribbanya (approved):
Welcome, and nice work!

PR #3183, opened 2026-08-22
Match and link axdriver

MarkMcCaskey:
I’m testing booting the game with non-100% functions and a mode with non-100% TUs trapping. This TU / function is the first part failing on boot.


ribbanya (approved):
"Yeah so I was slightly annoyed with this massive function and just decided to link it 🗿"

PR #3195, opened 2026-08-23
lbarq: fix infinite spin in non-matching builds

blasphemetheus:
## Problem

Non-matching / shifted builds hard-freeze at the start of every vs match (it looks like a freeze on the stage loading screen, on any stage).

### Background: what this affects and why it matters

A *shifted DOL* is what this repo builds when code is added, removed, or compiled in a way that changes function sizes: every function past the change lands at a different address than in the retail binary. The README advertises this as a supported use case ("the DOL the repository builds can be shifted, meaning you are able to add and remove code as you see fit, for modding or research purposes"), and `--non-matching` is the configure mode that produces such builds. Since address-preserving byte-matching is off in that mode, MWCC is free to optimize functions differently than the retail compile did — which is exactly what bites here.

The hang is the blocking wait in `lbArq_80014BD0`:

```c
while (lbArq_80014ABC(rp) != LB_ARQ_STATE_DONE) {
}
```

`lbArq_80014ABC`'s `#pragma dont_inline on` is guarded by `#ifdef MUST_MATCH`. In matching builds the pragma keeps the getter an out-of-line call, so every loop iteration re-reads `rp->state` from memory — that call is what makes the retail busy-wait work, since the field is not `volatile`. In non-matching builds the pragma is compiled out, MWCC inlines the getter and hoists the load out of the loop:

```asm
80014d88: lwz   r0,4(r28)    ; load hoisted out of the loop
80014d8c: cmpwi r0,2
80014d90: bne   80014d8c     ; branches to the cmp, not the lwz
```

so the wait spins forever on a stale register. The state is written by the ARQ completion callback (interrupt context), so unless the transfer completes before the very first read, the loop never exits.

Debugged live via Dolphin's GDB stub on a shifted DOL: PC pinned at `lbArq_80014BD0+0xdc` inside `Fighter_Create` → `ftData` ARAM preload at match start, with the node's state already `LB_ARQ_STATE_DONE` in memory and the ARQ queues drained — the completion had happened; the loop just never re-read it. Since this is the first blocking ARAM transfer the game performs, every non-matching build freezes here regardless of stage or characters.

## Fix

Take the node through a `volatile`-qualified pointer parameter (per review feedback — thanks @ribbanya), so the state read is a volatile access:

```c
static lbArqState lbArq_80014ABC(volatile lbArqNode* arg0)
{
    return arg0->state;
}
```

- **Matching builds:** unaffected — a single load of a `volatile` lvalue compiles to the same `lwz; blr`, and the `MUST_MATCH` pragma path is untouched. Verified: `build/GALE01/main.dol: OK` after the change.
- **Non-matching builds:** `volatile` forbids hoisting, so the load stays inside the loop. Matches now load and play correctly (tested on a shifted DOL in Dolphin 2603).

## Trade-offs / alternatives considered

Reasons to take this form of the fix:

- It states the actual semantics (the field is written from interrupt context and polled from the main thread — a textbook `volatile` situation), rather than relying on an inlining barrier as a side effect.
- Zero impact on the matching target, verified by the sha1 check.
- One line, local to the one wait loop that is known to break.

Reasons the maintainers might prefer something else:

- **Project convention:** if the codebase prefers keeping decompiled sources exactly as the original plausibly looked, the original almost certainly had no `volatile` here (the retail behavior depended on the call boundary). Making the `#pragma dont_inline` unconditional (dropping the `#ifdef MUST_MATCH` guard) reproduces the retail mechanism instead and is equally one-line — I'm happy to switch to that if preferred.
- **Scope:** `volatile` on this getter fixes only this loop. If the project wants a general policy for interrupt-shared state in non-matching builds (other `MUST_MATCH`-guarded `dont_inline` getters exist, though I found no other one guarding a busy-wait on the boot/match path), a broader convention decision may be preferable to a point fix.
- **Field vs. access:** marking the struct field itself `volatile` is arguably cleaner, but it changes codegen at every other access site and risks breaking the byte-match; the qualified parameter confines the change to this getter.


ribbanya:
What about changing the parameter type to `volatile`?

blasphemetheus:
yeah I think that's arguably cleaner, I'll do that, also works

ribbanya (approved):

PR #3198, opened 2026-08-24
lbmthp: fix infinite spin in non-matching builds

blasphemetheus:
Second instance of the bug class fixed in #3195, found by auditing every `#ifdef MUST_MATCH` site for behavioral divergence.

`fn_8001F294`'s `#pragma dont_inline on` is guarded by `MUST_MATCH`, so in non-matching builds it inlines into the wait loop in `lbMthp_8001F800`:

```c
MoviePlayer.unk_70 = 0;
while (fn_8001F294()) {
}
```

The loop body is empty, so after inlining MWCC hoists the load of `MoviePlayer.unk_110` out of the loop entirely:

```asm
8001fee0: lwz   r0,272(r31)   ; load hoisted out of the loop
8001fee4: cmpwi r0,0
8001fee8: bne   8001fee4      ; branches to the cmp, not the lwz
```

`unk_110` is cleared by the `HSD_DevComRequest` completion callback (interrupt context), making this an **intermittent** hang: it triggers only when a THP disc read is still in flight at loop entry — likely why it hasn't been reported from playtesting.

Fix follows the reviewed form of #3195: read through a `volatile`-qualified pointer.

- Matching build: unaffected — `build/GALE01/main.dol: OK` after the change.
- Non-matching build: the load stays inside the loop (`bne` back to the `lwz`), verified in the disassembly.


ribbanya:
I don't mind merging fixes like this, and this specific change seems fine to me, but do you have some kind of demonstrable proof if this is going to be a repeated type of change being PR'd? Maybe a Dolphin execution log or something?

The difficulty is that unlike comparing against a hash like the vast majority of commits made to this repo, you're requesting behavioral changes which _are_ important and desired but at some point must also be verified.

blasphemetheus:
hmm could do a static linter for this class of divergence, if there's interest in that could build it but that'd be like a separate PR probably.

so there's a static answer to this request as pertains to this PR which I will paste below:

> Static (as mechanical as a hash). Build non-matching and disassemble the wait loop in lbMthp_8001F800:
> ▎
> ▎ $ ./configure.py --non-matching && ninja
> ▎ $ powerpc-eabi-objdump -d build/GALE01/main.elf \
> ▎     --start-address=0x8001fed8 --stop-address=0x8001fef0
> ▎
> ▎ ### before this PR:
> ▎ 8001fedc:  90 1f 00 70   stw    r0,112(r31)
> ▎ 8001fee0:  80 1f 01 10   lwz    r0,272(r31)   ; load of MoviePlayer.unk_110
> ▎ 8001fee4:  2c 00 00 00   cmpwi  r0,0
> ▎ 8001fee8:  40 82 ff fc   bne    8001fee4      ; -> the cmpwi, NOT the lwz
> ▎
> ▎ ### with this PR:
> ▎ 8001fedc:  90 1f 00 70   stw    r0,112(r31)
> ▎ 8001fee0:  80 1f 01 10   lwz    r0,272(r31)
> ▎ 8001fee4:  2c 00 00 00   cmpwi  r0,0
> ▎ 8001fee8:  40 82 ff f8   bne    8001fee0      ; -> back to the lwz
> ▎
> ▎ Before the PR the load is hoisted out of the loop, so it spins on a stale register. Since unk_110 is cleared by the HSD_DevComRequest completion callback (interrupt context), the loop can never observe the clear once it's entered — an unconditional hang whenever a read is in flight at entry.

... Seems a little distant from the context needed to like understand why this change might be necessary, so a hopefully more dynamic answer (still not a dolphin log really).

The natural trigger is rare. You'd only hit it in the wild with a THP read (GamCube's native video format is .thp, so reading from the container) still in flight when the video player (Melee's THP player lives in the cdecomp as lbmthp.c, has a struct called MoviePlayer I believe) is told to stop. The THP read happens sometimes when a movie is playing if its large enough it loads the next part while the previous part of the video is playing. So the `unk_110` is what signifies `a read is currently in progress`. I haven't reproduced it with a sweep so have no dolphin log to paste. If tho, you hold `unk_110 = 1` in RAM until you enter the loop, then clear it to 0 (which is what an interrupt does and then sample (with a harness I set up in elixir, which I'm not sure where that should live tbh) ...| (generated text in quotes)
> ▎ ===== PRE-FIX build =====
▎ PC entered loop at 0x8001FEE4 (lbMthp_8001F800+0x38)
▎ stopped holding; set unk_110=0 (readback 0)
▎ post-clear PC readings (12 valid): 12 in loop, 0 outside
▎   samples: 0x8001FEE4 0x8001FEE4 0x8001FEE4 0x8001FEE4 0x8001FEE4 0x8001FEE4
▎            0x8001FEE4 0x8001FEE4 0x8001FEE4 0x8001FEE4 0x8001FEE4 0x8001FEE4
▎ VERDICT: still in the wait loop though unk_110 reads 0 in memory. SPIN.
▎
▎ ===== FIXED build (this PR) =====
▎ PC entered loop at 0x8001FEE0 (lbMthp_8001F800+0x34)
▎ stopped holding; set unk_110=0 (readback 0)
▎ post-clear PC readings (12 valid): 0 in loop, 12 outside
▎   samples: 0x8033CAFC 0x8034A320 0x80019A0C 0x8034E114 0x8033CAFC 0x8034A350
▎            0x803335D8 0x80337E84 0x80334900 0x80334CF0 0x80335948 0x8034E114
▎ VERDICT: left the loop the instant memory cleared. FIXED.

this is all somewhat specific and I'm new to this repo :) I built from main on nix and tried to play a game but it infinite stalled, so my first PR was just the fix an LLM suggested to sidestep a stall. Then I figured I'd build a little harness in elixir (my preferred stack) to find similar bugs to the one I ran into. Found a couple but they're sort of disconnected. I'll figure out a place to put the harness stuff and post that here soon.

blasphemetheus:
told an llm to make a repo, here's the repo https://github.com/blasphemetheus/melee-nonmatching-harness

MarkMcCaskey (commented):
This is interesting, the original code is solving what seems to be a signal-induced write having its read optimized away in

```C
        while (fn_8001F294()) {
        }
```

in `lbMthp_8001F800`. So the fact that that function is not inlined is load bearing but also the wrong way to solve this in C. In fact I believe this is UB without volatile as of C11 (6.8.5p6). 

The fact that the tiny helper is not inlined even though it easily could be is evidence that maybe not inlining was the trick to fix this bug originally, though it could just be a coincidence. Adding an extra volatile as you did or adding volatile on the field inside MoviePlayer both seem to work as well and are strictly more correct here.

As for catching / verifying these cases generally, it's a good question. In this case it's relatively clear why this makes sense statically but I bet that's not always the case. This code is a portability concern but it should be detectable from the CFG. I haven't thought about this enough to know if we can really always expect a comparable CFG when compiling with a modern compiler like Clang and I suspect "comparable" might be doing some heavy lifting but I think this specific case of "gets stuck in an infinite loop locally" should be something that can be detected more easily than the general cases.

ribbanya (approved):
Decided to just change the preprocessor conditions here and in #3195 rather than set the precedent of trying to bugfix on this repo.

PR #3199, opened 2026-08-24
ftCo_ThrownKirby: fix inverted scale copy in non-matching builds

blasphemetheus:

The non-matching branch has the data flow backwards: `HSD_JObjSetScale(jobj, &fp->mv.co.thrownkirby.scale)` writes the **uninitialized** union field into the bone, where the matching branch reads `jobj->scale` into the field (the hand-inlined `HSD_JObjGetScale` the `@todo` on the site describes). Consequences in non-matching builds: the fighter's root JObj scale is clobbered with stale union memory, and the same garbage is consumed later by the `scale.x/y/z` multiplies in this file — broken scaling on Kirby swallow-throws.

Fix: call `HSD_JObjGetScale` in the non-matching branch. Matching build untouched and verified (`build/GALE01/main.dol: OK`); non-matching build compiles.


ribbanya (approved):
Whoops.

PR #3200, opened 2026-08-24
mncount: fix time-string pointer in non-matching builds

blasphemetheus:

`buf` is `char[4]`, so the non-matching vararg `&buf - 4` is `char (*)[4]` arithmetic — it points **16 bytes below** `buf`. The matching expression `*(char (*)[4])&buf - 4` decays to `(char*)buf - 4`, which is where `mn_8022EA78(buf - 4, 2, ...)` wrote the digits two lines above. So non-matching builds hand `%s` a pointer 12 bytes short of the intended string and the VS-records time display prints whatever is on the stack there.

Fix: `(char*) buf - 4` in the non-matching branch — same value the matching expression produces. Matching build untouched and verified (`build/GALE01/main.dol: OK`).

(The site's `@todo` and the deliberate out-of-bounds write below `buf` are retail-faithful weirdness left as-is; this PR only corrects the non-matching pointer to agree with retail.)


ribbanya (dismissed):
Un-draft when ready to merge.

ribbanya (approved):
Nevermind, found a better solution, I'll just merge.

PR #3202, opened 2026-08-24
itzeldadinfireexplode: preserve retail velocity behavior

blasphemetheus:

Retail's `ip->x40_vel.z = ip->x40_vel.y = ip->x40_vel.z = 0.0F` is a typo (doubled `z`, no `x`) — but it's the shipped behavior: Din's Fire explosion X velocity is never cleared. The non-matching branch "fixed" it with `itResetVelocity(ip)`, which zeroes all three components, silently changing the item's physics in non-matching builds.

A decomp should reproduce the game, typos included. Fix: drop the `#ifdef` and keep the retail expression in both modes, with a comment marking the doubled `z` as intentional. Matching build byte-identical (`build/GALE01/main.dol: OK`).


ribbanya (changes requested):
Place the double assignment under the conditional to appease clang linting and drop the `#else`:
```c
#ifdef MUST_MATCH
    ip->x40_vel.z = 
#endif
    ip->x40_vel.y = ip->x40_vel.z = 0.0F;
}
```

ribbanya (approved):

PR #3203, opened 2026-08-24
lbfile: read the DVD-completion flag as volatile

blasphemetheus:
Found in the same `#ifdef MUST_MATCH` audit as #3198 — this one is **hardening, not a live bug**.

`waitForDisc` busy-waits on the static `cancel` flag, written by the `HSD_DevComRequest` completion callback (interrupt context) — the same interrupt-shared-state shape as #3195 and #3198. It doesn't currently hang in non-matching builds only because `discIsDone`'s body contains the extern call `lb_800195D0()`, which forces MWCC to reload globals each iteration. That's a codegen accident: `cancel` has internal linkage and its address is never taken, so any refactor of that call (or smarter escape analysis) turns this into the lbarq/lbmthp hang.

Fix: read the flag through a volatile cast, making the reload required rather than incidental. Matching build byte-identical (`build/GALE01/main.dol: OK`).

Happy to drop this one if the project prefers to only fix demonstrated hangs — but given two shipped instances of this exact class, the cheap hardening seems worth it.


ribbanya (changes requested):
Unless you can find a solution that matches and gets rid of the pragma in matching builds, just change the directives to `MWERKS_GEKKO` like in #3198.

ribbanya (approved):

PR #3210, opened 2026-08-25
Add broad data matching pass

MarkMcCaskey:
I let Claude do some jump table stuff limited to places where it cleans things up entirely. Can remove if that's annoying to do before linking


ribbanya:
It's not that jump tables are annoying, I just thought it might be a sticking point.

MarkMcCaskey:
I'll iterate on this for a bit to recover the lost function matches and get more data matches too

src/sysdolphin/baselib/hsd_3915.c
```diff
@@ -482,12 +481,622 @@ void hsd_80392194(u8* dst, s32 flags, void* unused1, void* unused2,
...
-extern GlyphEntry lbl_80408898[4];
+GlyphEntry lbl_80408898[4] = {
+    { 0x10808000, hsd_80392194 },
+    { 0x46808000, hsd_80392194 },
+    { 0x7C808000, hsd_80392194 },
+    { 0xB3808000, hsd_80392194 },
+};
+
+/// Debug-font glyph bitmap.
+u8 lbl_804088B8[0x1c00] = {
```
ribbanya: Copy the structure of `HSD_SisLib_FontAtlas`, including in `config.yml`.

ribbanya (dismissed):

src/melee/mn/mnnamenew.c
```diff
@@ -102,6 +102,222 @@ static AnimLoopSettings mnNameNew_803EDA58[3] = {
 /// Contents restored from the target unit's .data section (symbols.txt:
 /// mnNameNew_KeyMap at 0x803EDA7C size 0x268, mnNameNew_GlyphTable at
 /// 0x803EDCE4 size 0x64C); see build/GALE01/asm/melee/mn/mnnamenew.s.
+/// Kana/latin glyph literals for the name-entry keyboard; each is a
+/// Shift-JIS character referenced individually by mnNameNew_KeyMap.
+SDATA char mnNameNew_804D4C14[] = "\x82\xA0";
```
ribbanya: - Use unicode in the source file; it will be emitted as sjis.
- `wchar_t` is intended for this purpose.
- `SDATA` should not be necessary as each character is <8 bytes
```suggestion
wchar_t mnNameNew_804D4C14 = L'あ';
```

ribbanya (changes requested):

ribbanya (dismissed):
The glyph table uses pooled sjis string literals. Adapt this pattern and remove the named definitions.
```c
static MnNameNewKeyMap mnNameNew_KeyMap = {
    { 3, 10, 11, 7, 9, 8, 5, 5 },
    {
        "あ", "い", "う", "え", "お", "か", "き", "く", "け", "こ",
        "さ", "し", "す", "せ", "そ", "た", "ち", "つ", "て", "と",
        "な", "に", "ぬ", "ね", "の", "は", "ひ", "ふ", "へ", "ほ",
        "ま", "み", "む", "め", "も", "や", "ゆ", "よ",
    },
};
```
```symbols.txt
mnNameNew_804D4C14 = .sdata:0x804D4C14; // type:object size:0x3 scope:local data:sjis
```

ribbanya:
Try also marking `NonMatching` units as `Testing` in `configure.py`.

PR #3211, opened 2026-08-25
Match functions needed to boot to the menu

MarkMcCaskey:
Doing a broad pass on the remaining functions here to see if we can just get the game booted on 100% matched functions

Screenshot 2026-08-24 at 7 24 04 PM


src/sysdolphin/baselib/hsd_3A94.c
```diff
@@ -2488,42 +2621,187 @@ static inline void fn_803ADF90_calc_file_blocks(s32 file_size, s32 file_idx,
...
+    raw[-23 + 0] = 4;
+    raw[-23 + 1] = (s32) state;
+    raw[-23 + 4] = 0;
+    raw[-23 + 5] = 0;
+    raw[-23 + 6] = (s32) dst;
+    raw[-23 + 8] = size;
+    raw[-23 + 7] = 0;
+    raw[-23 + 2] = 0;
+    return fn_803AC168((s32*) ((u8*) &raw[0] - 92));
+#else
```
MarkMcCaskey: What a match

r-burns: I appreciate you keeping around the #else case here, for really unsafe code like this I think it's a good idea to keep the simpler/saner nonmatching version around as it's more readable and certainly more portable

MarkMcCaskey: For sure! It's an ugly hack, I'm going to try hard to fix this before shipping. Even for the matching build this is pretty unacceptable C. If we accept this code then we can cheat the stack layout generally which seems like a bad precedent unless there's really no other way.

ribbanya: `MUST_MATCH` implies `__MWERKS__` though so it should just be
```c
#ifdef MUST_MATCH
```

MarkMcCaskey: Sounds good. After doing another pass here, it seems like a big matching effort to solve this. And I was unable to get another match on that function with anything resembling sane code. If this is alright, I'll ship it as-is with the todo comment for now and come back later. Otherwise I'll happily cherry pick those hacks out and leave only properly matched functions.

The whole structure here seems suspect, returning references to locals on the stack doesn't seem right.

The specific issue is that there are 3 specific places on the stack that CardCmds should be but every solution either coalesces them into the same slot or uses additional registers or puts them in the wrong place on the stack.

I suspect the function may need to be more or less entirely rewritten to solve this at this point and none of my initial attempts to do that worked. This behavior of stack layout isn't something I've seen before with MWCC.

ribbanya: Matching hack is better, especially if it's guarded.

PR #3213, opened 2026-08-26
Split out gm_16A2

r-burns:
lbl_8046B488 appears to be the start of a file's bss section, based on it being used as a base offset for bss accesses. Since there is already float duplication in 1601 that will require splitting, let's split out a new file based on this bss boundary.


r-burns:
Bot is not gonna be useful here, so here's the file-rename-aware report:

```
./tools/diff_symbols.py -p ne build/GALE01/baseline.json build/GALE01/report.json
fn_80167638 | fuzzy_match  |   99.11 -> 99.08
fn_80169000 | fuzzy_match  |   98.09 -> 97.94
fn_80169A84 | fuzzy_match  |   96.65 -> 95.65
gm_8016A22C | fuzzy_match  |   93.58 -> 95.31
fn_8016A4C8 | fuzzy_match  |   98.71 -> 99.66
```

PR #3218, opened 2026-08-26
Match fn_8016A4C8

Danny-Dasilva:
> Matches fn_8016A4C8.

### DIFF
--- a/src/melee/gm/gm_16A2.c
+++ b/src/melee/gm/gm_16A2.c
@@ -706,9 +706,9 @@ void fn_8016A488(int arg0)
     }
 }
 
-static inline int fn_8016A4C8_spawn_offset(void)
+static inline int getSpawnPointIndex(int spawn_slot)
 {
-    return lbl_8046B488.xA - 1;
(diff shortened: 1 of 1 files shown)


src/melee/gm/gm_16A2.c
```diff
@@ -736,17 +736,33 @@ static inline int roll_cpu_type(void)
...
+    if (Player_GetPlayerSlotType(matching_slot) != Gm_PKind_NA &&
+        Player_GetFlagsBit1(matching_slot) == 0 &&
+        chr == Player_GetPlayerCharacter(matching_slot) &&
+        costume_id == (s32) Player_GetCostumeId(matching_slot))
+    {
+        return 1;
+    }
+    matching_slot += 1;
+    if (matching_slot < 6) {
+        goto find_matching_controller;
```
r-burns: Still matches after cleaning this up to be a for loop, so let's do that

r-burns (approved):
Nice! Much cleaner, this all looks very idiomatic now

PR #3233, opened 2026-08-27
Recover linkable matches

MarkMcCaskey:
We lost half a % or so marking TUs linkable. Pass to recover those matches


MarkMcCaskey:
@ribbanya can you confirm these changes don't step on anything you're doing? Don't want to undo anything you're relying on.

ribbanya (approved):
Yeah this is good, thanks. The matches were always going to need to be repaired.

ribbanya:
The named floats are a bit cringe but that will have to be resolved before linking (`Matching`) anyway.

PR #3250, opened 2026-08-28
Improve psDispParticles match to 99.59%

MarkMcCaskey:
This is going to be a long one. We'll need to regress a lot to get where we need to be

I'm not going to finish this right now so I'll just ship the improvements and we can pick this up later. This function is at the point where it generally can be hill-climbed. Impossible to say if we'll need to find a new path later but it seems to be on the right track now.


ribbanya:
Regress away.

PR #3268, opened 2026-08-30
CmSubject cleanup and documentation

sadkellz:


ribbanya (approved):
The only thing I want to point out is that `@sz` and `@at` are very old and we should probably get rid of them in favor of clang static analysis. You can also use `offsetof` if needed, maybe I'll make that a macro, but now that most things are linked I'd like to just let the dol decide what goes where.

In other words, it's not hurting anything but don't waste your time on it.

PR #3270, opened 2026-08-30
Improve mnStageSw_80235C58 to 99.51%

itsgrimetime:
## Summary

- Improve `mnStageSw_80235C58` from 98.96% to 99.51% by expressing the range scan and nearest-enabled-stage search directly.
- The remaining seven differences are an `r24`/`r31` register-allocation swap.

## Verification

- `python configure.py && ninja -j1`
- `mnStageSw_80235C58`: 99.51%, opcode-identical with register-only differences

### DIFF
--- a/src/melee/mn/mnstagesw.c
+++ b/src/melee/mn/mnstagesw.c
@@ -151,33 +151,21 @@ static void mnStageSw_802359C8(MnStageSwData* data)
     }
 }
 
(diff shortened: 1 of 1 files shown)


itsgrimetime:
oof maybe those other ones were already matched 1 sec

itsgrimetime:
rip others beat me to it lol. i guess improvement to `mnStageSw_80235C58` is still a win

PR #3281, opened 2026-09-01
Improve `Ground` at 0x40

Danny-Dasilva:
The `fn_801C8EF8` match from this PR is already on master, so this is now just the struct improvement.

`Ground` had raw fields covering 0x40-0xC0 (`self_vel`, `cur_pos`, `x58`..`x70`, `pad_74`). That range is a `ColorOverlay`, so replace them with a single `color_overlay` member. `grMaterial_GetOverlay` no longer needs the `(u8*) gp + 0x40` cast, and `gp->x6C` becomes `gp->color_overlay.x2C_hex`.

No codegen change: every unit and function in the report is identical to master.


ribbanya:
This function is matched on master now but the improvement to `Ground` at 0x40 is worth merging.

PR #3299, opened 2026-09-03
Move sysdolphin/baselib/ to hsd/

r-burns:
Drafting for now to allow some discussion, and if approved, defer to other PRs currently modifying HSD code to avoid conflicts.


r-burns:
Closing in favor of https://github.com/doldecomp/melee/pull/3319

The sysdolphin/baselib path, while verbose, is at least somewhat correct and seems to be our best guess at what they actually used.

PR #3307, opened 2026-09-04
Improve 3 functions, match 1

titus-ssbm:
Fixes four integer type declarations that caused MWCC to emit width and
signedness conversions that retail does not have.

When an argument is passed to a parameter narrower than itself, MWCC emits a
conversion at the call site (`extsb`, `extsh`, `clrlwi`). When the declared type
is wide enough, no conversion is emitted. Retail's codegen therefore pins the
width of each parameter: a plain register copy where we emit a mask means the
declaration is too narrow, and a mask where we emit a plain copy means it is too
wide. This is the `s8`-is-usually-an-`int` case described in CONTRIBUTING.

- `gm_801B5324`: `arg1` widened to `s32`. The indirect call through `x54` now
  narrows its first argument as retail does. Matches.
- `fn_8016B388`: `arg1` widened to `int`, removing an `extsh` in the caller
  `fn_801891F4` (99.517 -> 99.786).
- `mn_80230D18`: `arg2` widened to `int`, removing an `extsb` in the caller
  `mn_80230E38` (99.843 -> 99.961).
- `ifStock_802F8298_get_flag`: return type changed to `u8`. The getter and a
  later direct read of the same byte are one common subexpression, so the
  getter's `int` was carried into a comparison that retail performs unsigned
  (`cmpwi` -> `cmplwi`), improving `ifStock_802F8298` (97.064 -> 97.191).

`gm_801B5324` is the one case where `s32` and `int` are not interchangeable: it
matches with `s32` and regresses with `int`.

Claude Opus 5 did the function matching work.

### DIFF
--- a/src/melee/gm/gm_16AE.c
+++ b/src/melee/gm/gm_16AE.c
@@ -293,7 +293,7 @@ void gm_8016B378(s8 arg0)
     gm_16AE_GetUnkData_0()->unk_18 = arg0;
 }
 
-void fn_8016B388(int arg0, s16 arg1)
+void fn_8016B388(int arg0, int arg1)
 {
     gm_16AE_GetUnkData_0()->FighterMatchInfo[arg0].x6 = arg1;
 }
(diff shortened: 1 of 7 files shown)


MarkMcCaskey (approved):
Nice!

PR #3310, opened 2026-09-04
Match 4 functions and link toy

MarkMcCaskey:


tools/set_section_align.py
ribbanya, line 1: Absolutely unacceptable solution. If something like this is warranted, it should either be a [dtk](https://github.com/encounter/decomp-toolkit) PR or a mwcc patch like the `1.2.5n` patch.

I haven't looked in depth at this yet but I am extremely skeptical anything like this or the above are needed.

ribbanya (changes requested):

ribbanya (approved):

PR #3311, opened 2026-09-04
Match gm_8016A22C

itsgrimetime:
## Description

Matches `gm_8016A22C` by correcting the sixth parameter type from `int` to `u8`, producing the retail stack-frame layout. The declaration is updated consistently in the header, and the obsolete widening cast at the `gm_8017CE34` call site is removed to preserve that caller's exact argument-load order.

## Verification

- Full `ninja` build (`main.dol: OK`)
- `gm_8016A22C`: 100.0% fuzzy match (472 bytes)
- `gm_8017CE34`: 100.0% match (2,424 bytes)
- `clang-tidy`, `clang-format`, `editorconfig-checker`, and `style-check`


itsgrimetime:
> 🥀 1 broken match

😠
will fix

PR #3312, opened 2026-09-04
Match CompareNameStrings

itsgrimetime:
## Summary
- correct the function boundary so the compiler-generated final `blr` belongs to `CompareNameStrings`
- remove the false standalone `fn_802377A4` split and its unused declaration
- label the local Shift-JIS space literal with its compiler-generated symbol

## Verification
- clean build from `upstream/master`
- `CompareNameStrings`: 100.0% (340 bytes)
- full local build: `build/GALE01/main.dol: OK`
- pre-commit formatting, editorconfig, style, symbol, signature, and match-regression checks passed


itsgrimetime:
fixing busted match

itsgrimetime:
The diff bot’s reported `fn_802377A4` regression is the boundary correction itself, not lost code. At retail address `0x802377A4` there is one `blr`; MWCC emits it as the terminal epilogue of `CompareNameStrings`, giving that function an object size of `0x154`. Keeping the old empty stub emits a second `blr`, shifts `IsNameUnique` by four bytes, and makes the TU layout wrong. Removing the false split yields the exact retail byte layout, keeps `IsNameUnique` matched, passes the full link/hash, and nets +336 matched bytes (340 gained minus 4 reclassified).

itsgrimetime:
thanks codex

PR #3313, opened 2026-09-04
mn: Move menu kinds into kinds directory

itsgrimetime:
## Summary

- Group 22 MenuKind implementation TUs and their headers under `src/melee/mn/kinds`, following `ft/kinds` and `it/kinds`.
- Keep shared menu infrastructure (`mnmain`, `forward.h`, `types.h`, and `inlines.h`) at the `mn` root.
- Keep `mncharsel` and `mnstagesel` at the root because they are scene handlers rather than MenuKind implementations.
- Move `mndiagram2` and `mndiagram3` with `mndiagram` as internal helpers for that kind.
- Update include paths and build metadata for the relocated files.

This is a draft to get feedback on whether this is the right boundary for the menu subsystem.

## Verification

- `python configure.py && ninja`
- `build/GALE01/main.dol: OK`
- Normalized `mn` unit match metrics are unchanged.
- 21 relocated object files are byte-identical; `mnmainrule.o` differs only in line/debug metadata from include sorting and is identical after stripping debug sections.


ribbanya (changes requested):
Where in the codebase is the state machine? Which functions in these files are state callbacks?

itsgrimetime:
Closing after a closer inspection of the architecture. The menu-kind state table and its behavior callbacks are confined to mnmain.c; the modules moved here are invoked directly as leaf screens rather than registered as kinds/subclasses in that state machine. A kinds directory therefore implies a relationship that the current code does not support clearly enough.

PR #3314, opened 2026-09-04
mn: Model scrolling list menu data inheritance

itsgrimetime:
## Summary

- Model the shared eight-byte scrolling-list state as a first-member base struct.
- Embed that base in MnInfoData and MnCountData.
- Preserve the existing specialized layouts and direct leaf-menu dispatch.

## Stack

Depends on #3313.

This branch contains the file-move commit from #3313 followed by one new commit, 5a6401a0a8. Until #3313 merges, GitHub may show the moves as part of this diff; the new semantic change is confined to that top commit.

## Verification

- Full configure and build completed successfully.
- main.dol checksum matches.
- Global match comparison reports no changes.
- mninfo and mncount remain 100% code and data matched.


itsgrimetime:
Verified the intended pattern against gmscdata/gm_1A3F.c and ftdata/fighter.c. The relevant mn analogue is already the MenuKind-indexed mn_803EB6B0 descriptor array together with MenuFlow/MenuState and the centralized transition logic in mnmain.c. This PR instead modeled first-member struct-prefix inheritance, which is a different concept and is not supported by that analogy. Closing this follow-up; #3313 remains the relevant organizational change.

PR #3317, opened 2026-09-04
Rename fighter kind TUs to match HAL convention

ribbanya:
Initial naming pass before getting rid of the individual character subdirectories. Still needs more scrutiny on the individual state names.

Renames generated using this script:
https://gist.github.com/ribbanya/d8f2f1a4332769dcf5e7ee900f7dba1b

Which generates a bash script leveraging `rename_tu` in `tools/functions.sh`


.nix/CMakeLists.txt
```diff
@@ -63,13 +63,13 @@ set(SOURCES
...
-    src/melee/ft/kinds/ftCaptain/ftCa_SpecialN.c
-    src/melee/ft/kinds/ftCaptain/ftCa_SpecialS.c
-    src/melee/ft/kinds/ftCLink/ftCl_AppealS.c
-    src/melee/ft/kinds/ftCLink/ftCl_Init.c
+    src/melee/ft/kinds/ftCaptain/ftcaptain.c
+    src/melee/ft/kinds/ftCaptain/ftcaptainspecialhi.c
+    src/melee/ft/kinds/ftCaptain/ftcaptainspeciallw.c
+    src/melee/ft/kinds/ftCaptain/ftcaptainspecialn.c
+    src/melee/ft/kinds/ftCaptain/ftcaptainspecials.c
+    src/melee/ft/kinds/ftCLink/ftclinkappeals.c
```
r-burns, line 71:
```suggestion
    src/melee/ft/kinds/captain/ftcaptainspecials.c
    src/melee/ft/kinds/clink/ftclinkappeals.c
```

WDYT about also renaming the directories like this?

ribbanya: I didn't rename the directories because I don't want directories. `ft/kinds/ftcaptainspecials.c`, etc. Thoughts?

r-burns: Oh ok, sounds good to me

PR #3319, opened 2026-09-04
 Always use full sysdolphin include paths 

r-burns:


ribbanya (approved):
What was your method?

r-burns:
Just a bunch of sed commands, haha. Replaced `"baselib/` with `"sysdolphin/baselib/` and then a small script that did

```fish
for f in $files
    set fn "sysdolphin/baselib/$f"
    sed -i "s,\"$fn\",<$fn>,g" src/**.{c,h}
end
```

PR #3328, opened 2026-09-05
Improve hsd_3B5C

MarkMcCaskey:
image

gpt-6 knows what a JPEG is. I just want a picture of god dang hotdog


MarkMcCaskey:
Matching progress has stalled here. Need to try again another time.

Regardless this is a really nice clean up. Ideal problem for LLMs to work on as they're trained on every JPEG decoder and spec available I'm sure.

The last match seems like it shouldn't be hard here but I was unable to get it this time.

PR #3337, opened 2026-09-05
Match and link mnmainrule

MarkMcCaskey:


MarkMcCaskey:
Some of the naming choices here need more review. I'll see if I can look into this tomorrow and verify

itsgrimetime (approved):
I'm fine with the names - they don't seem entirely wrong, and we're gonna have to come up with a system for marking a function/TU's names/documentation as "validated" or something as we go through those phases, so even if they're a little off they'll get corrected

ribbanya (approved):
The names are fine. They're just locals and can be double-checked later.

PR #3339, opened 2026-09-05
Work on lbSnap_8001DA5C

itsgrimetime:
Improve `lbSnap_8001DA5C`, the RGB565-to-RGB5A3 snapshot banner conversion, from 67.65% to 92.69% matching. Rework the tiled pixel offsets and row accumulator while preserving the two-pixel loop, and use descriptive coordinate names.

Based on the [original scratch by Xkower8181](https://decomp.me/scratch/yY6Ca). Credit to Xkower8181 for the starting implementation.

The function remains unmatched: register allocation, one instruction-order difference, and the stack frame still differ from the target.

Validation: full build with required prototypes, formatting checks, and 16 full-image comparisons against the previous implementation under address and undefined-behavior sanitizers, including unchanged banner margins.


Xkower8181:
Improved the match to 93.41% (94.12% if accounting for the PAD_STACK), albeit I got no clue how to move that one addze opcode into the correct place.
[https://decomp.me/scratch/dtUkF](url)

PR #3341, opened 2026-09-05
Match fn_80169A84

itsgrimetime:
## Description

- Matches fn_80169A84 at 100.00%.
- Corrects fn_80169A84 and gm_8016A22C to return void.
- Narrows temporary lifetimes and uses an indexed sentinel-fill loop to reproduce the original code generation.

## Verification

- python configure.py && ninja
- GALE01 main.dol checksum passes.


itsgrimetime:
oh actually I think this can be linked now - will update

itsgrimetime:
nvm theres one more function w/ a regswap left. will try but no promises

PR #3343, opened 2026-09-05
improve gmresultplayer

jurrejelle:
Can probably match the rest of the TU with enough time but just wanted to commit/push what I had so there's visibility / no duplicated work


jurrejelle:
Moved my attempts at the remaining function off of this branch and onto another branch for later, putting this up for review

PR #3344, opened 2026-09-05
Match and link psdisp

MarkMcCaskey:
one more for the road.

I'm letting an agent run on this for a bit longer

Okay, I got it. I removed discord from my phone for now but actually signing off and stopping now 😆 for real this time


ribbanya:
@itsgrimetime can you resolve this?

MarkMcCaskey:
I'm going to remove some of the refactoring done on master because it breaks the match by a lot

MarkMcCaskey:
I'm resolving it now too

MarkMcCaskey:
Okay should be resolved. We can ship as-is and then iterate to regain any refactoring / code clean ups there. They may be workable but even small diffs in this function can be a big grind to fix.

MarkMcCaskey:
Best of luck finishing melee! You all are definitely going to have it done before I'm back 😄

PR #3346, opened 2026-09-05
Match and link mnruleplus

itsgrimetime:
Complete and link `mnruleplus` as a matching translation unit. Match `mn_80233218`'s visibility-count helpers and stack layout, and use the existing named assertion strings.

Let the digit-animation helpers inline automatically and place their index tables in source order so the compiler emits the original constant pool. Remove redundant constant definitions and obsolete header declarations, and correct the two `.sbss` declarations to preserve the linked data layout.

Verification with `--no-always-apply --sym off` and relocation-sensitive data comparisons:

- `mn_80233218`: 100%.
- All nine functions and all data sections in `mnruleplus`: 100%, including `mn_80232D4C`, `.sdata`, and `.sdata2`.
- Full build with `mnruleplus` linked from C and the `main.dol` SHA-1 check passed.
- `ninja diff` passed against the original binary.


itsgrimetime:
will fix this one up

PR #3360, opened 2026-09-06
Match, Split and Link gmresultplayer

jurrejelle:


ribbanya:
Linkable

jurrejelle:
Yep that's the next step :D

PR #3364, opened 2026-09-06
Port `format_percent` from decomp.dev

ribbanya:
https://github.com/encounter/decomp.dev/blob/df4331f13f177d5d5d71b7d0ce31c76ba3673284/crates/core/src/util.rs#L75-L82


ribbanya:
Submitted upstream to https://github.com/encounter/dtk-template/pull/79

PR #3366, opened 2026-09-06
Document fighter and ground alloc bug

sadkellz:


ribbanya (changes requested):
I think it might be better for mid-function comments not to be doc comments. Let me check what the rendered version looks like.

ribbanya (approved):
eh nah it's fine

Image

PR #3371, opened 2026-09-06
Match and link mnvibration

itsgrimetime:
Matches `mnVibration_HandleInput` and `mnVibration_Think`, completing all 12 functions in `mnvibration.c`, and marks the translation unit Matching.

The handler scopes cursor objects per movement branch and uses axis-specific translation helpers. The Think changes reuse the port counter, share child traversal, and use a float animation state; these changes are adapted from #3358. Thank you to @fjooord for making that work available.

Linking also requires retaining the indirectly referenced asset-name strings and making the four joint-asset blocks file-local so they keep their original BSS order. The original stack reservation is retained in Think; removing it or restoring the out-parameter wrapper changes the generated frame.

Validation: all 12 functions (7,688 code bytes) and 688 data bytes match at 100%. `python configure.py && ninja` passes the original GALE01 DOL checksum with the TU set to Matching.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1335,7 +1335,7 @@ config.libs = [
             Object(Linkable, "melee/mn/mndiagram.c"),
             Object(Matching, "melee/mn/mndiagram2.c"),
             Object(Matching, "melee/mn/mndiagram3.c"),
-            Object(Linkable, "melee/mn/mnvibration.c"),
+            Object(Matching, "melee/mn/mnvibration.c"),
             Object(Matching, "melee/mn/mnsound.c"),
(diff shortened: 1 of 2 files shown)


itsgrimetime:
Oh @fjooord has this in https://github.com/doldecomp/melee/pull/3358 already

itsgrimetime:
Oh actually I was able to get the Input one too

itsgrimetime:
Linking incoming

ribbanya:
> Oh @fjooord has this in #3358 already

You should not prioritize Ford's PRs over yours. He'll just merge.

PR #3386, opened 2026-09-07
Fix progress report paths when run outside the repository

CarlosLavayenJr:
> AI-assisted summary: Open the supplied report path directly instead of rebasing it against the repository root. This fixes FileNotFoundError for both absolute and relative report paths when changes_fmt.py is invoked from another directory. Remove the unused path helpers and add CLI regression coverage, including paths containing spaces.
>
> Validation: `python -m unittest discover -s tools/tests` passes all three subcases on Windows with Python 3.13.1. Both external-directory cases failed before the fix; the repository-root case passed. `git diff --check` passes. No game source changes or full game build.

### DIFF
--- a/tools/changes_fmt.py
+++ b/tools/changes_fmt.py
@@ -1,14 +1,9 @@
 #!/usr/bin/env python3
 
 from argparse import ArgumentParser
-import os
 import json
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


ribbanya:
Please submit upstream to https://github.com/encounter/dtk-template.

PR #3394, opened 2026-09-07
Use `matched` instead of `complete` for fancy progress

ribbanya:
Mirrors the category totals.


ribbanya:
Tracked upstream in https://github.com/encounter/dtk-template/pull/80.

PR #3402, opened 2026-09-07
Apply tools/replace-includes.py across src

vagnertxr:
- **Apply tools/replace-includes.py across src**
- **Apply tools/replace-includes.py to mnsnap and hsd_3B34**


ribbanya (approved):
Sure, but we still need to add it to pre-commit (#1565).

PR #3410, opened 2026-09-07
Improve register allocation in fn_803B6820 and mnSnap_80257F24

zapatagustin:
regalloc tweaks for the two remaining functions. neither matches yet, both
units stay Linkable.

fn_803B6820: 98.90% -> 99.05%. `cr = (cb = ...)` puts cb in r18 like the
target; the s64 cast and the `& ~0ull` mask move the other temps around.
what's left is a register permutation (r15/r16/r20).

mnSnap_80257F24: 97.79% -> 99.68%. accessing mnSnap_804A0A10 directly in the
thumbnail/label/count loops fixes both inlines in the second half; the
self-assignment after memzero compiles to nothing but fixes the stack frame.
what's left is the arg setup for the lbArchive_LoadSections call.

numbers are objdiff, didn't run the full build.

### DIFF
--- a/src/melee/mn/mnsnap.c
+++ b/src/melee/mn/mnsnap.c
@@ -2588,6 +2588,8 @@ void mnSnap_80257F24(void)
     lb_8001CDB4();
     memzero(&mnSnap_804A0A10, sizeof(mnSnap_State));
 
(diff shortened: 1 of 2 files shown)


jurrejelle:
These units have since been matched, thank you for your contribution

Melee is 100% matched :D

Issue #3411, opened 2026-09-07
Switch `--reloc-diffs` over to `all`


ribbanya:
There are several functions blocking this, need to investigate each one, but it should be doable.
```
[4/4] CHANGESFMT
                                                     Total | fuzzy_match  |  99.99% ->  99.99%
                                                     Total | matched_code | 100.00% ->  99.11%
                                  main/melee/pl/plbonuslib | fuzzy_match  | 100.00% ->  99.98%
                                  main/melee/pl/plbonuslib | matched_code | 100.00% ->  76.47%
                                               pl_8003ED0C | fuzzy_match  | 100.00% ->  99.93%
                                               fn_8003EE2C | fuzzy_match  | 100.00% ->  99.98%
                                               fn_8003F654 | fuzzy_match  | 100.00% ->  99.86%
                                               pl_8003FAA8 | fuzzy_match  | 100.00% ->  99.94%
                                               pl_800401F0 | fuzzy_match  | 100.00% ->  99.69%
                                               pl_80040614 | fuzzy_match  | 100.00% ->  99.72%
                                               pl_800407C8 | fuzzy_match  | 100.00% ->  99.76%
                                               pl_80040948 | fuzzy_match  | 100.00% ->  99.79%
               main/melee/ft/kinds/ftCommon/ftCo_AttackLw3 | fuzzy_match  | 100.00% ->  99.97%
               main/melee/ft/kinds/ftCommon/ftCo_AttackLw3 | matched_code | 100.00% ->  87.85%
                                                   doEnter | fuzzy_match  | 100.00% ->  99.71%
               main/melee/ft/kinds/ftCommon/ftCo_AttackAir | fuzzy_match  | 100.00% ->  99.95%
               main/melee/ft/kinds/ftCommon/ftCo_AttackAir | matched_code | 100.00% ->  83.42%
                          ftCo_AttackAir_GetMsidFromCStick | fuzzy_match  | 100.00% ->  99.77%
                              ftCo_AttackAir_EnterFromMsid | fuzzy_match  | 100.00% ->  99.60%
              main/melee/ft/kinds/ftCommon/ftCo_LandingAir | fuzzy_match  | 100.00% ->  99.88%
              main/melee/ft/kinds/ftCommon/ftCo_LandingAir | matched_code | 100.00% ->  73.60%
                          ftCo_LandingAir_EnterWithMsidLag | fuzzy_match  | 100.00% ->  99.55%
[...]

PR #3432, opened 2026-09-09
Clean up constant emission in matched code

keyz:
> Remove unused padding initializers and duplicate constants, keep Yoshi motion flags local, and inline the item scale helper to avoid unnecessary emitted data and code.
>
> Keep the existing relocation comparison default, SDK function order, sqrt/sqrtf definitions and calls, and OSDB relocations unchanged. This keeps the cleanup from #3411 while leaving the remaining strict-comparison blockers for separate work.
>
> Verified all 19,828 functions still match with the existing comparison default, with zero regressions or missing functions. Prototype/link checks pass, and the DOL is byte-identical.
>
> Implemented and verified with Codex + Astra.

----

Disclosure: I contribute to this project in my personal capacity, on my own time and using my own laptop. I am not acting on behalf of my employer.

### DIFF
--- a/extern/dolphin/include/libc/math.h
+++ b/extern/dolphin/include/libc/math.h
@@ -29,8 +29,8 @@ double floor(double x);
 
 extern inline float sqrtf(float x)
 {
-    const double _half = .5;
-    const double _three = 3.0;
+    static const double _half = .5;
+    static const double _three = 3.0;
     volatile float y;
     if (x > 0.0f)
     {
@@ -52,8 +52,8 @@ extern inline float sqrtf(float x)
 // TODO: this isn't correct! It's just to generate sdata2 in GXDraw.o
 extern inline float sqrt(float x)
 {
-    const double _half = .5;
-    const double _three = 3.0;
+    static const double _half = .5;
+    static const double _three = 3.0;
     volatile float y;
     if (x > 0.0f)
     {
--- a/extern/dolphin/include/libc/math.h
+++ b/extern/dolphin/include/libc/math.h
@@ -29,8 +29,8 @@ double floor(double x);
 
 extern inline float sqrtf(float x)
 {
-    static const double _half = .5;
-    static const double _three = 3.0;
+    const double _half = .5;
+    const double _three = 3.0;
     volatile float y;
     if (x > 0.0f)
     {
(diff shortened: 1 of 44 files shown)


src/melee/it/kinds/itflipper.c
```diff
@@ -174,7 +172,7 @@ static inline f32 itFlipper_SpinSpeedFromFighter(Item_GObj* gobj,
     ftLib_800866DC(fighter, pos);
     ftLib_80086BEC(fighter, vel);
     return attrs->x18_spinMultiplier *
-           sqrtf(vel->x * vel->x + vel->y * vel->y);
+           sqrtf__Ff(vel->x * vel->x + vel->y * vel->y);
```
r-burns: We definitely don't want to directly call the mangled C++ name here - that's a portability issue, and we should use its standard C name when possible

extern/dolphin/include/libc/math.h
```diff
@@ -29,8 +29,8 @@ double floor(double x);
 
 extern inline float sqrtf(float x)
 {
-    static const double _half = .5;
-    static const double _three = 3.0;
+    const double _half = .5;
+    const double _three = 3.0;
```
r-burns: We shouldn't mess with the sqrtf definition - this is known to match

r-burns (dismissed):
Thanks!

Most of the changes here are good, but I think this check is a little too aggressive to enable by default. It looks like most of the tweaks are to manually elide temporaries that already get elided during linking, so for a lot of these cases I think letting the linker take care of it is fine.

r-burns:
Specifically, I think we should not worry about the (vendored) dolphin extra functions, not touch sqrt/sqrtf, and not worry about the OSDB relocs, and we should probably not enable this behavior by default yet. Everything else looks good and mergeable though.

ribbanya:
When you say "yet," what needs to happen first? Or should we just close #3411?

r-burns:
It's definitely a good default on the majority of code, but I think we should look into those blockers I listed and see if we should either resolve them in a different way, or selectively disable the check on them somehow, or some other solution.

keyz:
thanks @r-burns! learned a ton from your feedback. just updated the PR

PR #3439, opened 2026-09-09
mncharsel cleanup

sadkellz:
remove casts and annoying defines and replace with proper struct for the css data table layout


ribbanya:
You can just auto-squash most PRs now that any regression breaks CI.
image

sadkellz:
> You can just auto-squash most PRs now that any regression breaks CI. image

sweet, good to know thanks!

Issue #2938, opened 2026-07-21
Doxygen gotchas


ribbanya:
There are a number of really annoying quirks with Doxygen that need to be worked around somehow. Overall the `.dox` file approach might not be very reliable. I'll update this list as I find more.

- [ ] If you declare a function in a `.dox` file (any file external to the header itself), and you don't use `@fn`, it will duplicate the function results in the search index:
  ```c
  // tu.dox

  // the below declaration appears in tu.c, tu.h, and tu.dox in the rendered docs
  /// @brief Does a thing
  void foo(void);
  
  // the below declaration appears only in tu.c and tu.h, and is propagated to both;
  // tu.dox is not created in the index
  /** @fn void foo(void)
   * @brief Does a thing
   */
  ```
  Essentially, it treats the inline declaration as an independent declaration from a different source file, but then because it has the same name and signature, merges it with the others. Whereas the `@fn` reference approach explicitly tries to bind to an existing declaration elsewhere.
- [ ] `clang-format` breaks on `@fn` (doesn't trim spacing, etc.) because it considers it a normal C comment. So at that point you might as well just stop using the C-comment format of `.dox` and switch to markdown or something.
- [ ] `static` functions with identical name and signature in two different TUs will merge their documentation, regardless of whether `@fn` or inline declarations are used. The only workaround is to place the documentation at the definition. Placing it at the declaration might also work, but not an external documentation file.
- [ ] The `#` syntax is not a shortcut for `@ref` but actually for `::` (global scope), meaning there are times such as when referring to a directory it will not work. This also means that if `AUTOLINK_SUPPORT` is disabled in the config, `#` stops working altogether (because it's "auto-linking to the global scope").

itsgrimetime:
what about:
- `.dox` just has `@file`, `@page`, and `@defgroup`/`@ingroup` (if we even end up wanting these)
- prototype blocks in the headers
- blocks for statics can stay in the `.c`

downside is it makes header files longer & goes against a convention we already have documented in `CONTRIBUTING.md`. imo that feels like the natural place to put those, but I also don't read a ton of C projects myself so my intuition of whether that's a pain in the ass or not is probably not super dialed in

itsgrimetime:
but I do think it solves all 4 of your main gripes

ribbanya:
[...]

PR #3445, opened 2026-09-11
Document mnDiagram distance and number formatting helpers

itsgrimetime:
## Summary

- document the five `mnDiagram` distance and number-formatting helpers at their declarations in `mndiagram.h`
- remove inaccurate inline claims about display units, caps, and time formatting
- align the `mnDiagram_FormatDecimalNumber` prototype parameter name with its definition
- keep symbol documentation inline, following the layout agreed in #2938; this PR adds no `.dox` redeclarations

## Runtime validation

The documented boundary and formatting contracts were checked with 73 controlled Dolphin 2509 / GALE01 v1.02 cases, including both saved-language branches.

## Build verification

- rebased onto current `upstream/master`
- `ninja`
- 1130/1130 objects matching and linked
- rebuilt DOL SHA-1: `08e0bf20134dfcb260699671004527b2d6bb1a45`
- `git diff --check`
- clang-format clean


ribbanya (changes requested):
I'm not sure this is something we want to do, see #2938. We might be forced to place documentation at the definition. I'm not really sure what the best layout is but it's not great to churn one way and then the other. We need to look more at how large/stable codebases are handling this problem.

ribbanya:
The main reason for using .dox files was to document struct fields separate from the definition, mostly because HAL likes to declare 2000-line structs that are annoying to wade through if each field also has multi-line documentation. I really don't know what the best solution is.

itsgrimetime:
makes sense. i'll play around with some stuff and do some thinking as well and see if I can't come up with some ideas

itsgrimetime:
Updated this onto current master. The helper documentation now lives on the real declarations in `mndiagram.h`, with no external `.dox` redeclarations, following the symbol-documentation layout agreed in #2938. The rebased branch still reproduces the expected DOL and remains 100% matched/linked.

ribbanya (approved):

PR #3451, opened 2026-09-11
Name the mndiagram fighter and name display-order arrays

itsgrimetime:
Renames `mnDiagram_804A0750` and `mnDiagram_804A076C` to `mnDiagram_FighterDisplayOrder` and `mnDiagram_NameDisplayOrder`, following `mnName_NameDisplayOrder`: both are default/identity-filled, sorted by `mnDiagram_Sort*ByKOs`, and read back through the `*ByIndex` accessors. They become plain `u8` arrays, and the two single-field wrapper structs that only existed to express the adjacent-array pointer walk in `mnDiagram_SortNamesByKOs` are removed; that walk now uses an explicit base pointer.


src/melee/mn/mndiagram.c
```diff
@@ -797,23 +783,22 @@ void mnDiagram_SortNamesByKOs(void)
...
                   (GetNameText(*candidate) != NULL))))
             {
                 max_idx = j;
             }
         }
         if (max_idx != i) {
-            u8* p = &assets->sorted_fighters[max_idx];
-            u8 temp = *(p += sizeof(mnDiagram_804A0750_t));
+            u8* p = base + max_idx;
+            u8 temp = *(p += sizeof(mnDiagram_FighterDisplayOrder));
```
itsgrimetime: I made the assumption that the flattened access is better - lmk if that's not the case

ribbanya: The array access is better; `base` is fake. Other than that it's fine.

itsgrimetime: I couldn't get the plain array access to match quite yet, so will revert those parts

PR #3454, opened 2026-09-11
Use size instead of hardcoded size for memzero in gmvs, reorganize VsSceneController with a VsSceneState substruct

PsiLupan:


src/melee/gm/gmvs.c
```diff
@@ -1672,7 +1672,7 @@ void fn_8016DCC0(StartMeleeData* arg0)
     VsSceneController* tmp = &controller;
     int i;
 
-    memzero(&controller, 0x24C8);
+    memzero(&controller, offsetof(VsSceneController, start));
```
ribbanya: Surely this is `sizeof` a sub-struct?

PsiLupan: Yeah, that one was my bad. Have been auto-pilot looking at fixing several of these.

PsiLupan: Actually, wait no, it would be the size of the struct up to that particular struct. So outside of making a substruct just for that, I see no reason to.

ribbanya: I think that is how it works because `gmVs_GetController_0` returns the full thing and `gmVs_GetController_1` returns just the first struct.

PsiLupan: Both `gmVs_GetController_0` and `gmVs_GetController_1` return the same thing. `gm_GetStartMeleeRules` is returning from the `start` struct.

I think doing that will lead to me having to touch a lot more code than just fixing this right now, though. Otherwise I'm going to be touching a lot of functions.

ribbanya: That's fine, just merge this and I'll do it. If you look at the usages, nothing accesses `start` when it calls `gmVs_GetController_1`. I'd been wondering why both exist.

PsiLupan: I've got some work on it cooked up now, so I'll just make it part of this.

PR #3460, opened 2026-09-11
Create index page tracking cleanup tasks

sadkellz:
A simple page to show what is left to clean up in the code base.

### DIFF
--- a/.nix/melee-docs.nix
+++ b/.nix/melee-docs.nix
@@ -6,8 +6,18 @@
   runCommand,
   wibo,
   mwcc,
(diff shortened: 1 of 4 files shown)


ribbanya (changes requested):
@r-burns The nix setup looks hacky to me, can you take a look?

r-burns:
The Nix seems reasonable to me AFAICT, haven't checked the python code in-depth but I'm fine with this if the rest looks good to you.

ribbanya (approved):

PR #3461, opened 2026-09-11
Move .dox documentation inline and delete the .dox files

itsgrimetime:
Follows the consensus in #2938: symbol docs move inline, real prose becomes markdown, redeclaration boilerplate is deleted.

- All 59 `.dox` files are removed. Their 711 doc blocks now sit at the declaration: above the prototype in the header, above the `static` forward declaration in the C file, or on the struct field (`///<` for one-liners). The 975 undocumented prototype redeclarations and guard/include copy-pasta are dropped.
- The itflipper naming experiment log becomes `docs/itflipper-naming.md` (a Doxygen page that keeps the `itflipper_naming` anchor the briefs reference). Directory-layout todos and two notes whose targets no longer exist are collected in `docs/dox-migration-notes.md`.
- `@at`/`@sz` aliases are removed from the Doxyfile and stripped from sources; offsets are already carried by the `/* +NN */` comments and `ASSERT_SIZE`.
- `*.dox` is dropped from the Doxyfile file patterns, the clang-format hook, and `tools/functions.sh`; CONTRIBUTING.md describes the inline convention.
- Three prototypes gained parameter names so their moved docs bind: `Camera_SetBackgroundColor`, `Ground_801C4B50`, `gm_FindGameSceneHandler`.

Doxygen before/after with the repo Doxyfile: 59 `.dox` pages and 2512 redeclared members gone, warnings 254 -> 247, and every documented symbol with more than three words of text is still documented apart from the `@at`/`@sz`-only blocks. `main.dol` OK.


docs/dox-migration-notes.md
ribbanya, line 1: Can't we use `@dir` instead of these headers?

src/melee/ft/kinds/ftCaptain/ftcaptainspecials.h
```diff
@@ -2,27 +2,94 @@
 /// Side Special - Raptor Boost / Gerudo Dragon (SpecialS)
 #ifndef GALE01_0E3278
 #define GALE01_0E3278
+/**
+ * @file
+ * Captain Falcon & Ganondorf's Side Special -
+ * Raptor Boost / Gerudo Dragon (SpecialS)
+ */
```
ribbanya, line 5: This goes above the include guard. Merge with one-line brief.

ribbanya (changes requested):
If a declaration (including field definitions) has any documentation, it should have an extra newline before the docs begin and after the declaration.

Will re-review after these changes.

itsgrimetime:
> If a declaration (including field definitions) has any documentation, it should have an extra newline before the docs begin and after the declaration.

@ribbanya including `///<`'s?

ribbanya:
Let's keep the ones that don't break line with clang-format.

Oh, I thought you meant moving to markdown file. Yeah, just keep one-line comments without the extra spacing.

ribbanya (approved):
Seems good enough as a first pass, thanks.

PR #3465, opened 2026-09-12
gr: Mostly document GroundParam

RighteousRyan1:
# Summary 
Documented most of the parameters within `gr/types.h/GroundParam`. Some are still undocumented because changes were not apparent or obvious.

Tested via my own tool, [ExternalMeleeTool](https://github.com/RighteousRyan1/ExternalMeleeTool), and the Dolphin debugger. Changed values and observed differences during stage gameplay.


RighteousRyan1:
VS chose not to rename references to said parameters. Will re-PR tomorrow.

ribbanya:
You can just mark as draft for things like that. I might also be able to fix it up myself in a couple hours.

PR #3466, opened 2026-09-13
Add unoptimized debug build support

Jacoby6000:
## Overview

This adds support for producing a functioning non-optimized build, which opens up the doors to much improved debugging tools. 

Flags: `python3 configure.py --debug --sym on --map --no-optimize`

I have been able to successfully use this with my own fork of dolphin that implements a DAP debugger.  It  supports loading ELF structs, symbols, and source mappings.  Using it with this elf file allows us to do source level debugging; stepping through functions and inspecting structured memory using all the definitions from the incredible hard work y'all have done.

image

In optimized builds ,locals do not properly update. Stepping through the source does not work. So I worked out some stuff for unoptimized builds.

## Quirks

* vi.c, mtx.c, and THPDec.c all fail to compile when optimized.
* vec.c causes crashes when unoptimized. Menus are also not visible when it is not optimized.
* Some global allocations move around in this build because it is naturally non-matching.

## Related useful tools

- [dolphin-dap](https://github.com/LiveMindIO/dolphin-dap/blob/master/Tools/dap/README.md#mode-2-run-a-debug-elf-with-an-iso): A Dolphin fork with an embedded DAP server and full ELF support
- [dolphin-dap-nvim](https://github.com/LiveMindIO/dolphin-dap-nvim): A Neovim plugin for a dolphin-dap client
- [dolphin-dap-vscode](https://github.com/LiveMindIO/dolphin-dap-vscode): A VSCode plugin for a dolphin-dap client
- [dolphin-dap-mcp](https://github.com/LiveMindIO/dolphin-dap-mcp): An MCP Server for LLMs to interact with dolphin via DAP

The Dolphin fork allows specifying the elf separately from the ISO so you don't have to deal with replacing the dol file every time you rebuild.


configure.py
```diff
@@ -323,6 +328,26 @@
...
+NO_OPTIMIZE_CONTROL_OBJECTS = {
+    # MWCC fails register allocation for these first two at -O0. vi.c retains
+    # an otherwise dead reference to the unlinked __VIInitPhilips.
+    "dolphin/thp/THPDec.c",
+    "dolphin/mtx/mtx.c",
+    "dolphin/vi/vi.c",
+    # -O0 spills PSVECNormalize's paired-single temporaries through scalar FPR
+    # helpers, losing PS1 state and corrupting PSMTXRotAxisRad matrices.
+    "dolphin/mtx/vec.c",
+}
```
ribbanya: This is horrible, please refactor by adding `extra_cflags` at the object site or wrapping `Object` in a helper function, or something else that modifies the object listing instead of creating a special case.

configure.py
```diff
@@ -1996,6 +2034,19 @@ def link_order_callback(module_id: int, objects: list[str]) -> list[str]:
 # config.link_order_callback = link_order_callback
 
 
+# Unoptimized MWCC code calls EABI save/restore thunks that the retail link does not need.
+def no_optimize_link_order_callback(
+    module_id: int, objects: list[str]
+) -> list[str]:
+    if module_id == 0:  # DOL
+        return objects + ["Runtime/eabi_save_restore.s"]
```
ribbanya:
```suggestion
        return [*objects, "Runtime/eabi_save_restore.s"]
```

ribbanya (changes requested):

src/.gitignore
```diff
@@ -1 +1,2 @@
 *.s
+!Runtime/eabi_save_restore.s
```
ribbanya: Hm, not sure how this .s got here, might have been upstream dtk, but maybe we should be ignoring asm/ and committing .s by default, if the intention is to support asm mods.

configure.py
```diff
@@ -291,11 +296,11 @@
     "-align powerpc",
     "-nosyspath",
     "-fp_contract on",
-    "-O4,p",
+    "-O0" if args.no_optimize else "-O4,p",
     "-multibyte",
     "-enum int",
     "-nodefaults",
-    "-inline auto",
+    "-inline off" if args.no_optimize else "-inline auto",
```
ribbanya: I would rather see literal flags with format string parameters for the variants, like `f'-O{"0" if args.no_optimize else "4,p"}'`

configure.py
Jacoby6000: extra_cflags doesn't work unfortunately, because it just appends flags and doesn't replace duplicates.  I'll try a helper

ribbanya: You can add a dummy option to `Object.options` and interpret it through `Lib`, then unset it before it reaches `project.py`.

ribbanya: Then Lib is responsible for applying the optimization level.

ribbanya: Something like `options={"force_optimize": True}`

ribbanya (changes requested):
Hang on, I have a better idea.

ribbanya (approved):

PR #3476, opened 2026-09-13
Name hsd3a94 identifiers, first draft 

MarkMcCaskey:
Doing self review now, I can split this out as needed


ribbanya (approved):
Just going to merge this so I can work on it. Will hand it back later but I happen to really need this unit for some code I'm writing.

ribbanya:
Thank you for looking at this!

PR #3480, opened 2026-09-14
Name stage callback thin wrappers 

MarkMcCaskey:
Draft state. Based on earlier clean up we have lots of functions that just call a shared function. We can name these. No rush to ship this, there’s some potential inconsistency in the naming of similar symbols and want to get alignment and feedback before shipping anything here.

the ground animation functions are now basically all just forwarding to the same place. Those should be named. We also need a consistency pass here.


ribbanya (approved):

config/GALE01/symbols.txt
```diff
@@ -11996,18 +11996,18 @@ grTfalco_UnkStage0_OnLoad = .text:0x8022086C; // type:function size:0x4 scope:lo
...
+grTFalco_GObj0_GObjProc = .text:0x802209B8; // type:function size:0x4 scope:local
+grTFalco_GObj0_Callback3 = .text:0x802209BC; // type:function size:0x4 scope:local
+grTFalco_GObj2_OnInit = .text:0x802209C0; // type:function size:0x50 scope:local
+grTFalco_GObj2_Callback1 = .text:0x80220A10; // type:function size:0x8 scope:local
+grTFalco_GObj2_GObjProc = .text:0x80220A18; // type:function size:0x34 scope:local
+grTFalco_GObj2_Callback3 = .text:0x80220A4C; // type:function size:0x4 scope:local
+grTFalco_GObj1_OnInit = .text:0x80220A50; // type:function size:0x50 scope:local
+grTFalco_GObj1_Callback1 = .text:0x80220AA0; // type:function size:0x8 scope:local
+grTFalco_GObj1_GObjProc = .text:0x80220AA8; // type:function size:0x20 scope:local
+grTFalco_GObj1_Callback3 = .text:0x80220AC8; // type:function size:0x4 scope:local
```
ribbanya: These are regressions actually. We should be moving to this pattern, not away.

ribbanya (changes requested):

ribbanya (commented):
Where possible, localizing a symbol and giving it private naming conventions should be considered a cleanup target because it reduces API surface area _and_ naming complexity.

config/GALE01/symbols.txt
MarkMcCaskey: Great, thanks! That's what I wanted to check in with you on

ribbanya:
I have no idea about the clang-format job. The pre-commit action is really old and pointless, we need to move it to nix or raw yml.

MarkMcCaskey:
> I have no idea about the clang-format job. The pre-commit action is really old and pointless, we need to move it to nix or raw yml.

Looking into it, Claude found that it's likely the change around running things concurrently https://github.com/doldecomp/melee/pull/3482. Happy to replace the action with the raw steps too if needed but the reasoning of preinstalling clang format makes sense to me at a high level.

ribbanya (approved):

PR #3482, opened 2026-09-14
Preinstall clang-format for pre-commit

MarkMcCaskey:
From Claude:

> Fixes the intermittent `clang-format` failure in "Check code style" on pull requests (seen on #3480 and pr/lint), which dies before formatting anything with:
>
>    PermissionError: [Errno 13] Permission denied: '.../site-packages/clang_format/data/bin/clang-format'
>
> #3479 set `require_serial: false` on the clang-format hook, so pre-commit now runs it as several parallel processes. cpp-linter-hooks does not ship clang-format in its hook environment: on each invocation it runs `clang-format --version` and, when the pinned 22.1.8 is not on PATH (it never is on the runner), pip-installs the wheel into the shared environment from inside the hook process. Every parallel process starts that install at the same time, and one executes the binary while another is still unpacking it. It's timing-dependent, which is why reruns and master pushes pass.
>
> Declaring the wheel as `additional_dependencies` makes pre-commit install it once while creating the environment, before any hook process runs, so the run-time install path is never taken. Locally the race doesn't reproduce because a system clang-format 22.1.8 satisfies the version probe.
>
> The pre-commit/action cache was not involved: every run, passing or failing, logged "Installing environment" after restoring it.

Sounds like we don't need to change anything else but this but good to keep an eye on if this is a sufficient fix.


ribbanya (approved):
What, that's insane... Why...

PR #3489, opened 2026-09-14
Use tagged unions in HSD_3A94

MarkMcCaskey:
Semantics and names of card commands and enums tested here https://github.com/MarkMcCaskey/melee-runtime-audits/tree/main/pr3489/dolphin in both dolphin and simulating behavior.

Pretty big clean up. Some of the structure here with the padding of these card commands suggests the union might actually be subfields given that like the data pointer is always the same offset even if the preceding offsets are unused.

Can iterate more but I feel reasonably good about these changes in terms of correctness.

Also deletes a fake TU that a comment implies was needed for matches but is not needed and actually prevents deleting some of those hacky pointer math macros if it exists.


ribbanya (commented):
Could you do me a favor and merge in https://github.com/MarkMcCaskey/melee/compare/MarkMcCaskey:cardcmd-union...ribbanya:wip/cleanup?

PR #3497, opened 2026-09-15
Fix runtime conversion in gm_1884

r-burns:


src/melee/gm/gm_1884.c
```diff
@@ -780,9 +779,8 @@ void fn_801891F4(void)
             {
                 f32 selected_speed =
                     speed_stack.speeds.values[sub->menu_values[0]];
-                lb_80019880(
-                    __cvt_dbl_usll((f64) (0.016666668f / selected_speed *
-                                          (f32) gm_801891F4_GetTickRate())));
+                lb_80019880(1 / 60.0F / selected_speed *
```
ribbanya, line 782: We have `GM_FPS`.

PR #3499, opened 2026-09-15
dolphin: preserve fixed-width types on LP64 hosts

itsgrimetime:
## Summary

- keep Dolphin SDK `s32` and `u32` definitions 32-bit on non-Metrowerks LP64 hosts
- use the compiler pointer-width types for MSL `intptr_t` and `uintptr_t` outside Metrowerks builds
- declare SRAM fields and RTC APIs with `u32` where the target data and interface are 32-bit
- align CARD write and MetroTRK protocol declarations with their fixed-width contracts
- express the MSL UART console functions using their host-width callback types
- remove the custom source-location filter from the primitive `short`/`long` clang-tidy ban so the pre-commit hook controls its scope
- replace the newly exposed MSL, MetroTRK, and Runtime declarations with fixed-width, size, and pointer-width types

The Metrowerks definitions remain unchanged, while native builds no longer inherit LP64 `long` widths for fixed-format SDK data. The runtime memory and string implementations now use `size_t` for buffer lengths, `u32` for target words, and `uintptr_t` for pointer arithmetic.

## Verification

- full project clang-tidy hook passes without the custom source-location filter
- local pre-commit checks pass
- `ninja`
- `build/GALE01/main.dol: OK`
- 100.00% matched and linked; all code and data bytes matched


ribbanya (commented):
You should expand (or remove) the ban filter in [`.clang-tidy`](https://github.com/doldecomp/melee/blob/662250b9db8b3584b4739749ae0a64bb179e6390/.clang-tidy#L25).

itsgrimetime:
Oh hm I missed a few places these types are used, fixing now

.clang-tidy
```diff
@@ -22,7 +22,7 @@ CustomChecks:
   - Name: ban-primitive-c-types
     Query: |
       match varDecl(
-        isExpansionInFileMatching(".*/src/(melee|sysdolphin)/.*"),
+        isExpansionInFileMatching(".*/src/.*"),
```
ribbanya: Just remove it; our pre-commit hook decides what should be tidied (which is src/).
```suggestion
```

ribbanya: Unless we still need it on MetroTRK?

itsgrimetime: Nope should be good to remove, doing that now

PR #3508, opened 2026-09-15
hsd: use pointer-width keys for object IDs

itsgrimetime:
## Summary

- introduce `HSD_IDKey` for object-ID table keys
- keep keys as `u32` in Metrowerks builds and use `uintptr_t` in native builds
- apply the key type to the ID table API, entries, and stored JObj/AObj IDs
- update pointer-derived insertions and lookups so native 64-bit addresses are not truncated

The target ABI and structure layouts remain unchanged, while native builds can preserve the full width of pointer-derived object IDs.

## Verification

- clang-format passes
- clang-tidy passes
- `ninja`
- `build/GALE01/main.dol: OK`
- 100.00% matched and linked; all code and data bytes matched


ribbanya (commented):
1. Why can't `MWERKS_GEKKO` use `uintptr_t`?
2. Can some of these casts be elided?

itsgrimetime:
> 1. Why can't `MWERKS_GEKKO` use `uintptr_t`?
> 2. Can some of these casts be elided?

`uintptr_t` is `unsigned int` under `MWERKS`, whereas `u32` is `unsigned long`, even tho both are 32-bit, using `uintptr_t` for the target changes codegen in `HSD_AObjLoadDesc`, adding an extra `mr` before `HSD_IDGetDataFromTable`

all the casts replace existing (`u32`) casts and their inputs are pointers, removing them gives pointer-to-integer conversion errors in both Clang and MWCC. I also tried making `HSD_IDKey` pointer-typed, which removes the casts and passes native clang-tidy, but it causes the same mismatch in `HSD_AObjLoadDesc`, maybe there's another way around them? but this seems like a reasonable incremental improvement

src/sysdolphin/baselib/forward.h
```diff
@@ -102,6 +102,12 @@ typedef union HSD_ObjData HSD_ObjData;
 typedef union HSD_Rumble HSD_Rumble;
 typedef union HSD_TExp HSD_TExp;
 
+#ifdef MWERKS_GEKKO
+typedef u32 HSD_IDKey;
+#else
+typedef uintptr_t HSD_IDKey;
+#endif
```
r-burns:
```suggestion
typedef uintptr_t HSD_IDKey;
```

Does this break anything? Seems like it should unconditionally be pointer-width.

r-burns: Oh sorry, should have checked the discussion before leaving a comment :P

src/sysdolphin/baselib/forward.h
```diff
@@ -102,6 +102,12 @@ typedef union HSD_ObjData HSD_ObjData;
 typedef union HSD_Rumble HSD_Rumble;
 typedef union HSD_TExp HSD_TExp;
 
+#ifdef MWERKS_GEKKO
```
r-burns:
```suggestion
#ifdef MUST_MATCH
```

I think MUST_MATCH better reflects the situation here then, because a pointer-sized should be functionally correct but apparently fails to match due to int/long shenanigans. Hopefully we can resolve the mismatch and eliminate the condition in the future.

src/sysdolphin/baselib/forward.h
itsgrimetime: yep, it adds that extra `mr` in `HSD_AObjLoadDesc`

src/sysdolphin/baselib/forward.h
itsgrimetime: yeah good point `MUST_MATCH` is better here

r-burns (approved):

ribbanya (approved):

PR #3519, opened 2026-09-16
Partial cobj cleanup

HidoranBlaze:
Tried to clean up most of the cobj issues except for the MUST_MATCH in setupTopHalfCamera and the two PAD_STACKS.
First time contributing, so please let me know if I did something I shouldn't have.


HidoranBlaze:
I'm not too sure about the latest commit here, so feel free to reject it if necessary. I wasn't sure how to deal with the readability-non-const-parameter error pre-config was giving me, and I noticed that cobj_fabs_p was really only being used in one place to get the absolute value of a y field of a vector, which is why I thought changing the function this way could satisfy the pre-config hook and improve readability.

ribbanya:
Welcome, and thanks!

PR #3522, opened 2026-09-16
Speed up Nix builds via case-sensitive wibo

r-burns:
If wibo is given a nonexistent path to translate from windows, it currently walks the entire directory structure looking for case-insensitive file matches before eventually giving up.

I believe this is the current source of the runtime discrepancy in the Nix CI build - while this is an issue on all CI jobs, it affects the Nix job much more severely due to the large number of /nix/store paths that wibo is checking against.

Dropping this case-insensitive lookup seems to improve the performance of the Nix builds by about 50% (depending on the size of the store) and doesn't appear to cause any breakage because the paths we are looking up are already exact-case. The loss of case-insensitivity may be desirable actually, as it increases the strictness of the filesystem lookups and enforces that include paths are capitalized correctly.


ribbanya:
About 40 seconds, nice.

PR #3534, opened 2026-09-16
Use aggregate CardContext in native builds

itsgrimetime:
## Summary
- keep the three distinct retail card symbols, and the existing casts, unchanged in matching builds
- represent that storage as one real `CardContext` object in non-matching/native builds
- map the legacy `active_requests`, `commands`, and `requests` names to members of that aggregate so existing code accesses one declared object
- remove the duplicate `CARD_REQUESTS` definition

## Why
The previous revision only hid the repeated casts behind macros; it still treated three separately declared objects as one aggregate and therefore did not fix the portability issue. This revision changes the storage representation for native builds. The retail-only overlay is now explicitly confined to `MUST_MATCH`.

The earlier `mndiagram.c` change has been dropped because it had the same problem and provided no portability improvement.

## Verification
- `ninja`
- `build/GALE01/main.dol: OK`
- 100% code, data, objects, and linked output remain matched
- strict Clang syntax check of `card.c` with `MUST_MATCH` unset
- GitHub x86/GCC native build passed on the preceding equivalent aggregate revision
- all commit-time project checks pass


ribbanya (changes requested):
This does not the fix the portability problem. It still reads past the struct definition. It's also worse to look at than the casts.

itsgrimetime:
Thanks—you’re right. The accessor macros only centralized the same cross-object read and made the underlying issue less visible. I’ve replaced that revision: matching builds keep the original three retail symbols and casts, while non-matching/native builds now allocate one real `CardContext`. I also dropped the `mndiagram.c` change because it had the same flaw.

itsgrimetime:
Closing this after investigating the TU/layout suggestions from Discord.

A real `CardContext` in the existing card TU gets close, but not exact: 44/48 functions still match, while `fn_803AA790`, `fn_803AC168`, `fn_803AC2E0`, and `hsd_803B2374` change. In those functions retail materializes `commands`/`requests` as independent symbol addresses; MWCC materializes a real aggregate as `card_context` plus a member offset. Alternate standards-valid pointer expressions did not recover the retail code shape.

I also tried putting one aggregate in a data-only TU, following the former `hsd_4D11.c` boundary. That dropped the card unit to 28/48 matching functions because MWCC changed its addressing decisions, reproducing the reason that split had been artificial.

So the current `MUST_MATCH` version is runtime-safe for native builds, but it introduces an unprecedented conditional storage model for code that appears to have genuinely used these casts. That is not a sufficiently well-founded cleanup. Thanks for pushing on the distinction.

PR #3535, opened 2026-09-16
Fix native aliasing state views

itsgrimetime:
Consolidates gmregclear and Castle state views, and corrects owner-specific accesses in Brinstar, Big Blue Route, and Great Bay.

The GALE01 DOL remains 100% matched; pre-commit passes.


src/melee/gr/types.h
```diff
@@ -1608,10 +1627,34 @@ struct grCastle_GroundVars11 {
...
+STATIC_ASSERT(offsetof(struct grCastle_GroundVars4, xD6) ==
+              offsetof(struct grCastle_GroundVars9, xD6));
+STATIC_ASSERT(offsetof(struct grCastle_GroundVars4, xD8) ==
+              offsetof(struct grCastle_GroundVars9, xD8));
+STATIC_ASSERT(offsetof(struct grCastle_GroundVars4, xDA) ==
+              offsetof(struct grCastle_GroundVars9, xDA));
+STATIC_ASSERT(offsetof(struct grCastle_GroundVars4, xDC) ==
+              offsetof(struct grCastle_GroundVars9, xDC));
+STATIC_ASSERT(offsetof(struct grCastle_GroundVars3, x1C) ==
+              offsetof(struct grCastle_GroundVars9, dynamics));
```
ribbanya:
```suggestion
```

src/melee/gr/types.h
```diff
@@ -1486,13 +1496,20 @@ struct grBigBlueRoute_Track {
...
+    /* +0C gp+CC */ Vec3 xCC;
+    /* +18 gp+D8 */ struct grBigBlueRoute_Track tracks[4];
 };
 
+STATIC_ASSERT(offsetof(struct grBigBlueRoute_GroundVars2, xC4) ==
+              offsetof(struct grBigBlueRoute_GroundVars, xC4));
+STATIC_ASSERT(offsetof(struct grBigBlueRoute_GroundVars2, xC8) ==
+              offsetof(struct grBigBlueRoute_GroundVars2, xC4) +
+                  sizeof(HSD_GObj*));
+
```
ribbanya:
```suggestion
```

src/melee/gr/types.h
```diff
@@ -650,6 +654,13 @@ struct grGreatBay_GroundVars4 {
     Item_GObj* xF0;
 };
 
+STATIC_ASSERT(offsetof(struct grGreatBay_GroundVars4, xD4) ==
+              offsetof(struct grGreatBay_GroundVars, xD4));
+STATIC_ASSERT(offsetof(struct grGreatBay_GroundVars4, xD8) ==
+              offsetof(struct grGreatBay_GroundVars, xD8));
+STATIC_ASSERT(offsetof(struct grGreatBay_GroundVars4, xE0) ==
+              offsetof(struct grGreatBay_GroundVars, xE0));
+
```
ribbanya:
```suggestion
```

src/melee/ft/kinds/ftPikachu/types.h
```diff
@@ -96,9 +96,26 @@ union ftPikachu_MotionVars {
...
 
+#ifndef MUST_MATCH
+STATIC_ASSERT(offsetof(struct ftPikachu_SpecialLwVars, x4) ==
+              offsetof(struct ftPikachu_SpecialHiVars, x4));
+STATIC_ASSERT(offsetof(struct ftPikachu_SpecialLwVars, x4) ==
+              offsetof(union ftCommon_MotionVars, barrel.x4));
+STATIC_ASSERT(offsetof(struct ftPikachu_SpecialLwVars, x0) ==
+              offsetof(union ftCommon_MotionVars, barrel.x8));
+#endif
+
```
ribbanya:
```suggestion
```

ribbanya (changes requested):
These really aren't accomplishing much.

itsgrimetime:
> These really aren't accomplishing much.

makes sense, will remove these asserts

src/melee/ft/kinds/ftPikachu/types.h
```diff
@@ -96,8 +96,16 @@ union ftPikachu_MotionVars {
...
+#ifdef MUST_MATCH
+        /// Original member order required by the PowerPC build.
         Item_GObj* x0;
-        bool x4;
+        s32 x4;
+#else
+        /// Keep the cross-state x4 write at host offset 4.
+        u8 pad_x0[4];
+        s32 x4;
+        Item_GObj* x0;
```
r-burns: Could you explain why this is needed? Not sure I understand

itsgrimetime: so for context I was trying to do what you recommended and look at some of the commits on ports and see what problems they had to fix (this one was based on https://github.com/999sian/melee-pc/commit/393a4b5970d378d700b84dd9465fd286779b1ec9), and I think I misinterpreted what was actually happening in that commit, because i'm pretty sure there's actually more needed to complete this. 

the idea though is to preserve the +4 field of Fighter::mv across 32 and 64bit build because otherwise it will grow to 8 bytes and can overwrite the next pointer.

does that make sense? I ask because I think it does but i'm still kinda getting my bearings with this portability stuff. if this isn't the right way to do this, or if the decomp isn't the right place to keep something like this, just let me know

r-burns: Ah, gotcha. This may be a real portability issue, but I don't think flipping the order here is the right fix for the decomp - I think what they are seeing is that the decomp code is relying on these being contiguous in memory, but their compiler flips the order vs MWCC, so they're flipping the declaration order back to counteract this. If this is what's happening, I don't think it's feasible for the decomp to determine the memory ordering for every possible compiler - a more proper fix would be to group these into a struct, if that's really what's happening here.

src/melee/gr/grcastle.c
```diff
@@ -1485,8 +1485,7 @@ static inline void grCastle_PickSatellite(Ground* gp, s32* wp)
     }
     gp->u.castle12.xD0 = slot;
 
-    sat_gp = (Ground*) ((HSD_GObj*) gp->u.castle12.xC4[gp->u.castle12.xD0])
-                 ->user_data;
+    sat_gp = (Ground*) (gp->u.castle12.xC4[gp->u.castle12.xD0])->user_data;
```
ribbanya: We do have `GET_GROUND`, if that matches.

src/melee/gr/types.h
```diff
@@ -639,7 +639,11 @@ struct grGreatBay_GroundVars3 {
 
 struct grGreatBay_GroundVars4 {
     s32 xC4;
-    s32 xC8;
+    /// Align this view with grGreatBay_GroundVars::xC8 on native hosts.
+    union {
+        HSD_Generator* pad_gen;
+        s32 xC8;
+    };
```
ribbanya: See https://github.com/doldecomp/melee/pull/3535#discussion_r4033717607.

src/melee/gr/types.h
```diff
@@ -1501,16 +1505,18 @@ struct grCastle_GroundVars {
...
-    /* +12 gp+D6 */ s16 xD6;
-    /* +14 gp+D8 */ s16 xD8;
-    /* +16 gp+DA */ s16 xDA;
-    /* +18 gp+DC */ s16 xDC;
+    /* +00 gp+C4 */ HSD_GObj* pad_gobj[3];
+    /* +18 gp+D0 */ u8 pad_D0[0xD6 - 0xD0];
+    /* +1E gp+D6 */ s16 xD6;
+    /* +20 gp+D8 */ s16 xD8;
+    /* +22 gp+DA */ s16 xDA;
+    /* +24 gp+DC */ s16 xDC;
```
ribbanya: If the problem is that these are pointers, can't we just merge `grCastle_GroundVars4` into `grCastle_GroundVars9`? All of these union types are placeholder anyway, we just start by identifying which gobj id is using the type.

src/melee/gr/types.h
```diff
@@ -1608,9 +1614,9 @@ struct grCastle_GroundVars11 {
 };
 
 struct grCastle_GroundVars12 {
-    /* +00 gp+C4 */ u32 xC4[3];
-    /* +0C gp+D0 */ s16 xD0;
-    /* +0E gp+D2 */ s16 xD2;
+    /* +00 gp+C4 */ HSD_GObj* xC4[3];
+    /* +18 gp+D0 */ s16 xD0;
+    /* +1A gp+D2 */ s16 xD2;
```
ribbanya: Same here, this is surely the same union type as the others.

ribbanya (changes requested):

src/melee/ft/kinds/ftPikachu/types.h
ribbanya: While I agree it's not feasible to try to cater to every compiler, seeing the divergent behavior at specific struct fields is pretty useful because it indicates that we had a pointer type wrong. Maybe we can do a gcc-native build in x64 to flag these somehow?

ribbanya (approved):

PR #3539, opened 2026-09-18
Clean up gmmainlib

MarkMcCaskey:
A few misc cleanups for gmmainlib.

Use an array for referencing modes.

Share a stats struct in more places.

Corrected some types too.

Add opinionated bitwise helpers. This may be a bit overdone, I don't know how much a helper for `1 << N` is really useful but it at least makes it obvious that many top level functions are very similar visually so I considered these helpers a net win overall but I'm not attached to them if we want to just have the code using bitwise logic directly which is honestly closer to how'd I'd write it in greenfield myself.

Replaced the super verbose arithmetic with 0xFF... with a saturating add helper. Makes the code much less visually noisy but open to feedback here too.


src/melee/gm/forward.h
```diff
@@ -11,7 +11,8 @@
 /// @{
 #define GM_MAX_PLAYERS 6
 #define GM_MAX_TEAMS 5
-#define GM_NAMETAG_NONE 120
+#define GM_NAMETAG_COUNT 120
+#define GM_NAMETAG_NONE GM_NAMETAG_COUNT
```
ribbanya: Introduces unnecessary semantic complexity

src/melee/gm/types.h
```diff
@@ -416,25 +403,39 @@ struct gmm_x0_vsdata {
...
+            /* 0x0F90 */ VsModeData unk_F90;         ///< giant melee
+            /* 0x10D0 */ VsModeData vs_stamina;      ///< stamina melee
+            /* 0x1210 */ VsModeData unk_1210;        ///< slowmo melee
+            /* 0x1350 */ VsModeData vs_lightning;    ///< lightning melee
+            /* 0x1490 */ VsModeData unk_1490; ///< multiman, 3/15 min, endless,
+                                              ///< cruel
+            /* 0x15D0 */ VsModeData unk_15D0; ///< unused?
+            /* 0x1710 */ VsModeData unk_1710; ///< opening movie?
+        };
+        /* 0x0590 */ VsModeData table[GM_VSMODE_COUNT];
```
ribbanya: Does this always match? If so, remove union.

ribbanya (changes requested):

src/melee/gm/types.h
```diff
@@ -411,31 +411,24 @@ struct gmm_x0_44_t {
...
-            /* 0x0F90 */ VsModeData unk_F90;         ///< giant melee
-            /* 0x10D0 */ VsModeData vs_stamina;      ///< stamina melee
-            /* 0x1210 */ VsModeData unk_1210;        ///< slowmo melee
-            /* 0x1350 */ VsModeData vs_lightning;    ///< lightning melee
-            /* 0x1490 */ VsModeData unk_1490; ///< multiman, 3/15 min, endless,
-                                              ///< cruel
-            /* 0x15D0 */ VsModeData unk_15D0; ///< unused?
-            /* 0x1710 */ VsModeData unk_1710; ///< opening movie?
-        };
-        /* 0x0590 */ VsModeData table[GM_VSMODE_COUNT];
```
ribbanya: Sorry, I meant that if the table matches everywhere, the table is better than the struct.

src/melee/gm/forward.h
```diff
@@ -65,6 +65,26 @@ typedef enum GameModeKind {
     /* 2D */ GM_COUNT,
 } GameModeKind;
 
+/// Index into #gmm_x0_vsmodes::table.
+typedef enum VsModeKind {
+    /* 00 */ VSMODE_MELEE,
```
ribbanya:
```suggestion
    /* 00 */ GmVsMode_Melee,
```
etc

ribbanya (approved):

